PR #143, opened 2014-04-01
Add support for feeding GCC's build warnings.

francoisferrand:
The sonar.cxx.compiler.parser property can be used to select either vc (default)
or gcc build log parser.


guwirth:
Docu: https://github.com/wenns/sonar-cxx/wiki/Compilers

PR #252, opened 2014-07-03
introduce project base dir property  to search for report path

jmecosta:
in multi module configuration we dont have access to project base dir. it usefull to include all the reports in this folder and be able to resolve the reports from there. This will be the case if using the visual studio bootstrapper. 

Pull request created to start discussion.

Later it might need to be extended to support include directories. I did not do that because i always use absolute paths to resolve include directories


jmecosta:
#358

PR #335, opened 2014-12-23
Extend XPathCheck with path filter

guwirth:
- matchFilePattern allows Ant-style matching patterns for the path.
- If no matchFilePattern is defined all files are evaluated.

Main idea of this extension is to restrict a check to special:
- file extension
- filename
- path


guwirth:
docu: https://github.com/wenns/sonar-cxx/wiki/Create-Custom-Rules

Issue #179, opened 2014-04-18
Verify for precompiled header files


guwirth:
I like to verify if precompiled header file is used in the right way. For Visual Studio this means:
* Each source code file must have a include StdAfx.h. Include must be the first include in the file.
* It is not allowed to use StdAfx.h in other header files. 

Add new check or use regular expression or xpath?

guwirth:
Closed with #339, #341

PR #339, opened 2014-12-27
File RegEx rule

guwirth:
This rule template can be used to create rules which will be triggered when a file matches a given regular expression. Additional a Ant-style matching pattern for the path can be defined.

Also a solution for #179.


guwirth:
docu: https://github.com/wenns/sonar-cxx/wiki/Create-Custom-Rules

Issue #278, opened 2014-09-01
cppcheck report results not loaded as issues in Sonar: how to troubleshoot


jfallot:
Current env:
. Sonar 4.4
. C++ (Community) plugin version 0.9
. Using sonar-runner to run sonar analysis

My problem: although the cppcheck report result is available and the sonar analysis successful (metrics OK), No issues are showing up in Sonar (although there are a lot...).

Checks done:
. Run sonar-runner with -X: not much useful information is dumped about cxx plugin execution
. Numerous tests on cppcheck report path (absolute, relative)

In a previous env:
. Sonar 4.2
. Maven used to launch Sonar analysis
The issues were correctly reported in Sonar.

My Questions:
How can I more deeply debug the C++ plugin when launching the analysis by sonar-runner?
Attach a remote debugger to Sonar server?
What is the right code branch for 0.9 version of the C++ (Community) plugin?

hikoust:
Hi,

Can you please share the modified CxxReportSensor class file or share Git location ? I am using Cxx scan in Sonar and facing the same issue. I dont have required environment to compile sonar source code.

Thanks,
Kaushik

guwirth:
@jfallot could you test that please with 0.9.3-RC1

jmecosta:
@guwirth -> close -> release -> re-open if does not work

PR #344, opened 2014-12-30
Allow absolute report path to fix issue #278, which can help for issue #252

jfallot:


wenns:
Thanks for the PR. I do not, however, think that this is the right solution for issues with multi-module setups. I'll return to this after the release.

jfallot:
@wenns: A better solution to help in case of multi-module setups is welcome. This PR is however focused on allowing absolute report paths, which can be useful in multiple cases (although not all as @jmecosta mentionned). For the time being, I continue using my own fork.

jmecosta:
@JF https://github.com/wenns/sonar-cxx/pull/358 should handle your use
case... or if not can you produce a example project where you are unable to
import reports using the current relative paths (once #358 is merged)

jfallot:
@jmecosta, it looks like https://github.com/wenns/sonar-cxx/pull/358 is still "Work in progress" and it looks too early for me to make a try with it.
I will try to build a sample project showing that absolute report path is needed.

jmecosta:
it should be green now, but its not :) for some reason. they are passing in my environment. i need to check in unix

jmecosta:
seems http://sonarqube.15.x6.nabble.com/Trying-to-get-gcov-reports-working-td5031007.html the c++ sonar source allows absolute paths... probably you should do the same...

guwirth:
Seems to be outdated.
Update PR if still needed.

PR #346, opened 2015-01-01
Bullseye bdd

jmecosta:
bdd tests:
 - bullseye coverage
 - gtest filename tag
 - tests with isolated databased containing vera, cppcheck, rats, cpplint, common (other repository)
 - shows cases windows build and analysis with specif msbuild tasks


jmecosta:
i will need to update some files after the filename tag in test case is permitted

jmecosta:
@wenns any idea how we could define include paths based on os?

      Assertion Failed: Found following errors and/or warnings lines in the logfile:
      16:55:06.288 WARN  - the include root /home/travis/build/wenns/sonar-cxx/integration-tests/testdata/googletest_bullseye_vs_project/C:/Program Files (x86)/Microsoft Visual Studio 12.0/VC/include doesnt exist
      16:55:06.289 WARN  - the include root /home/travis/build/wenns/sonar-cxx/integration-tests/testdata/googletest_bullseye_vs_project/\C:/Program Files (x86)/Microsoft SDKs/Windows/v7.1A/Include doesnt exist

might be include an additional specifier ? something like this:
sonar.cxx.includeDirectories=../Packages/gtestmock.1.7.2/build/native/include, \\\
                             ../Packages/gtestmock.1.7.2/build/native/include/googletest

sonar.cxx.windows.includeDirectories=C:/Program Files (x86)/Microsoft Visual Studio 12.0/VC/include, \\\
                             C:/Program Files (x86)/Microsoft SDKs/Windows/v7.1A/Include

sonar.cxx.unix.includeDirectories=/usr/include/vc++/include

guwirth:
Think this is only a problem for integration/regression tests. In real world scenarios users have only one OS.

Would more propose to do it like Microsoft: 
* set/define environment variables
* support environment variables in paths

```
sonar.cxx.includeDirectories=$(XXX)/include
```

For other users this has also the advange to reuse predefined VS environment variables after calling ```C:\Program Files (x86)\Microsoft Visual Studio 10.0\Common7\Tools\vsvars32.bat```

jmecosta:
reusing properties would be great... if you run from maven runner this actually works, with sonar runner there is a ticket to support this
http://jira.codehaus.org/browse/SONARUNNER-76 but its seating there for really long time. set to be fixed in 2.5 cant be sure...

anyway am not sure if this would solve my issue. since the paths will be different anyway... and yes this is mostly  a case for the bdd tests.... but we need to have something robust to able us to run tests in any environment...

guwirth:
Like the idea. sonar.cxx is under our control so maybe we can add it at a central place in CxxConfiguration? Can have a look next week...

jmecosta:
we have to be careful, i think replacing env vars should be done by the runner. so that if they implement it it might brake our implementation.. most likely those will be replaced by the sonar runner already.

jmecosta:
a few things needed:
. we need to be able to enable rules in profiles.
. we need to create rules so we can test the external rules.
. we need to be able to reference the correct include folders depending on the system

after the test should pass

jmecosta:
i think i found a way to work around the create/enable rules.

its possible to create a h2 database in a external sonar server and then deploy it just before starting sonar

wenns:
I hope you're not expecting me to merge this? ;)
This is *way* too much stuff. I really like it to be minimal, i.e. containing just peaces needed to test the "coverage-using-bullseye-reports"-feature. Nothing more. I really hate complex tests.

As for platform specific includes: try to strip your sources to the bones, such that they avoid including stuff from the enviroment. For the remaining cases accept a couple of "WARN .... cannot find ... " messages in the log file. That the way I did the the other tests so far.

Support of env. vars in the properties file: I dont think its wise to come up with our own solution.

Your database hack: lets avoid it. This is likely to break with newer releases. Try to use the REST-API for activating/deactivating rules and also the rest of setup/teardown tasks. If we cannot do it this way, I rather would skip the tests than resort to such tricks.

jmecosta:
oh... :) for that i will take a bit more time, no idea how to create setup method in behave, also dealing with the rest api is going to take a bit of time to get all those into python... but we really need a test to verify the external sensor. at this point the only way is to use the template rule and create rules to validate the report. this test is actually quite simple... its just importing existing reports but it needs those define.

wenns:
I absolutely agree that the external rules stuff is something that is quite likely to break with newer releases (it already has two(?) times) and it needs a tests. I will have a look at this.

But *this* PR is really **way** too big:

140 changed files  with 38,634 additions and 2 deletions.

also, it calls itself 'bullseye bdd' but seems to contain tests for the external rules...

How about a nice & cute PR just dealing with BDD tests for bulleye-coverage parts?

wenns:
BTW: this:
http://nemo.sonarqube.org/api_documentation#api/qualityprofiles
looks like it can be used to activate/deactivate rules in quality profiles. I'll try to write an according behave step.

jmecosta:
those rest calls require a profile key that is not the same as what we have
in the web ui. ive done already in f#...

see
https://github.com/jmecosta/VSSonarQubeExtension/blob/master/SonarRestService/SonarRestService.fs

to get the profile you need GetProfilesUsingRulesApp:
httpconnector.HttpSonarGetRequest(conf, "/api/rules/app")

and to enable the rule you do:
member this.ActivateRule(conf:ISonarConfiguration, rule:Rule,
profilekey:string) = let errorMessages = new
System.Collections.Generic.List<string>() let url
="/api/qualityprofiles/activate_rule" let optionalProps =
Map.empty.Add("profile_key", HttpUtility.UrlEncode(profilekey))
.Add("rule_key", rule.Key) .Add("severity", rule.Severity.ToString())

   let errorMessages = new System.Collections.Generic.List<string>()
            let response = httpconnector.HttpSonarPostRequest(conf, url,
optionalProps)

the responses for each you can find in
https://github.com/jmecosta/VSSonarQubeExtension/blob/master/SonarRestService/IssuesTypes.fs

this only works if you are autheticated... so the post request needs to be
authenticaed with basic autentication with admin/admin...

just not that proficient and python to handle this autentication mechanism,
if you manage to get those in there i could then update the test.

for the creation of the rule i can then take your example and extend it to
create the relevant behave step

jmecosta:
@wenns ok this is easier than i tough... check last commit to see if thats something we can use

wenns:
This definitely goes in the right direction, and I would like to pull that as in its 'clean&lean' form. But lets do the release firtst and without that. We are on the finish line and this is a litt' bit more work...

jmecosta:
That's OK, was not expecting to go in this release... I will create a clean
pull request maybe today or tomorrow. Having issues deleting template rules.

jmecosta:
#352

PR #352, opened 2015-01-03
Bdd/bullseye and custom rules

jmecosta:
striped down to only needed files for me to create new data.

taken out the build look up tables from xunit sensor in case filename tags are found


jmecosta:
i would like to create a bdd for multimodule, but would like this one to be merged so i can use the same feature and reuse the test data...

jmecosta:
rebased

integration-tests/features/test_create_rules_and_bullseye_import.feature
```diff
@@ -0,0 +1,65 @@
+Feature: GoogleTestWithBullseyeAndVsProject
+
+  This test verifies that analysis is able to import bullseye coverage reports and import custom rules reports
```
guwirth: Could you add a more detailed description what the test is doing please. Makes it easier for others to understand and extend it. THX.

guwirth:
PR looks good for me.

integration-tests/features/test_create_rules_and_bullseye_import.feature
jmecosta: Will do

PR #358, opened 2015-01-04
Enhancement/multi module with reactor

jmecosta:
kind of rebase... clean reactor


wenns:
Now I have played a little bit this this stuff I know better whats going on.

This multi-module thing is a mine field. This PR hacks the plugin to make the modules pull reports which actually meant to be project-level. This has all sorts of effects, e.g.:
- the xunit simple mode is not working
- reports are processed N times (1 would be sufficient)
- reports are producing lots of wanings of type 
  'could not find file <bla>' because this file is in an other module
- project-level violations will be assigned to each module
etc.

I have mixed feeling with this.  Can we do better? Comments?
Meanwhile I'll test the scenario where reports are per module and see how it works.

jmecosta:
the problems you pointed out have been already also pointed out for c# plugin.  and in fact they have a ticket http://jira.codehaus.org/browse/SONARVS-59 to make it optional. this would mean that we actually do not need any of this (?). also having mix feelings

am not sure what people that use multimodule not in this way say about. any way... 

imo the issues you mention are not a implementation issue of the cxx plugin, rather a way the platform deals with this. any way, you should probably state that when using multi modules all these limitations apply. at least until we figure out

jmecosta:
not sure if this is a good idea, but it prevents issues with duplicated metrics

jmecosta:
yeah this is a mess :) but the alternative for us is to run the analysis per module and save xmls per module. this is a much harder setup.

ive made so now that the xunit sensor in simple mode should run only at root of the project.. added the correspondent bdd.

wenns:
Mess it is... Raw and hacky. And Im not even half through the most important scenarios for multi-module. Also, there is no clear policy or spec from the platform or other plugins, at least none Im aware of. This feature seems to come from the witsh to support maven modules, without much design work.
Anyway. I want to experement further and Im willing to finish it. We'll see.

jmecosta:
not sure whats going on, on window the tests are passing just fine. in unix they are not... apparenty the search function is not able to find the resources

jmecosta:
should now be green... it was a issue with the path separator in the reports

wenns:
Sorry, will skip this in 0.9.2

jmecosta:
ok

jmecosta:
@Bertk what do you wish to do with? is it worth rebasing? For me am happy with not using any of the module approach...

guwirth:
For me there are two questions:
* Who is currently using the multi module approach? Who is familiar with it? Who can test it if we add it?
* Which is the first SQ version where this feature is stable and functional?

jmecosta:
Its easy to test. Try to run a c#/c++ project with visual studio
bootstrapper on. Try to place your reports in the project root in this case
they will not be imported.

With 4.5.2 this should be functional and stable. But have not investigate
if improvement s have been done in other plugins

Bertk:
I use it in a productive environment (SQ 4.5.4) and it reduces the maintenance effort for the builds (prerequisite: VS bootstrapper plug-in is used).
An alternative is the commercial views plug-in (Portfolio views) and this will create aggregation of multiple build targets but you need to have strict naming conventions to create views based on regular expressions. Unfortunately this is a problem for C++ legacy projects and definition the views with lists of build targets is time consuming and error prone.
We use TFS builds and the VS bootstrapper is definitely a benefit creating SonarQube metrics for solutions with and without the views plug-in.
By the way, the VS bootstrapper has only limited C++ support and there is only a small part missing (https://github.com/SonarCommunity/sonar-visual-studio/pull/7)

jmecosta:
So should we merge this? I can rebase it

guwirth:
This was a lot of work and is an useful feature.
@Bertk If you could do final testing I would merge it after rebase.

Bertk:
I can test it next Sunday only.

jmecosta:
@Bertk i had to disable one of your tests. the vc log contains absolute paths that do not correspond to the place where i run the tests. so they fail. can you take a look to see if test data needs to be modified?

jmecosta:
@guwirth something wrong with Travis some builds are failing while running behave

guwirth:
@jmecosta I had this also twice this week. After triggering again a build with the same sources it worked. I'm not that familiar with Travis. I'm not sure if this is a Travis issue or a problem with our build script.

integration-tests/testdata/cpp-multimodule-project/sonar-project.properties
```diff
@@ -0,0 +1,56 @@
+sonar.host.url=http://localhost:9000
+
+sonar.modules=cli,lib,package1,package2
+cli.sonar.projectBaseDir=cli
```
guwirth: Are all these settings (cli, doxygen, ...) needed or can they removed?

integration-tests/testdata/cpp-multimodule-project/sonar-project.properties
```diff
@@ -0,0 +1,56 @@
...
+package1.sonar.projectName=package1
+package2.sonar.projectBaseDir=package2
+package2.sonar.sources=.
+package2.sonar.projectName=package2
+
+sonar.projectKey=cpp-multimodule-project
+sonar.projectName=cpp-multimodule-project
+sonar.projectVersion=Work
+sonar.dotnet.key.generation.strategy=safe
+sonar.scm.enabled=true
```
guwirth: In the other configuration files I set scm to false. Good to see how this is working.

jmecosta:
@guwirth @Bertk if all goes well in travis this pull request should be ok. ive enabled the test that ive earlier disable and change the paths in the report to be relative. 

Paths there were of type : c:\tfs_server\bla ive patched those one to be relative. (usually this is not a issue because c:\tfs_server is the root of the project) i think this shoud be fine to test the sensor

guwirth:
There were some complaints from @wenns in January, are they solved/answered?
- the xunit simple mode is not working
- reports are processed N times (1 would be sufficient)
- reports are producing lots of wanings of type 
  'could not find file <bla>' because this file is in an other module
- project-level violations will be assigned to each module
etc.

jmecosta:
I think some, but I guess improvements can be done later. This initial
support is good enough for multi module users. But is up to you guys.

integration-tests/testdata/cpp-multimodule-project/sonar-project.properties
jmecosta: Only doxygen is redundant but its not worth it

guwirth:
@jmecosta could rebase and squash (combine multiple commits into one) your PR please. I will merge it then for testing.

jmecosta:
You ended up merging, was doing the rebase a bit later. But not now :)

guwirth:
Sorry don't catch you. Should I redo the merge? Thought is finished for testing?

> Am 18.04.2015 um 10:59 schrieb Jorge Costa <notifications@github.com>:
> 
> You ended up merging, was doing the rebase a bit later. But not now :)
>

jmecosta:
No its fine, you ask to rebase and squash. But you ended up merging as is

Issue #49, opened 2013-11-15
[Parser Error ] "::" expected but ">>" [BW_RSHIFT] found


wenns:
The relevant thread where ambiguity is described:
http://sonar.15.x6.nabble.com/sonar-dev-SSLR-for-C-community-Plugin-td5009119.html

ametselaar:
In https://github.com/ametselaar/sonar-cxx/tree/nested-templates I have solved the issue for nested templateId's by allowing to close two template argument lists with a single '>>'. C++ style casts are not supported (yet).

jmecosta:
great, i gave a try... and results are much better... seems there is a case that fails

template<class T, class Compare = std::less<T>>

anyway, hope you dont mind. ive rebase you last commit against current upstream: https://github.com/wenns/sonar-cxx/pull/378

guwirth:
What is the state of this?

jmecosta:
#400 has been updated with latest commit

PR #378, opened 2015-01-11
Support closing two template argument lists with ">>" rather than "> >".

jmecosta:
CxxGrammarImpl.java(templates): Introduce new language elements
    innerTemplateId, innerTemplateArgumentList, innerTemplateArgument
    to enable parsing of new nested template syntax.
CxxGrammarImpl.java(declarations): Relax rules for gnu style attribute syntax.
CxxGrammarImpl.java(declarations): Allow restrict als trailing type specifier.
.../own/declarations.cc: Add nested declarations for testing.

(cherry picked from commit 48f9750d64c748ecec1725eb7239bea3cb666344)

Conflicts:
	cxx-squid/src/main/java/org/sonar/cxx/parser/CxxGrammarImpl.java

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/parser/CxxGrammarImpl.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/parser/CxxGrammarImpl.java
@@ -256,6 +256,9 @@ public enum CxxGrammarImpl implements GrammarRuleKey {
   templateName,
   templateArgumentList,
   templateArgument,
+  innerTemplateId,
+  innerTemplateArgumentList,
+  innerTemplateArgument,
   typenameSpecifier,
   explicitInstantiation,
   explicitSpecialization,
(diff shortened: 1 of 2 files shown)


guwirth:
Please remember the Test cases:
```C++
X<(1>2)> x2;
Y<X<1>> x3;
vector<pair<int,int>> s;
static_cast<List<B>>(ld);
dynamic_cast<List<B>>(ld);

jmecosta:
relaates #49, no supported things described by Arnold

ametselaar:
The test cases above have been added in df54eb14ca48204669fd38580fea3dde77a6f692, as well as a fix for ```C++ 
template<class T, class Compare = std::less<T>>```. The 'dynamic_cast' was not added as test case because I failed to come up with a context in which it is valid C++.

jmecosta:
@ametselaar cherry pick commit. try in same code base. problem now is solved... thank you a lot

guwirth:
Great if this is really fixing the template problem.

> dynamic_cast<List<B>>(ld);

With dynamic_cast your are probably right. This must be always 'type *' or 'type &' which avoids >>.

Issue #375, opened 2015-01-10
boosttest integration test


guwirth:
Started with boost test integration test on my fork. Will continue during the next weeks...

https://github.com/guwirth/sonar-cxx/tree/integration/boosttest/integration-tests/testdata/boosttest_project

Issue #372, opened 2015-01-07
Boost Unit Test, xunit.provideDetails=true, filename tag


guwirth:
Does someone know how to create boost unit test xml reports with file information?

jmecosta:
Windows? I've created a msbuild task that runs gtest and converts reports.
Not sure if there something similar out there for boost

guwirth:
Yes Windows. I'm interested in the boost options to create test-suite or test-case tags with a file attribute.

jmecosta:
this is the task ive created:
https://github.com/jmecosta/MSBuild-SonarQube-CXX/blob/master/GtestXunitConverterTask/GtestXunitConverterTask.fs

and its used like this:
https://github.com/jmecosta/sonar-cxx/blob/bdd/bullseye-and-custom-rules/integration-tests/testdata/googletest_bullseye_vs_project/PathHandling.Test/PathHandling.Test.vcxproj

just not sure how much work you will require to adapt it to boost... sorry

guwirth:
Did some tests, see my results below. Did also an update in the wiki. Also this needs a new XSL.

Example:
```C++
#define BOOST_TEST_MODULE my_module
#include <boost/test/unit_test.hpp>

int add(int i, int j) { return i + j; }

BOOST_AUTO_TEST_SUITE(my_test_suite)

BOOST_AUTO_TEST_CASE(my_test)
{
	// seven ways to detect and report the same error:
	BOOST_CHECK(add(2, 2) == 4);         // #1 continues on error

	BOOST_REQUIRE(add(2, 2) == 4);       // #2 throws on error

	if (add(2, 2) != 4)
		BOOST_ERROR("Ouch...");          // #3 continues on error

	if (add(2, 2) != 4)
		BOOST_FAIL("Ouch...");           // #4 throws on error

	if (add(2, 2) != 4) throw "Ouch..."; // #5 throws on error

	BOOST_CHECK_MESSAGE(add(2, 2) == 4,  // #6 continues on error
		"add(..) result: " << add(2, 2));

	BOOST_CHECK_EQUAL(add(2, 2), 4);	 // #7 continues on error
}

BOOST_AUTO_TEST_SUITE_END()
```
Preferred command line execution to get all information (```--log_level=all``` or ```--log_level=success```):
```BASH
<test executable> --log_format=XML --log_sink=<filename> --log_level=all --report_level=no
```
Resulting Unit Test XML report:
```XML
<TestLog>
[...]

Issue #368, opened 2015-01-07
Boost Unit Test, xunit.provideDetails=true, AST-based lookup


guwirth:
Played around with new xunit implementation and boost unit tests. I'm not able to get detailed information. Initial problem seems to be to create a boost unit test report with file information. 

Has someone a working setup?

wenns:
The AST-based detection doesnt work?

guwirth:
> The AST-based detection doesnt work?

Seems not. Please have a look to the example below. What is the AST-based detection looking for in this concrete example?

Boost Unit Test report:
```XML
<TestLog>
  <TestSuite name="Level1">
    <TestSuite name="Level2">
      <TestCase name="Test1">
        <TestingTime>856000</TestingTime>
      </TestCase>
      ...
```
after transformation:
```XML
<testsuite tests="86" errors="0" failures="0" name="MergedTestSuite" skipped="0">
  <testcase classname="Level1.Level2." name="Test1" time="0.856"/>
  ...
```

wenns:
Doesnt surpise me. Look at the value of "classname". For the AST-based lookup to work, the classname should contain a name of a class. Unqualified, AFAIK. Maybe the original report does contain this data and you can change the transformation layer to produce better fitting junit reports.

If this doesnt work, I would inject 'filename' attributes by custom scripting. Most probably you would need to provide those script with the necessary knowledge about filepaths sand maintain this knowledge as the according code base evolves.

guwirth:
* Means you are searching for *classname::name*?
 * Seams the current xlst is adding always a . at the end of classname.
* The current xlst use a . for nested testsuites which should be ::?
 * Are you supporting namespaces in classname *namespace::classname::name*?

Looks like the current xlst does not fit anymore?

wenns:
> Means you are searching for classname::name?

The test that I've written use simple unqualified classnames, without the leading :: or other namespace. And they do work. Looking at the pattern:
```C++
  static Pattern classNameMatchingPattern = Pattern.compile("(?:\\w*::)*?(\\w+?)::\\w+?:\\d+$");
```
Namespaces should be supported, too. 

> Looks like the current xlst does not fit anymore?

It probably never did. At the time as virtual files were used it didnt matter.

guwirth:
Ok not supported currently. Seems to be more work. 

* parser must be able to read boost code - not sure if this is the case yet
* test-cases are in namespaces not classes
* current boost test xsl creates invalid xunit xml

[...]

PR #387, opened 2015-01-15
Integration/boosttest

guwirth:
* add initial files to setup an integration test for boost and boost unit tests
* currently only files without build & check

### DIFF
--- a/integration-tests/features/test_execution_statistics.feature
+++ b/integration-tests/features/test_execution_statistics.feature
@@ -258,7 +258,7 @@ Feature: Providing test execution numbers
               | test_success_density | 0     |
               | test_execution_time  | 3     |
 
-  @skip
+
   Scenario Outline: Importing unchanged boosttest reports in detailed mode (filename tag)
 
       Testcases in boosttest reports with setting 'log_level=all' also
@@ -289,7 +289,7 @@ Feature: Providing test execution numbers
       # with nested testsuites
       | btest_test_nested-all.xml          | 4, 0, 4, 0, 0, 5                   |
 
-  @skip
+
   Scenario Outline: Importing unchanged boosttest reports in detailed mode (AST)
 
       Testcases in boosttest reports with setting 'log_level=test_suite'
@@ -318,8 +318,8 @@ Feature: Providing test execution numbers
       # with nested testsuites
       | btest_test_nested-test_suite.xml   | 4, 0, 4, 0, 0, 5                   |
 
-    @skip
-    Scenario: Test with real boost test framework
+
+  Scenario: Test with real boost test framework
 
         Real boost test framework is a very complex usecase for preprocessor and parser.
         Test if plugin is able to handle this.
--- a/integration-tests/testdata/boosttest_project/invalid_report.xml
+++ b/integration-tests/testdata/boosttest_project/invalid_report.xml
@@ -1,0 +1,5 @@
+trash
+trash
+trash
+trash
+trash


wenns:
>1200 files for a sample project? Its way too big for my taste. Can we get it smaller?

jmecosta:
Would you consider moving the integration test data into a sub module?

guwirth:
;-> that's boost.

Boost test framework is on top of boost. I already removed a lot of directories. Next step would be to cleanup within the directories. Problem I see is that this is a lot of work each time we upgrade to a new boost version. Will try to find out touched files with sysinternal tools...

wenns:
How about installing it via
```
$ sudo apt-get install libboost-test-dev
```

wenns:
> Would you consider moving the integration test data into a sub module?
I dont think its necessary, at least at this point of time. I think we can manage the problem otherwise.

guwirth:
Will try with sudo. Do you know to which location boost will be installed? Are there afterwards environment variables I can use? Someone experience with all that stuff?

wenns:
Just to be clear: this command line would work on Ubuntu and Debian derivates and looks differently with other package managers. My point is just that: you can install boost test on the system, just like e.g. cppcheck. I.e.:
* tests, build scripts, scripts for running the tests come into the sample project
* libs, headers are installed on the system. You need it to only to develop the tests anyway. Once development is ready, you just checkin the ready XML reports and run the tests against them. 

As for locations of the libs and headers etc.: on Linux they just land in standard locations (headers in /usr/include, libs in /usr/lib) and it just works. To make building work cross platform, one can just keep (and change, if necessary) in constants inside of build files.

guwirth:
@wenns : would like to write the tests now. 
* Should I add it to test_execution_statistics.feature or is it better to split it google-test/boost-test?
* What is the entry point? Just add a new .feature file?

wenns:
Please add new scenarios to the test_execution_statistics.feature. Your test cases are the same feature, after all. Behave's execution procedure is like so:
* behave knows the folder where to look for features via .behaverc
* the according step implementations (the python code) is found via convention
* the python code in steps is imported, analysed and put into the library of step implementations
* the features are read in, the step names parsed, according step implementations lookuped and called

guwirth:
@wenns tried to add a first test to *test_execution_statistics.feature* but it's not executed. See nothing in the log file on the Travis server. Do you have any idea?

.travis.yml
```diff
@@ -4,7 +4,10 @@ jdk:
   - oraclejdk8
   - openjdk7
 
-
+  
+before_install:
+        - sudo apt-get install libboost-dev libboost-test-dev
```
jmecosta: this will not work on windows. or on any non deb base system... isnt there a better way of handling these...

guwirth:
> sudo apt-get install libboost-dev libboost-test-dev

@jmecosta: there is also the possibility
* to download the zip and extract it
* to mock the boost unit test header with some simple defines

@wenns, @jmecosta:
But maybe some word what I like to do: My main focus is to add new tests for boosttest, here I like to test the filename way and the AST-based way. To test the AST-based way I need the header files to see if the plugin is able to parse boost headers.
At the moment my main interest is to add all needed test and to bring it into operation. This boostlib thing is a priority B point which could be solved at the end.

.travis.yml
wenns: The .travis.yml will be executed on a Travis VM, which is garanteed to be a debian derivate ;)
Having said this, the line is not actually needed, as the behave will only run 'sonar-runner' against checked in reports, no?
The boost header, libs, etc. are needed only for developing the tests, i.e. on your workstation. Not on the CI slave.

So not probs here

wont be executed on Windows

jmecosta: Of course, did not notice that was in the yaml file.. Good as it is :)

guwirth:
> Remove 'Outline' and it will execute. 

:-< copied from the wrong example: 'Scenario: googletest report is invalid' is also wrong

> The boost header, libs, etc. are needed only for developing the tests

Still not sure about that. Like to test AST based approach for this I need the header files otherwise unit tests can't be parsed in the right way?

wenns:
> :-< copied from the wrong example: 'Scenario: googletest report is invalid' is also wrong

Ahh. Damn. You're right!

>> The boost header, libs, etc. are needed only for developing the tests
> Still not sure about that. Like to test AST based approach for this I need the header files otherwise 
> unit  tests can't be parsed in the right way?

[Had a look into the test files] Oh yes, looks like...

guwirth:
- adapt boost xsl to support changed xunit feature: support all log_level settings
- extend SourceCodeBuilderVisitor for functions to return qualified names including namespaces
- extend xunit lookup table: supports search for qualified and unqualified classname now
- add a new unit test to search for qualified classname with namespace
- use 'sudo apt-get install libboost-dev libboost-test-dev' for test with real boost
- add boost test examples
- add boost test reports
- add new scenarios to integration tests:
  - add 'Scenario Outline: boosttest reports cannot be found or are empty'
  - add 'Scenario: Importing nested boosttest report in default mode'
  - add 'Scenario: Importing unchanged boosttest reports in default mode'
  - add 'Scenario Outline: Importing unchanged boosttest reports in detailed mode (filename tag)'
  - add 'Scenario Outline: Importing unchanged boosttest reports in detailed mode (AST)'
  - add 'Scenario: Test with real boost test framework'

- close #401
- close #372
- close #368
- close #375

cxx-squid/src/main/java/org/sonar/cxx/CxxAstScanner.java
```diff
@@ -124,7 +125,18 @@ public SourceCode createSourceCode(SourceCode parentSourceCode, AstNode astNode)
           sb.append(token.getValue());
         }
         String functionName = sb.toString();
-        SourceFunction function = new SourceFunction(functionName + ":" + astNode.getToken().getLine());
```
wenns: This ...

cxx-squid/src/main/java/org/sonar/cxx/CxxAstScanner.java
```diff
@@ -209,4 +221,23 @@ public SourceCode createSourceCode(SourceCode parentSourceCode, AstNode astNode)
     return builder.build();
   }
 
+  // Concatenate two strings, but if there is overlap at the intersection,
```
wenns: ... and this looks somewhat worrisome. Can you explain the reason for doing such gymnastics here ? :)

guwirth: For sure ;->

If you are inside of a namespace/class it is allowed to do a definition with the namespace/class again. This function ensures that namespace/class is not doubled. 

 namespace example {
     class A {
          void f1();
          void f2();
          void f3() {} // example::A::f3
          void A::f4() {} // example::A::f4
      };
     void example::A::f1() {} // example::A::f1
     void A::f2() {} // example::A::f2
 }

cxx-squid/src/main/java/org/sonar/cxx/CxxAstScanner.java
guwirth: This part is creating the qualified name including the namespaces. Going from current node upwards the tree adding all namespace names.

wenns:
The tests dont run on my machine (Linux 32 bit). wth? They did pass Travis and they obviously passed on your machine. But I get a bunch of:
```
19:09:57.386 DEBUG - Performing the 'filename'-based lookup using the value 'tests/unittests/test_component1.cc'
19:09:57.386 WARN  - ... no resource found, the testcase 'my_module:foo_successfull/foo_successfull' has to be skipped
```
where the lookup is expected to succeed. What can cause this? The stuff I merged in the meanwhile? Replacement of the deprecated ModuleFileSystem? Hmm...

guwirth:
:-< rebase to see if last changes are the issue

guwirth:
Fails after rebase, so last changes on master are the root cause.

guwirth:
seems to be changed behaviour in ```DefaultResourceFinder.findInSonar```?

guwirth:
@Typz, @wenns original search in fs.testDirs() is missing now?
Bug or feature?

Old:
```Java
 public org.sonar.api.resources.File findInSonar(File file, SensorContext context, ModuleFileSystem fs, Project project) {
    org.sonar.api.resources.File unitTestFile = org.sonar.api.resources.File.fromIOFile(file, project);
    if (unitTestFile == null) {
      unitTestFile = org.sonar.api.resources.File.fromIOFile(file, fs.testDirs());
    }
    if (context.getResource(unitTestFile) == null) {
      unitTestFile = null;
    }
    return unitTestFile;
  }
```
New:
```Java
 public org.sonar.api.resources.File findInSonar(File file, SensorContext context, FileSystem fs, Project project) {
    org.sonar.api.resources.File unitTestFile = org.sonar.api.resources.File.fromIOFile(file, project);
    if (context.getResource(unitTestFile) == null) {
      unitTestFile = null;
    }
    return unitTestFile;
  }
```

wenns:
Yep, saw that already. The point is, the older version had two access trials because of < 4.0 compatibility issue of the like... The new one should work too. Will try to investigate further.

guwirth:
http://javadocs.sonarsource.org/4.5/apidocs/org/sonar/api/resources/File.html

public static File fromIOFile(File file,  Project module)
Deprecated. since 4.5 use FileSystem.inputFile(org.sonar.api.batch.fs.FilePredicate)

wenns:
We're still using 4.2 API.

guwirth:
@wenns are you looking for a solution?

wenns:
Right now Im cleaning rooms ;)
But I want to invest an hour or two this weekend. Are you?

guwirth:
Have a time slot between 4 and 6 o'clock. If you want I can do another try...

wenns:
Lets make it that way: I'll just merge the stuff and skip failing tests. And we'll address the problem on the master. I mean: the PR is OK, its master thats broken. This PR just exposes a problem.

wenns:
Pushed to master. Failing tests are marked with @skip. If you want to run them, remove this annotation.

PR #392, opened 2015-01-19
new check BooleanEqualityComparisonCheck

Bertk:
Please merge only:
- LiteralBooleanEqualityComparisonCheck  
- Fix quality flaws of BooleanEqualityComparisonCheck 

and ignore others


cxx-checks/src/main/resources/org/sonar/l10n/cxx/rules/cxx/BooleanEqualityComparison.html
```diff
@@ -0,0 +1,18 @@
+<p>
+Boolean expressions should not be compared against boolean literals, as their value can be directly used.
+</p>
+
+<p>The following code:</p>
+
+<pre>
+if (booleanVariable == true) { /* ... */ }         // Non-Compliant
```
guwirth: Better is anyway if ( true == var ). Is this also working?

PR #394, opened 2015-01-20
Bug/indentation check reports against previous case

jmecosta:
to demonstrate the issue... what do you guys think


guwirth:
We are not using such warnings, we automatically format with tools like [Artistic Style](http://astyle.sourceforge.net/) => so no interest in this check.

wenns:
I do not need it too. My plan is to evaluate cpplint for style. And to use it on devs sandbox, on CI and in pre-commit hooks.

wenns:
Discarding: there seems to be not much demand.

PR #397, opened 2015-01-22
Metrics for methods

ametselaar:
This changeset saves the McCabe complexity and the number of lines of code for each function or method and adds a new metric parameter count, the number of parameters a functions takes.
A resource Method is created in sonar-cxx to have a place to save these metrics for each function or method.


Bertk:
This 2 new checks should also be considered for the technical debt. Please add properties for the two check rules in cxx-model.xml similar to the following:
```
           <chc>
                <rule-repo>cxx</rule-repo>
                <rule-key>TooManyLinesOfCodeInFile</rule-key>
                <prop>
                    <key>remediationFunction</key>
                    <txt>linear</txt>
                </prop>
                <prop>
                    <key>remediationFactor</key>
                    <val>5</val>
                    <txt>mn</txt>
                </prop>
                <prop>
                    <key>offset</key>
                    <val>0</val>
                    <txt>d</txt>
                </prop>
            </chc>
```

Bertk:
I like this checks especially the parameter count check but I removed them because the PR does not work with SQ 5.1 (exception in CxxSquidSensor.java).

guwirth:
Maybe we can try this one:
https://github.com/SonarCommunity/sonar-python/blob/master/python-checks/src/main/java/org/sonar/python/checks/TooManyParametersCheck.java

guwirth:
Outdated

PR #398, opened 2015-01-22
include a buildlog parser to retrieve build configuration

jmecosta:
@wenns @guwirth @Bertk 

earlier ive mention that we run into a problem with nuget packages, include paths keep being changed because nuget handles those. i said that we could create a parser to handle visual studio project files. but i tough this is a more elegant form of setting up the plugin... 

how does this work:

user builds solution like:
* msbuild solution.sln /v:Detailed > buildlog.txt
* sets in properties we have a new property: sonar.cxx.compilationBuildLog=buildlog.txt
* runs sonar-runner

includes and defines are processed from the buildlog so no need for additional user input...

the current solution does not touch to much on the include import implementation, but for the future it would be nice to only use the include/defines that are actually used per compilation unit. now all includes are initially stored leading likely to slow down builds.

ive include another property: sonar.cxx.compilationFormat = default now vc++, but if needed can be used to use with cmake or other build system

would this be something usefull for you guys

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
@@ -19,18 +19,30 @@
  */
 package org.sonar.cxx;
 
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.jfree.util.Log;
+import org.slf4j.LoggerFactory;
 
 import org.sonar.squidbridge.api.SquidConfiguration;
 
 public class CxxConfiguration extends SquidConfiguration {
 
+  private static final org.slf4j.Logger LOG = LoggerFactory.getLogger("CxxConfiguration");
+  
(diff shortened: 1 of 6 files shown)


guwirth:
In general I think this is a very useful feature. Most trouble we have is also with setting up macros and includes. I'm only not sure how good 'the output' of the log file is. Typical problems we have with such tools or interpreting the projects:
* There are implicit 'default' settings you don't see listening to command line or parse a log.
* There could be 'environment variables' in the path (from system, VS or libraries).
* Settings are sometimes in property sheets where you derive from, this %(AdditionalXXX). Property sheets could be nested.
* Some macros depend on command line options and are not in the preprocessor definitions (e.g. /EH (Exception Handling Model) defines/undefines _CPPUNWIND)

@jmecosta Don't know if your solution is able to support all these things?

jmecosta:
not sure i understand... the buildlog, given that you specify /v:Detailed, will contain the exact cl.exe command used to build a obj file. So if the compilation succeeds the correct includes/defines will be in the log. All things you mention have been already processed by msbuild and the output of this will only contain the needed included  and defines. plus build options (but those i have not considered because we dont support them). property sheets at this point are already translated into human readable form if not mistaken...



environment variables are also taken, so INCLUDES env is also taken into account.

jmecosta:
Or maybe there is more stuff needed, but anyway the parser can be extended to retrive more stuff.  We can even enable /v:diagnostic to get more stuff ... I still need to try this in our big projects

jmecosta:
https://github.com/wenns/sonar-cxx/wiki/_history

documented here : b2352d4

PR #399, opened 2015-01-22
include a buildlog parser to retrieve build configuration

jmecosta:


guwirth:
@jmecosta thanks for providing this. Could you say some words how this is working or is intended to work please.

guwirth:
Sorry found #398

Bertk:
I will check this PR this weekend and it will definitily reduce the maintenance effort for us.

Meanwhile I analyzed the sources and found this issue:
**Found reliance on default encoding in org.sonar.cxx.CxxConfiguration.parseVCppLog(File): new java.io.FileReader(File)** 
*Found a call to a method which will perform a byte to String (or String to byte) conversion, and will assume that the default platform encoding is suitable. This will cause the application behaviour to vary between platforms. Use an alternative API and specify a charset name or Charset object explicitly.*

Bertk:
Shouldn't the configuration items be moved from **General** to the **Compiler warnings** and we might only need a Boolean to enable this feature.

At the moment I use the following settings:
```
sonar.cxx.compilationBuildLog=compiler.log
sonar.cxx.compiler.reportPath=compiler.log
sonar.cxx.compiler.charset=UTF-8
```

jmecosta:
Yes that's good idea, I forgot about the compile property. I will change.
Is the property also supporting make?

Bertk:
Today we only distinguish the report formats (sonar.cxx.compiler.parser: Visual C++, GCC) and there is no information for nmake, cmake or MS-build. The log file generated by MS-build (Visual C++) has all the information but I do not know the latest make tools.

Bertk:
There is a issue with the include path parsing.
```
19:28:00.546 ERROR - Cannot parse include path using element '/I"C:\Program' : 'The filename, directory name, or volume label syntax is incorrect'
``` 
The build log file has entries like:
```
/I"C:\Program Files (x86)\Windows Kits\8.1\include\um" 
/I"C:\Program Files (x86)\Windows Kits\8.1\include\shared" 
/I"C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\atlmfc\include" 
/I"C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include"
```
By the way, I did some updates but did not fix the issue above - see https://github.com/Bertk/sonar-cxx/tree/msbuild-buildlog-parser

jmecosta:
thanks, ive moved you changes there... excepted for the code you use to try to fix the issue. i will fix a later this week...

jmecosta:
@Bertk ive now reuse the sonar.cxx.compiler.reportPath so only that value is needed

Bertk:
@jmecosta I still get the error messages and the parsing of include folder definition fails. The following code part fails to handle the string ```/I"C:\Program Files (x86)\Windows Kits\8.1\include\um"``` correctly.
```
  private void parseVCppCompilerCLLine(String line, String projectPath) {
    File file = new File(projectPath);
    String project = file.getParent();
    String[] elems = line.split("\\s+");
    for (int i = 0; i < elems.length; i++) {
      if (elems[i].startsWith("/I")) {        
        ParseInclude(elems[i], project);
      }
```
So I guess the regular expression needs to be enhanced and for includes a combination of 1) and 2) might work. 
1) ```/I".*?"```
2) ```/I.*?\s```

jmecosta:
can you check with

Author:		Jorge Costa <Jorge.Costa@tekla.com>
Date:		1 hour ago (28.1.2015 20:14:12)
Commit hash:	42884c7943a42fc635f1e2395197ea22543c8cf5
Parent(s):	9fc1346dda


but i guess i can try to put those regular expression (dont understand much about those)

Bertk:
I will Check it tomorrow

Bertk:
@jmecosta Please check the solution with the regular expression. 

```
  private void parseVCppCompilerCLLine(String line, String projectPath) {
    File file = new File(projectPath);
    String project = file.getParent();

    for (String includeElem : getMatches(Pattern.compile("/I\"(.*?)\""), line)) {
      ParseInclude(includeElem, project);
    }

    for (String includeElem : getMatches(Pattern.compile("/I([^\\s\"]+) "),
        line)) {
      ParseInclude(includeElem, project);
    }

    for (String macroElem : getMatches(Pattern.compile("[/-]D\\s([^\\s]+)"),
        line)) {
      ParseMacro(macroElem);
    }
  }

  private List<String> getMatches(Pattern pattern, String text) {
    List<String> matches = new ArrayList<String>();
    Matcher m = pattern.matcher(text);
    while (m.find()) {
      matches.add(m.group(1));
    }
    return matches;
  }

  private void ParseInclude(String element, String project) {
    try {
      File includeRoot = new File(element.replace("\"", ""));
      String includePath = "";
      if (!includeRoot.isAbsolute()) {
        includeRoot = new File(project, includeRoot.getPath());
        includePath = includeRoot.getCanonicalPath();
      } else {
        includePath = includeRoot.getCanonicalPath();
      }
      if (!uniqueIncludes.contains(includePath)) {
        uniqueIncludes.add(includePath);
      }
    } catch (java.io.IOException io) {
      LOG.error("Cannot parse include path using element '{}' : '{}'", element,
          io.getMessage());
    }
  }

  private void ParseMacro(String macroElem) {
    macroElem.replace("=", " ");
    if (!uniqueDefines.contains(macroElem)) {
      uniqueDefines.add(macroElem);
    }
  }
```

jmecosta:
I will @Bert, i just did havent got to time to do it...

jmecosta:
@Bertk ive added now your code, tests failed so ive added a new condition into one of the defines extractor... seems to work

... So far in our projects this works really nicely. no the warnings we get from cannot find include are actually problems with the build files... do we have a way of log a issue at project level to report these in the dashboard...?

Bertk:
@jmecosta Please check the *duplicateIncludeRule* which is used by *DependencyAnalyzer* to create issues. The rule defined within cxx.checks and DependencyAnalyzer.java creates the issue. The build configuration problem could be handled in a similar manner.

jmecosta:
am happy with this, ive merge master and solve conflicts. the creation of the issue needs to wait as the suggested approach requires a resource that we dont have when buildlog is being parsed

cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
```diff
@@ -142,4 +165,132 @@ public void setMissingIncludeWarningsEnabled(boolean enabled){
...
+            }
+          }
+          
+          // get base path of project to make 
+          // Target "ClCompile" in file "C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V120\Microsoft.CppCommon.targets" from project "D:\Development\SonarQube\cxx\sonar-cxx\integration-tests\testdata\googletest_bullseye_vs_project\PathHandling.Test\PathHandling.Test.vcxproj" (target "_ClCompile" depends on it):
+          if (line.startsWith("Target \"ClCompile\" in file")) {
+            currentProjectPath = line.split("\" from project \"")[1].split("\\s+")[0].replace("\"", "");              
+          }
+          
+          if (line.contains("C:\\Program Files (x86)\\BullseyeCoverage\\bin\\CL.exe")) {
```
guwirth: Call via BullsEyeCoverage? What if a other tool is used to create coverage?

cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
```diff
@@ -142,4 +165,132 @@ public void setMissingIncludeWarningsEnabled(boolean enabled){
...
+            parseVCppCompilerCLLine(line, currentProjectPath);
+          }
+          
+          if (line.contains("C:\\Program Files (x86)\\Microsoft Visual Studio 10.0\\VC\\bin\\CL.exe") ||
```
guwirth: Hard coded paths. Installation path of VS could be different?

cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
```diff
@@ -142,4 +165,132 @@ public void setMissingIncludeWarningsEnabled(boolean enabled){
...
+            parseVCppCompilerCLLine(line, currentProjectPath);        
+          }  
+        }
+        br.close();
+      } catch (IOException ex) {
+        LOG.error("Cannot parse build log", ex);
+      }
+  }
+
+  private void parseVCppCompilerCLLine(String line, String projectPath) {
```
guwirth: Not sure if this approach is really working. In VS each source file could have own settings. My understanding of this function is that you are collecting the settings of all files which could be wrong.

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/compiler/CxxCompilerSensor.java
```diff
@@ -47,6 +47,7 @@
   public static final String REPORT_CHARSET_DEF = "sonar.cxx.compiler.charset";
   public static final String PARSER_KEY_DEF = "sonar.cxx.compiler.parser";
   public static final String DEFAULT_PARSER_DEF = CxxCompilerVcParser.KEY;
+  public static final String DEFAULT_CHARSET_DEF = "UTF-16";
```
guwirth: Why is default UTF-16? I think it is UTF-8.

cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
```diff
@@ -142,4 +165,132 @@ public void setMissingIncludeWarningsEnabled(boolean enabled){
...
+    for (String includeElem : getMatches(Pattern.compile("/I\"(.*?)\""), line)) {
+      ParseInclude(includeElem, project);
+    }
+
+    for (String includeElem : getMatches(Pattern.compile("/I([^\\s\"]+) "),
+        line)) {
+      ParseInclude(includeElem, project);
+    }
+
+    for (String macroElem : getMatches(Pattern.compile("[/-]D\\s([^\\s]+)"),
```
guwirth: * There are implicit 'default' settings you don't see listening to command line or parse a log.
* Some macros depend on command line options and are not in the preprocessor definitions (e.g. /EH (Exception Handling Model) defines/undefines _CPPUNWIND)

cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
jmecosta: Dont know of other coverage tools that intercept the compiler the way
bullseye does. But once users use this they will raise issues and we can
support other interceptors

cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
jmecosta: Yes this is correct, the next step is to implement settings per file based
approach. But will not have time until the end of the month to get this .
so we can leave it for next major release. Later once this is on, we can
support #421 more easily

cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
jmecosta: perhaps can be made by only checking "bin\\CL.exe" is found. Should simplify a bit the code... i try and see if test are ok

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/compiler/CxxCompilerSensor.java
jmecosta: @Bertk i think that one is yours, i had UTF-8 but i change to UTF-16 for some reason i cannot remember anymore

cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
jmecosta: i think those we need to implement one by one. once issues are found. (if there are really issues). usually users when manually defining those they do not bother with them either, so if there is a concrete specification and a parsing error that occurs we can then fix it... but i would go and implement all kind of options at this moment since there are 2 many vc++ compilers to support.

cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
jmecosta: fixed

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/compiler/CxxCompilerSensor.java
Bertk: UTF-16 is required for the VS2010 html reports. VS2010 was the first version for VC++ reports support and the compiler command syntax did not change.
Using UTF-8 for the buildlog Parser is OK but we should be consistent and also update https://github.com/wenns/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/compiler/CxxCompilerVcParser.java

jmecosta: Ok I will update then on both sides to use utf8

cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
guwirth: Below an Python example how to do this. That's maybe a good starting point:

```Python
# VS 2010 predefined macros:
# http://msdn.microsoft.com/en-us/library/vstudio/b0084kay(v=vs.100).aspx

def macro_expansion(line,logobj,target):
    vs_macros = {}
    
    if target=='64':
        vs_macros.update({'__ptr32=':''})
    
    if '/GR-' not in line:
        vs_macros.update({'_CPPRTTI':''})
    
    if '/EHs ' in line:
        vs_macros.update({'_CPPUNWIND':''})
    elif '/EHa ' in line:
        vs_macros.update({'_CPPUNWIND':''})
    elif '/EHsc ' in line:
        vs_macros.update({'_CPPUNWIND':''})
    elif '/EHac ' in line:
        vs_macros.update({'_CPPUNWIND':''})
        
    if '/MD ' in line:
        vs_macros.update({'_MT':''})
        vs_macros.update({'_DLL':''})

    if '/MDd' in line:
        vs_macros.update({'_DEBUG':''})
        vs_macros.update({'_MT':''})
        vs_macros.update({'_DLL':''})
     
    if '/MT ' in line:
        vs_macros.update({'_MT':'0'})

    if '/MTd' in line:
        vs_macros.update({'_DEBUG':''})
        vs_macros.update({'_MT':''})

    if '/LD ' in line:
        vs_macros.update({'_MT':''})
        
    if '/LDd' in line:
        vs_macros.update({'_MT':''})
        vs_macros.update({'_DEBUG':''})
        
    if '/clr' in line:
        vs_macros.update({'_M_CEE':''})
        vs_macros.update({'_MANAGED':''})

    if '/clr:pure' in line:
        vs_macros.update({'_M_CEE_PURE':''})

    if '/clr:safe' in line:
        vs_macros.update({'_M_CEE_SAFE':''})
    
    if target=='32':
        if '/arch:IA32' in line:
            vs_macros.update({'_M_IX86_FP':'0'})
        elif '/arch:SSE' in line:
            vs_macros.update({'_M_IX86_FP':'1'})
        elif '/arch:SSE2' in line:
            vs_macros.update({'_M_IX86_FP':'2'})
        elif '/arch:AVX' in line:
            vs_macros.update({'_M_IX86_FP':'2'})
            vs_macros.update({'__AVX__':''})
        elif '/arch:AVX2' in line:
            vs_macros.update({'_M_IX86_FP':'2'})
            vs_macros.update({'__AVX__':''})
            vs_macros.update({'__AVX2__':''})
        else:
            vs_macros.update({'_M_IX86_FP':'0'})
        
    if '/Ze' not in line:
        vs_macros.update({'_MSC_EXTENSIONS':'1'})
        
    if '/RTC' in line or '/GZ' in line:
        vs_macros.update({'__MSVC_RUNTIME_CHECKS':''})
        
    if '/Zc:wchar_t' in line:
        vs_macros.update({'_NATIVE_WCHAR_T_DEFINED':''})
        vs_macros.update({'_WCHAR_T_DEFINED':''})

    if '/Wp64' in line:
        vs_macros.update({'_Wp64':''})
        
    if '/Zl' in line:
        vs_macros.update({'_VC_NODEFAULTLIB':''}) 
        
    if '/G5' in line:
        vs_macros.update({'_M_IX86':'500'})
    elif '/G6' in line or '/GB' in line:
        vs_macros.update({'_M_IX86':'600'})
    elif '/G3' in line:
        vs_macros.update({'_M_IX86':'300'})
    elif '/G4' in line:
        vs_macros.update({'_M_IX86':'400'})
        
    return vs_macros
```

jmecosta: Ok good I try to include this, later I need to test this in our 2013 build.
But this should be a good start definitely

jmecosta:
@guwirth ive added a initial skeleton for the 2010, 2012 and 2013 default macros. let me know if this is ok so i can complete the work

guwirth:
Looks good. 
- not sure if we need different macros for vs10, vs12, ...
- AddMacro should support with and without value.

jmecosta:
There are some differences between 10 and 12. Not that many so should not
make much difference to include them. This way its clear about the
evolution of the compiler options. I will then push the changes today

guwirth:
Ok makes sense. 

Two more questions:
- could you describe VS settings to create buildlog please. Needed for docu. 
- is it possible to read more than one buildlog? In our setup we have sub folders with projects. 

> Am 11.04.2015 um 10:33 schrieb Jorge Costa <notifications@github.com>:
> 
> There are some differences between 10 and 12. Not that many so should not
> make much difference to include them. This way its clear about the
> evolution of the compiler options. I will then push the changes today
>

jmecosta:
This reuses the already document visual studio import issues section. It
needs only a update there to mention includes and defines are also taken
when importing vs issues. But yes, I will update that.

For several build logs, this is not possible. But I guess you can glue them
together as a pre build step.

jmecosta:
actually it supports several buildlogs. this was also reused from the vs issue thingy

jmecosta:
If the tests are fine I guess we can use this version, there is some
context specific stuff that I did not included but we can think of them
later

cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
```diff
@@ -142,4 +198,37 @@ public void setMissingIncludeWarningsEnabled(boolean enabled){
...
+        LOG.debug("Parse build log OK: includes: '{}' defines: '{}'", uniqueIncludes.size(), uniqueDefines.size());
+      } else {
+        LOG.error("Compilation log not found: '{}'", buildLog.getAbsolutePath());
+      }    
+    }
+    
+    RaiseIssuesForNotFoundIncludes();
+  }
+
+  private void RaiseIssuesForNotFoundIncludes() {
```
guwirth: What is the idea of this?

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
```diff
@@ -0,0 +1,484 @@
...
+            }
+          }
+          
+          // get base path of project to make 
+          // Target "ClCompile" in file "C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V120\Microsoft.CppCommon.targets" from project "D:\Development\SonarQube\cxx\sonar-cxx\integration-tests\testdata\googletest_bullseye_vs_project\PathHandling.Test\PathHandling.Test.vcxproj" (target "_ClCompile" depends on it):
+          if (line.startsWith("Target \"ClCompile\" in file")) {
+            currentProjectPath = line.split("\" from project \"")[1].split("\\s+")[0].replace("\"", "");              
+          }
+          
+          if(line.contains("\\V100\\Microsoft.CppBuild.targets")) {
```
guwirth: This is if ... else if?

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
```diff
@@ -0,0 +1,484 @@
...
+      AddMacro(macroElem, fileElement);
+    }
+    
+    // https://msdn.microsoft.com/en-us/library/vstudio/b0084kay(v=vs.100).aspx
+    // https://msdn.microsoft.com/en-us/library/vstudio/b0084kay(v=vs.110).aspx
+    // https://msdn.microsoft.com/en-us/library/vstudio/b0084kay(v=vs.120).aspx 
+    // https://msdn.microsoft.com/en-us/library/vstudio/b0084kay(v=vs.140).aspx
+    ParseCommonCompilerOptions(line, fileElement);
+    
+    if (platformToolset.equals("v100")) {
```
guwirth: This is if ... else if?

guwirth: Would use enum for constants (v100, ...)

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
```diff
@@ -0,0 +1,484 @@
...
+    AddMacro("__COUNTER__", fileElement);     
+    //__DATE__ The compilation date of the current source file. The date is a string literal of the form Mmm dd yyyy. The month name Mmm is the same as for dates generated by the library function asctime declared in TIME.H.
+    AddMacro("__DATE__", fileElement);
+    //__FILE__ The name of the current source file. __FILE__ expands to a string surrounded by double quotation marks. To ensure that the full path to the file is displayed, use /FC (Full Path of Source Code File in Diagnostics).
+    AddMacro("__FILE__", fileElement);
+    //__LINE__ The line number in the current source file. The line number is a decimal integer constant. It can be changed with a #line directive.
+    AddMacro("__LINE__", fileElement);
+    //__STDC__ Indicates full conformance with the ANSI C standard. Defined as the integer constant 1 only if the /Za compiler option is given and you are not compiling C++ code; otherwise is undefined.
+    AddMacro("__STDC__", fileElement);
```
guwirth: For cpp projects macro is not defined?

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
```diff
@@ -0,0 +1,484 @@
...
+    if (line.contains("/clr:pure ")) {
+      AddMacro("_M_CEE_PURE", fileElement);    
+    }    
+    //_M_CEE_SAFE Defined for a compilation that uses /clr:safe.
+    if (line.contains("/clr:safe ")) {
+      AddMacro("_M_CEE_SAFE", fileElement);    
+    }        
+    //__CLR_VER Defines the version of the common language runtime used when the application was compiled. The value returned will be in the following format:    
+    //__cplusplus_cli Defined when you compile with /clr, /clr:pure, or /clr:safe. Value of __cplusplus_cli is 200406. __cplusplus_cli is in effect throughout the translation unit.    
+    //_M_CEE Defined for a compilation that uses any form of /clr (/clr:oldSyntax, /clr:safe, for example).
```
guwirth: Maybe ' if (line.contains("/clr")' is sufficient?

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
```diff
@@ -0,0 +1,484 @@
...
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.slf4j.LoggerFactory;
+
+/**
+ *
+ * @author jocs
+ */
+public class CxxVCppBuildLogParser {
```
guwirth: Is it somewhere possible for the user to see which includes and macros are added by the parser? In this 'automatic' mode it would be useful to list it in the log file? What are you thinking?

guwirth:
@jmecosta Please tell me if things are done from your side. I would merge it then and would do further stabilization on trunk. For VS users this feature is really a great improvement!

cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
jmecosta: Havent finish this. The idea is to raise issues if a include path was used
that is invalid.

jmecosta:
Will do once I go through comments, perhaps tomorrow. Come down with some
stomach flu

guwirth:
Get well soon!

Von meinem iPhone gesendet

> Am 12.04.2015 um 11:42 schrieb Jorge Costa <notifications@github.com>:
> 
> Will do once I go through comments, perhaps tomorrow. Come down with some
> stomach flu
>

jmecosta:
This is ready, you might want to review the new rule. For printing includes per file . we need to redactor the parsing part to support included, defines per file. I can take a look in another pull request

guwirth:
Looks good. Could you rebase and push this PR please. Makes it easier for me to merge it.

jmecosta:
done

jmecosta:
https://github.com/wenns/sonar-cxx/wiki/_history

documented here : b2352d4

jmecosta:
https://github.com/wenns/sonar-cxx/wiki/_history

documented here : b2352d4

PR #400, opened 2015-01-22
Parse ">>" instead of "> >" in template declarations and c++ style casts...

jmecosta:
@wenns  accidentally push my master to upstream instead of origin (must have remotes mess, after windows reinstallation)... Redo the template thing, after reverted the merge commit...

Sorry about that

....

CxxGrammarImpl.java: Added enum items typeIdEnclosed,
    templateParameterListEnclosed, innerTypeParameter, innerTypeId
CxxGrammarImpl.java(expressions): Support ">>" instead of  "> >" in c++ style casts.
CxxGrammarImpl.java(templates): Support ">>" instead of "> >" in template declarations.
.../own/declarations.cc: Add some test case4s from issue #49

(cherry picked from commit df54eb14ca48204669fd38580fea3dde77a6f692)


wenns:
Ok, nevermind.

ametselaar:
You may wish to look at commit 907119414a3eccc932001e4c6e6c632cd1f00c9a, which helps parsing numeric expressions as template arguments, and which I have tested on the code from http://openmsx.sourceforge.net .

jmecosta:
OK nice I will cherry pick... Thanks

guwirth:
Good final test could be to test against latest boost version.

jmecosta:
yep that would be good

guwirth:
Maybe you can update .travis.yml file
current: - sudo apt-get install libboost-dev libboost-test-dev
new: - sudo apt-get install libboost1.55-dev libboost-test1.55-dev

jmecosta:
Sure, but why 1.55

guwirth:
You can also trry 1.57. Don't know which packages are available?

jmecosta:
my point was why not use libboost-dev libboost-test-dev

guwirth:
Because this is currently 1.46 which is not C++11.

jmecosta:
Ok that makes sense

wenns:
I dont particularly like this PR, as it is not a general solution for the problem at hand (which is an ambiguity in the grammar). It only solves the nesting-level-of-two-case. With a hack, which changes the shape of the AST. It isnt the fault of the PR author, though. This case is just hard. The standard suggests to resolve this ambiguity with semantic analysis (which we *not* gonna implement, of course)

jmecosta:
So what to do?

jmecosta:
OK got it however might be good idea to have a last release with ast and
support this

ametselaar:
While I admit that this pull request produces an ugly AST for constructs that are impossible to parse with the curreent stable version, I disagree with the assertion that it only solves the nesting-level-of-two case.
For deeper nesting levels, the parser only needs to cope with the tokens ">>" and ">" because the lexer will never produce tokens like ">>>" or ">>>>", which are not tokens in C++. 
I have seen parsing succeed on this nesting-level-of-five case:
```c++
template<typename T> struct Saver
        : if_<is_primitive<T>,
              PrimitiveSaver<T>,
          if_<serialize_as_enum<T>,
              EnumSaver<T>,
          if_<serialize_as_pointer<T>,
              PointerSaver<T>,
          if_<serialize_as_collection<T>,
              CollectionSaver<T>,
              ClassSaver<T>>>>> {};
```

guwirth:
@ametselaar Why not just adding some more examples with deeper nesting to test this? Would answer this question.

> The standard suggests to resolve this ambiguity with semantic analysis ...

Because this is the main blocker for C++11 compatibility I would also accept an "ugly" solution. The standard is here also ugly :-).

wenns:
@ametselaar: sorry if I got this wrong with the nesting level support. If this fixes all the other cases too, then I'll pull this thing. Maybe you can add a more complex sample to the tests, as @guwirth suggests?

guwirth:
@ametselaar Could you provide the cascaded example please. Then we can settle the matter....

cxx-squid/src/main/java/org/sonar/cxx/parser/CxxGrammarImpl.java
```diff
@@ -984,6 +996,10 @@ private static void declarators(LexerfulGrammarBuilder b) {
         b.firstOf("&", "&&")
         );
 
+    b.rule(restrictQualifier).is(
+        b.firstOf("restrict", "__restrict__", "__restrict")
```
guwirth: Found this restrict variants. Which should be done with preprocessor.

wenns: +1

wenns:
Merged with 2b35cd9f5bf84c9de8ad38cd1439e5b3eae1287e

Issue #427, opened 2015-02-18
xunit, filename tag: relative paths are not supported


guwirth:
**V0.9.3 alpha:**  After refactoring relative paths in filename tags seems to be no more supported.

https://github.com/wenns/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/xunit/DefaultResourceFinder.java
```Java
List<File> files = Lists.newArrayList(fs.files(fs.predicates().is(file)));
```
Also test cases seems to be missing?

@wenns You did the refactoring. Did you test this with relative paths in filename tag?

wenns:
Sorry for disappearing, there are times I dont have spare time for this project... Having a look at this.

guwirth:
checked it again:
```Java
fs.predicates().is(file)
```
is supporting absolute & relative paths

PR #428, opened 2015-02-21
simulate virtual files for xunit/provideDetails

guwirth:
- support relative path for filename tags
- add new boosttest-1.x-to-junit-dummy-1.0.xsl which set filename always to '.\cxx_xunit_dummy.cpp'
- add new integration test

### DIFF
--- a/integration-tests/features/test_execution_statistics.feature
+++ b/integration-tests/features/test_execution_statistics.feature
@@ -258,7 +258,30 @@ Feature: Providing test execution numbers
               | test_success_density | 0     |
               | test_execution_time  | 3     |
 
- 
+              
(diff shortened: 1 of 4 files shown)


guwirth:
close #427

wenns:
There must be something fishy here: Your test runs successfully with the production code changes (in DefaultResourceFinder) undone. Huh?

guwirth:
For me it seems

List<File> files = Lists.newArrayList(fs.files(fs.predicates().is(file)));

Is not supporting relative paths?

guwirth:
documented:
https://github.com/wenns/sonar-cxx/wiki/Supported-configuration-properties
https://github.com/wenns/sonar-cxx/wiki/Get-test-execution-metrics

PR #429, opened 2015-02-21
mark inconclusive messages in cppcheck V2

### LINKED ISSUE #247: Inconclusive issues from cppcheck not marked
The new version 0.9.1 is able to handle cppcheck XML format version 2 output so users may now see inconclusive messages (messages with a higher chance to be a false positive).

It would be nice if their status could be shown in the GUI.

guwirth:
- add [inconclusive] at the beginning of the message
- close #247


guwirth:
Was also wondering first. Background is that the reports in format V1 and V2 contain the same issues (file, line and msg). The plugin is filtering them and adds it only once. Starting with this feature there is additional text at the beginning of the message text (V1 without and V2 with inconclusive). Means they are different know.

guwirth:
documented: https://github.com/wenns/sonar-cxx/wiki/Upgrade-Instructions

PR #459, opened 2015-04-12
Fix hashhash related parsing problem

guwirth:
Let ## operator work on two tokens on the left when the token before the
last non-whitespace token is not whitespace and has no trivia.
Enable test cases.


guwirth:
Could someone else look to this PR please. Not sure if we should merge it?

PR #469, opened 2015-04-17
demo how to set default values for sqale

guwirth:
@todo: set the "right" Characteristic

@jmecosta: maybe you have more experience with this? Is it possible to define an own 'characteristic'?
I tested this and currently it is under 'Security > Errors'

### DIFF
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRuleRepository.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRuleRepository.java
@@ -58,8 +58,16 @@ public class CxxExternalRuleRepository implements RulesDefinition {
         }
       }
     }
-
-    //i18nLoader.load(repository); //@todo?
+    
+    for (NewRule rule : repository.rules()) {
+      rule
+        //FIXME: set internal key to key to ensure rule templates works properly : should be removed when SONAR-6162 is fixed.
+        .setInternalKey(rule.key())
+        // set default Technical Debt values for 'other' rules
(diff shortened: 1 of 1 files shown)


sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRuleRepository.java
```diff
@@ -58,8 +58,16 @@ public void define(Context context) {
...
     }
-
-    //i18nLoader.load(repository); //@todo?
+    
+    for (NewRule rule : repository.rules()) {
+      rule
+        //FIXME: set internal key to key to ensure rule templates works properly : should be removed when SONAR-6162 is fixed.
+        .setInternalKey(rule.key())
+        // set default Technical Debt values for 'other' rules
+        .setDebtSubCharacteristic("ERRORS")
```
jmecosta: i think we should not do this, setting these defaults values are going to cause more issues than not. Because most issues from other rules are not belonging any "default". Same goes for time to fix. adding default values like you are over or uderestimating everything that goes into sonar

I tought you would more including a parser for users to define their own values, imo that would be more usefull.

guwirth:
> I tought you would more including a parser for users to define their own values, imo that would be more usefull. 

Yes that is the thing I like additional to do.

Question with this demo was if something like this makes sense? In our case most of the issues come from 'other' which results in 'no' Technical Debt currently.

jmecosta:
I dont think it makes sense since even the categorization should be totally
off, there will be rules belonging to test category that will show under
what you define as default so people will need to change those anyway.

guwirth:
see #529 instead

PR #473, opened 2015-04-19
c++/cli projects contains weird path from temp folders, improve error ha...

jmecosta:
...ndling to report those as warnings in build log


jmecosta:
@guwirth some maintenance work, the original feature went broken after including the includes per file. It should now be good.

guwirth:
Could you rebase and squash that please. This time I will wait 😄.

jmecosta:
#487

Issue #477, opened 2015-04-19
parser error with const reference to std::string


Bertk:
Hi ,
I extended the samples within *cxx-squid\src\test\resources\parser\own* folder and added 2 class member functions (https://github.com/Bertk/sonar-cxx/commit/33f3d7c0ee016bc85308df3219e315bf7c52091b)
The first one passes the second fails. I observed this also in our legacy code many times. 
Do you have any idea how to fix this issue?

```
testParsingOnDiverseSourceFiles(org.sonar.cxx.parser.CxxParserTest)  Time elapsed: 0.016 sec  <<< ERROR!
com.sonar.sslr.api.RecognitionException: Parse error at line 17 column 46:

   10: ;
   11:
   12: void ClientContextRegistry::onChanged(const ClientIdentifier & aIdentifier_in)
   13: {
   14:  std::cout << aIdentifier_in.unique_id << std::endl;
   15: }
   16:
  -->  bool PropertyEdit::CheckForCharactersInString(const std::string& str_in)
   18: {
   19:  int aLen = str_in.length;
   20:  for (int i = 0; i<aLen; i++)
   21:  {

        at org.sonar.sslr.internal.vm.Machine.parse(Machine.java:74)
        at com.sonar.sslr.impl.Parser.parse(Parser.java:87)
        at com.sonar.sslr.impl.Parser.parse(Parser.java:72)
        at org.sonar.cxx.parser.CxxParserTest.testParsingOnDiverseSourceFiles(CxxParserTest.java:54)
```

guwirth:
Just a try: Solved with #484?

Bertk:
Yes, the test passed now but the parser error handling need some improvement  :wink:

guwirth:
Looking forward to see your improvement. 😄

PR #484, opened 2015-04-20
align grammar with version from @jmecosta

### LINKED ISSUE #479: Regression: Parser Error
The current 0.9.3 snapshot introduced a bug, 

this code cannot be parsed:

    namespace common
    {
       int cncl_set_progress_pos(int NumberOfBar, int PercentCompleted)
       {
           if(PercentCompleted < 0 || PercentCompleted > 100)
          {
          }
           return 0;
       }}

Removing 

        if(PercentCompleted < 0 || PercentCompleted > 100)
        {
        }

Or

        if(PercentCompleted < 0)
        {
        }

Then everything parses. 0.9.2 this was fine. Cannot understand why the error is reported in function line, as it seems the condition are the ones causing the parsing error

guwirth:
close #479


jmecosta:
yes this works now

guwirth:
close #477
close #479

PR #492, opened 2015-04-22
bugfix for pc-lint (#488)

### LINKED ISSUE #488: PC-Lint import stops with Exception
I try to import a PC-Lint report into SonarQube and it is working very fine but only until some points. I can see how the plugin is remapping the MISRA rules to the internal MISRA rules. But it hangs everytime on the same point and throws an exception: 

ERROR: Caused by: Cannot feed the data into sonar, details: 'java.lang.IllegalStateException: No match found'
ERROR: Caused by: No match found

What is this error message about? When is the plugin throwing this exception?

guwirth:
- improve pc-lint error handling
- extend parsing for MISRA rules, example:
 <result>
  <issue file ="somefile.c" line = "141" number = "524" desc = "Loss of precision (assignment) (float to unsigned int) [MISRA 2004 Rule 10.1]"/>
 </result>
- add unit test for new format
- close #488


guwirth:
@Bertk / @chrissy25 could you review and test please.

PR #497, opened 2015-04-24
fix #495

### LINKED ISSUE #495: VCCppBuildLogParser - resolve include folders fails for relative path
There might be an issue with the baseDir for the multi module projects and relative paths e.g. "/I..\..\..\boost\include". I use the sonar-visual-studio bootstrapper and the baseDir of the solution is used instead of the baseDir of the module.

A warning message like the following will be logged:
```
19:04:09.853 WARN  - the include root 'C:\boost\include' doesn't exist
```
My VC log file does not have "from Project" and this seams to be the reason. Anyway the vcxproj file is located in baseDir of the module and potential fix could look like this:
```
currentProjectPath = Paths.get(".");
// should be replaced by
currentProjectPath = fs.baseDir().getPath();
```
@jmecosta What do you think?

jmecosta:


jmecosta:
@Bertk  please review

Bertk:
Please check this commit 49de2b68afb6126faacd20549a716bf5058cf620 . This works on my multi module projects with Visual Studio bootstrapper. The issue is the assignment *Path currentProjectPath =Paths.get(".")* which will set currentProjectPath to the folder with the Visual Studio Solution file.
I use the baseDir from cxx squid and this works perfect for me.

jmecosta:
So I just forgot to include the baseDir in the second paths.get call right?
Does make difference if we pass fs in constructor?

jmecosta:
@Bertk ive checked again. isnt that solution making the squid sensor to parse the log everytime for each module, slowing down the analaysis? anyway ive merged those changes to this branch... we can discuss this later

PR #506, opened 2015-05-01
Improve namespace indentation check

Bertk:
related to https://github.com/wenns/sonar-cxx/issues/441


guwirth:
Looks good. Only thing which could be added is a positive example for nested namespaces.

PR #517, opened 2015-05-22
Add new check for nesting level

Zetten:
Using a method inspired by Sonar-Java, check code for the depth of
nested blocks and raise issues accordingly.

### DIFF
--- a/cxx-checks/src/main/java/org/sonar/cxx/checks/CheckList.java
+++ b/cxx-checks/src/main/java/org/sonar/cxx/checks/CheckList.java
@@ -73,7 +73,8 @@ public final class CheckList {
       UseCorrectIncludeCheck.class,
       XPathCheck.class,
       BooleanEqualityComparisonCheck.class,
-      CompileIncludePathNotFoundOrInvalid.class
+      CompileIncludePathNotFoundOrInvalid.class,
+      NestedStatementsCheck.class
     );
   }
 
--- a/cxx-checks/src/main/java/org/sonar/cxx/checks/NestedStatementsCheck.java
+++ b/cxx-checks/src/main/java/org/sonar/cxx/checks/NestedStatementsCheck.java
@@ -1,0 +1,120 @@
+/*
+ * Sonar C++ Plugin (Community)
+ * Copyright (C) 2011 Waleri Enns and CONTACT Software GmbH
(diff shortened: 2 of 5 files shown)


guwirth:
What is the difference to cyclomatic complexity (http://en.m.wikipedia.org/wiki/Cyclomatic_complexity)?

jmecosta:
They measure different things based on same concept? For example you can say you should only 5 levels of nesting in a function. And the complexity can still not flag any violation because its 10. Complexity gives the idea as if the function is only with one nested level.

Zetten:
Yeah, they are related, but cyclomatic complexity is more of a flat measure, e.g. lots of sequential unrelated `if` blocks = high complexity, while nesting level is a measure of a second dimension and is a good indicator for spaghetti code.

I think the nesting level is more indicative of a need for refactoring, but also easier refactoring. Pulling a deeply-nested block out to a separate function is easier and usually conceptually 'better' than splitting any general high-complexity method.

This PR does have one subtle point that may need consideration. In the AST, the way descendent nodes of `ifStatement` tokens are calculated mean that else-if blocks like:
```
if (cond1) {
  // do something
} else if (cond2) {
  // do something else
} else {
  // do a third thing
}
```
are actually equivalent to this:
```
if (cond1) {
  // do something
} else {
  if (cond2) {
    // do something else
  } else {
    // do a third thing
  }
}
```

So a series of else-ifs will increase the nesting level by one each time. But I think this is standard behaviour for most C++ static analysers (I recall one of my colleagues coded a rule for cppcheck which found the same thing).

However it would also be possible to catch the differing cases of an `else if () {}` and `else { if () {} }` and increase nesting level in the second case but not the first. It just takes an additional check on parent/preceding nodes in the AST.

Perhaps I should make this alternative behaviour a parameter for the check?

guwirth:
If this metric is really measuring the nesting level (second case above) it makes sense. Would add no parameter because first case is equal to cyclomatic complexity. Especially with 'if/else if' and 'switch/case' the cyclomatic complexity metric is mostly wrong.

Zetten:
Okay, the if-else behaviour has been rectified.

The test case is reasonably comprehensive, but please do let me know if there are any glaring errors or omissions in there.

PR #518, opened 2015-05-24
fix sqale definitions

jmecosta:
jump version to 0.9.4 snashot improve some sqale difintions that cause some local analysis failure

### DIFF
--- a/cxx-checks/pom.xml
+++ b/cxx-checks/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.codehaus.sonar-plugins.cxx</groupId>
     <artifactId>cxx</artifactId>
-    <version>0.9.3-SNAPSHOT</version>
+    <version>0.9.4-SNAPSHOT</version>
   </parent>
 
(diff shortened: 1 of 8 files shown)


guwirth:
To merge it pom files (version 0.9.4) should be removed.

jmecosta:
Dont we increment the versions straight after a release?

guwirth:
No, see https://github.com/wenns/sonar-cxx/wiki/Release-Procedure, that is the last step.

jmecosta:
We need to adjust this, usually we need to provide snapshot versions for
users to use to verify defects. We should not provide the same 0.9.3
because it has been released.

IMO the 0.9.4-snapashot should be the first commit after a release. The
same happens with all plugins in sonar community

guwirth:
Hi Jorge. Your are right I'm wrong. Didn't see the SNAPSHOT after the version number, sorry.

jmecosta:
But no need to merge already, let's see if someone can test this

PR #529, opened 2015-06-07
SQALE characteristics for other rules

### LINKED ISSUE #467: Sqale support for 'other' rules
With new API it is no problem to set default values for Sqale/Technical Debt:
http://grepcode.com/file/repo1.maven.org/maven2/org.codehaus.sonar/sonar-plugin-api/4.5.2/org/sonar/api/server/rule/RulesDefinition.java#RulesDefinition.NewRule
```
[Public Method] setDebtRemediationFunction(DebtRemediationFunction) : RulesDefinition.NewRule
[Public Method] setDebtSubCharacteristic(String) : RulesDefinition.NewRule
[Public Method] setEffortToFixDescription(String) : RulesDefinition.NewRule
```
Question is only what is 'default'?

The current used RulesDefinitionXmlLoader does not support Technical Debt values:
http://javadocs.sonarsource.org/4.4/apidocs/org/sonar/api/server/rule/RulesDefinitionXmlLoader.html

```XML
<rules>
   <rule>
     <!-- required fields -->
     <key>the-rule-key</key>
     <name>The purpose of the rule</name>

     <!-- optional fields -->
     <description>
       <![CDATA[The description]]>
     </description>
     <internalKey>Checker/TreeWalker/LocalVariableName</internalKey>
     <severity>BLOCKER</severity>
     <cardinality>MULTIPLE</cardinality>
     <status>BETA</status>
     <param>
       <key>the-param-key</key>
       <tag>style</tag>
       <tag>security</tag>
       <description>
         <![CDATA[the param-description]]>
       </description>
       <defaultValue>42</defaultValue>
     </param>
     <param>
       <key>another-param</key>
     </param>

     <!-- deprecated fields -->
     <configKey>Checker/TreeWalker/LocalVariableName</configKey>
     <priority>BLOCKER</priority>
   </rule>
 </rules>
```

Idea could be to extend this file with Sqale values for the rules, so everything is together in one file?

```XML
<rule>
  ...
  <sqale>
    <chc>
      <key>PORTABILITY</key>
      <name>Portability</name>
        <chc>
          <key>COMPILER_RELATED_PORTABILITY</key>
          <name>Compiler related portability</name>
            <chc>
              <rule-repo>cppcheck</rule-repo>
              <rule-key>AssignmentAddressToInteger</rule-key>
              <prop>
                <key>remediationFunction</key>
                <txt>linear</txt>
              </prop>
              <prop>
                <key>remediationFactor</key>
                <val>5</val>
                <txt>mn</txt>
              </prop>
              <prop>
                <key>offset</key>
                <val>0.0</val>
                <txt>d</txt>
              </prop>
            ...
```

guwirth:
Experimental: Add an option to add SQALE characteristics for external (other) rules. Like rules the user can add them in the configuration under C++ settings / (2) Code analysis.

close #467

@jmecosta, @Bertk and others: What are you thinking?

### DIFF
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRuleRepository.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRuleRepository.java
@@ -28,6 +28,7 @@ import org.sonar.api.server.rule.RulesDefinitionXmlLoader;
 import org.sonar.plugins.cxx.CxxLanguage;
 import org.sonar.plugins.cxx.utils.CxxUtils;
 import org.sonar.squidbridge.rules.SqaleXmlLoader;
+import org.sonar.plugins.cxx.utils.CxxSqaleXmlLoader;
 
 /**
  * Loads the external rules configuration file.
@@ -36,6 +37,7 @@ public class CxxExternalRuleRepository implements RulesDefinition {
 
   public static final String KEY = "other";
   public static final String RULES_KEY = "sonar.cxx.other.rules";
+  public static final String SQALES_KEY = "sonar.cxx.other.sqales";
   public final Settings settings;
   private final RulesDefinitionXmlLoader xmlRuleLoader;
   private static final String NAME = "Other";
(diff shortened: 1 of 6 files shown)


guwirth:
Design decision was to split rules & Sqale characteristics and keep them in two files to reuse existing reader/helper classes from SQ. I think this will keep maintenance effort lower in future.

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxSqaleXmlLoader.java
```diff
@@ -0,0 +1,186 @@
+/*
```
jmecosta: i would prefer not to include this. its likely better to save the xml definition into temp file and use SQ class. having this is we might end up braking our sqale model eveytime core changes something.

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRuleRepository.java
```diff
@@ -48,20 +50,29 @@ public CxxExternalRuleRepository(RulesDefinitionXmlLoader xmlRuleLoader, Setting
...
         try {
           xmlRuleLoader.load(repository, new StringReader(ruleDefs));
         } catch (Exception ex) {
-          CxxUtils.LOG.info("Cannot Load XML '{}'", ex.getMessage());
+          CxxUtils.LOG.info("Cannot load rules XML '{}'", ex.getMessage());
+        }
+      }
+    }
+
+    SqaleXmlLoader.load(repository, "/com/sonar/sqale/cxx-model.xml");
```
jmecosta: with this i think we might want to split the models so its easier to review a improve. i might try to do that at some point

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxSqaleXmlLoader.java
jmecosta: sucks a bit, in this case we need to have a integration test for this. or we might have the same issue as it happen with 0.9.4

guwirth: Thx for the feedback. Problem with this SqaleXmlLoader is that most methods are private so you can't derive and extend functionality. Second point is that it is using InputStreamReader and not Reader. 

What do you mean with read to a temp file? How should this work?

jmecosta: when i first look at that method i tought that thing could be used to read a file with the definitions. but no, so this is ok. my only concern is that we need a integration test to go with this and to cover also our own internal sqale model

guwirth:
@jmecosta there is a unit test for sqale now, please review.

jmecosta:
Looks good.

PR #537, opened 2015-06-20
Improve VC++ build log parser

Bertk:
Tweak VC++ build log parser for visual studio builds logs using normal
verbosity

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
@@ -45,7 +45,7 @@ public class CxxVCppBuildLogParser {
   private final HashMap<String, Set<String>> uniqueIncludes;
   private final HashMap<String, Set<String>> uniqueDefines;
 
-  private VSVersion platformToolset = VSVersion.V100;
+  private VSVersion platformToolset = VSVersion.V120;
   private String platform = "Win32";
 
   public CxxVCppBuildLogParser(HashMap<String, Set<String>> uniqueIncludesIn,
@@ -83,15 +83,15 @@ public class CxxVCppBuildLogParser {
               }
             }
 
-          if(line.contains("\\V100\\Microsoft.CppBuild.targets")) {
-            platformToolset = VSVersion.V100;
-          } else if(line.contains("\\V110\\Microsoft.CppBuild.targets")) {
-            platformToolset = VSVersion.V110;
-          } else if(line.contains("\\V120\\Microsoft.CppBuild.targets")) {
-            platformToolset = VSVersion.V120;
-          } else if(line.contains("\\V140\\Microsoft.CppBuild.targets")) {
-            platformToolset = VSVersion.V140;
-          }
+        if (line.contains("\\V100\\Microsoft.CppBuild.targets") || line.contains("Microsoft Visual Studio 10.0\\VC\\bin\\CL.exe")) {
+          platformToolset = VSVersion.V100;
+        } else if (line.contains("\\V110\\Microsoft.CppBuild.targets") || line.contains("Microsoft Visual Studio 11.0\\VC\\bin\\CL.exe")) {
+          platformToolset = VSVersion.V110;
+        } else if (line.contains("\\V120\\Microsoft.CppBuild.targets") || line.contains("Microsoft Visual Studio 12.0\\VC\\bin\\CL.exe")) {
+          platformToolset = VSVersion.V120;
+        } else if (line.contains("\\V140\\Microsoft.CppBuild.targets") || line.contains("Microsoft Visual Studio 14.0\\VC\\bin\\CL.exe")) {
+          platformToolset = VSVersion.V140;
+        }
 
           // 1>Task "Message"
           // 1>  Configuration=Debug
@@ -138,27 +138,24 @@ public class CxxVCppBuildLogParser {
       ParseInclude(includeElem, projectPath, fileElement);
     }
 
-    for (String includeElem : getMatches(Pattern.compile("/I([^\\s\"]+) "),
-        line)) {
+    for (String includeElem : getMatches(Pattern.compile("/I([^\\s\"]+) "), line)) {
       ParseInclude(includeElem, projectPath, fileElement);
     }
 
-    for (String macroElem : getMatches(Pattern.compile("[/-]D\\s([^\\s]+)"),
-        line)) {
+    for (String macroElem : getMatches(Pattern.compile("[/-]D\\s([^\\s]+)"), line)) {
       AddMacro(macroElem, fileElement);
     }
     
-    for (String macroElem : getMatches(Pattern.compile("[/-]D([^\\s]+)"),
-        line)) {
+    for (String macroElem : getMatches(Pattern.compile("[/-]D([^\\s]+)"), line)) {
       AddMacro(macroElem, fileElement);
     }
-    
+
     // https://msdn.microsoft.com/en-us/library/vstudio/b0084kay(v=vs.100).aspx
     // https://msdn.microsoft.com/en-us/library/vstudio/b0084kay(v=vs.110).aspx
     // https://msdn.microsoft.com/en-us/library/vstudio/b0084kay(v=vs.120).aspx 
     // https://msdn.microsoft.com/en-us/library/vstudio/b0084kay(v=vs.140).aspx
     ParseCommonCompilerOptions(line, fileElement);
-    
+
     if (platformToolset.equals(VSVersion.V100)) {
       ParseV100CompilerOptions(line, fileElement);
     } else if (platformToolset.equals(VSVersion.V110)) {
@@ -167,7 +164,7 @@ public class CxxVCppBuildLogParser {
       ParseV120CompilerOptions(line, fileElement);
     } else if (platformToolset.equals(VSVersion.V140)) {
       ParseV140CompilerOptions(line, fileElement);
-    }    
+    }
   }
 
   private List<String> getMatches(Pattern pattern, String text) {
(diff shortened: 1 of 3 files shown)


jmecosta:
OK and now I really wany to see the main changes on this. Interested on seeing how you manage to get all info using only normal info

cxx-squid/src/test/java/org/sonar/cxx/CxxConfigurationTest.java
```diff
@@ -177,7 +177,7 @@ public void shouldHandleSpeciificV120OptionsCorrectly() {
     
     assertThat(config.getIncludeDirectories().size()).isEqualTo(0);    
     List<String> defines = config.getDefines();
-    assertThat(defines.size()).isEqualTo(commonDefines + 8);
```
jmecosta: why less defines now?

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
```diff
@@ -83,15 +83,15 @@ public void parseVCppLog(File buildLog, String baseDir, String charsetName) {
...
-          if(line.contains("\\V100\\Microsoft.CppBuild.targets")) {
-            platformToolset = VSVersion.V100;
-          } else if(line.contains("\\V110\\Microsoft.CppBuild.targets")) {
-            platformToolset = VSVersion.V110;
-          } else if(line.contains("\\V120\\Microsoft.CppBuild.targets")) {
-            platformToolset = VSVersion.V120;
-          } else if(line.contains("\\V140\\Microsoft.CppBuild.targets")) {
-            platformToolset = VSVersion.V140;
-          }
+        if (line.contains("\\V100\\Microsoft.CppBuild.targets") || line.contains("Microsoft Visual Studio 10.0\\VC\\bin\\CL.exe")) {
```
jmecosta: suppose then we can remove the cppBuild,targets?

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
```diff
@@ -234,10 +228,43 @@ private void ParseCommonCompilerOptions(String line, String fileElement) {
     //__TIME__ The most recent compilation time of the current source file. The time is a string literal of the form hh:mm:ss.
     AddMacro("__TIME__=\"??:??:??\"", fileElement);
     //__TIMESTAMP__ The date and time of the last modification of the current source file, expressed as a string literal in the form Ddd Mmm Date hh:mm:ss yyyy, where Ddd is the abbreviated day of the week and Date is an integer from 1 to 31.
-    AddMacro("__TIMESTAMP__=\"??? ?? ???? ??:??:??\"", fileElement);    
-    //_ATL_VER Defines the ATL version. In Visual Studio 2010, _ATL_VER is defined as 0x0A00.
-    AddMacro("_ATL_VER=1", fileElement);
```
jmecosta: did this one disappear?

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
```diff
@@ -290,24 +319,22 @@ private void ParseCommonCompilerOptions(String line, String fileElement) {
     }
     
     //_DEBUG Defined when you compile with /LDd, /MDd, and /MTd.
-    if (line.contains("/LDd ") ||
```
jmecosta: i would rather leave these as they were, since the comments are closer to the code. easier to follow

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
```diff
@@ -324,155 +351,101 @@ private void ParseCommonCompilerOptions(String line, String fileElement) {
...
       AddMacro("_WIN32", fileElement);
+      // This is not defined for x86 processors.
       AddMacro("_WIN64", fileElement);
       AddMacro("_M_X64", fileElement);
       AddMacro("_M_IA64", fileElement);
       AddMacro("_M_AMD64", fileElement);
     }
-    
-    if(platform.equals("Win32")) {
+    else if(platform.equals("Win32")||line.contains("/D WIN32")) {
```
jmecosta: i think this is problematic, if you have a configuration where you build both win32 and x64 you will be defining all macros to all files. because all of them apply to all files i would still not do this. instead i would keep it as it was so that only one set of macros is defined for the all source code. we have a few of these were we build everything in x64 and because of single project that builds in win32 it will cause issues.

we can do this when we set the macros, defines per file

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
```diff
@@ -324,155 +351,101 @@ private void ParseCommonCompilerOptions(String line, String fileElement) {
...
-  }       
-
-  private void ParseV100CompilerOptions(String line, String fileElement) {
-    // _CPPUNWIND Defined for code compiled with /GX (Enable Exception Handling).
+      //This is not defined for x64 processors.
+      AddMacro("_M_IX86", fileElement); 
+    }
+    // VC++ 16.0
     if (line.contains("/GX ")) {
       AddMacro("_CPPUNWIND", fileElement);
```
jmecosta: this i would move it back to the original place, one less comparation that needs doing if we are not in vs2010

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
```diff
@@ -324,155 +351,101 @@ private void ParseCommonCompilerOptions(String line, String fileElement) {
...
-      AddMacro("__cplusplus_winrt", fileElement);    
     }
-
+    // VC++ 17.0, 18.0, 19.0
     // _CPPUNWIND Defined for code compiled by using one of the /EH (Exception Handling Model) flags.
     if (line.contains("/EHs ") || 
             line.contains("/EHa ") || 
             line.contains("/EHsc ") ||
             line.contains("/EHac ")) {
       AddMacro("_CPPUNWIND", fileElement);
```
jmecosta: same here, if we run vs 2010. when dropping 2010 then it can be made a common?

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
```diff
@@ -324,155 +351,101 @@ private void ParseCommonCompilerOptions(String line, String fileElement) {
...
-    // _M_ARM_FP Expands to a value indicating which /arch compiler option was used:
-    //    In the range 30-39 if no /arch ARM option was specified, indicating the default architecture for ARM was used (VFPv3).
-    //    In the range 40-49 if /arch:VFPv4 was used.
-    //    See /arch (x86) for more information.       
-    if (line.contains("/arch:IA32 ") || 
-            line.contains("/arch:SSE ") || 
-            line.contains("/arch:SSE2 ") ||
-            line.contains("/arch:AVX2 ") ||
-            line.contains("/arch:AVX ")) {
-      AddMacro("_M_ARM_FP", fileElement);
```
jmecosta: removed?

cxx-squid/src/test/java/org/sonar/cxx/CxxConfigurationTest.java
Bertk: I moved some defines to ParseCommonCompilerOptions() and the compiler version specific part is defined individually e.g.:
```
    AddMacro("__cplusplus=199711L", fileElement);
    if ... AddMacro("__cplusplus_winrt=201009", fileElement);
    AddMacro("_MSC_VER=1900", fileElement);
    AddMacro("_MSC_FULL_VER=190022816", fileElement);
    AddMacro("_MFC_VER=0x0C00", fileElement);   
```
The parameter parsing must not be separated as long there is no conflict.
The define for ```"_M_ARM_FP"``` is removed (see other comment).

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
Bertk: I am not sure on the condition for this platform setting and I do not have any sample Windows phone application to do some checks. 
I use PC-lint and the macros are implemented with a helper cpp file http://www.gimpel.com/html/pub90/macros-msc.cpp. I create this Output as a post build step and use it for the Analysis with PC-lint.

jmecosta: I suppose it's not a big issue if we keep it?

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
Bertk: The parameter for cl.exe is not present for newer VC++ compiler versions and therefore this is only relevant for VS2010. The definition will only be created once even if it is multiple defined. I moved it here to have the "_CPPUNWIND" setting at one location but his can be moved back without any Impact.

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
Bertk: Fine with me because it will have no effect on my analysis. 
Should this go again to ParseV110CompilerOptions(), ParseV120CompilerOptions(), ParseV140CompilerOptions() or shall I add it to  ParseCommonCompilerOptions()?

jmecosta: I think so, so there's less stuff to review

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
Bertk: Can you please check with the traget you mentioned. The definition should be either "WIN64" or "WIN32" and the "else if" should be OK.

jmecosta: Right, sorry this looks ok

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
Bertk: Correct. I need to add this again and I will whether I find some more information on the version numbers.
```
_ATL_VER - Defines the ATL version, encoded as an integer literal.
 ```

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
Bertk: I will remove it if you are fine with it. This will be a restriction for the feature and it will only work with the default Visual Studio installation paths.

jmecosta: Actually there is one reason to use the targets file. When using bullseye those gets replaced by bullseye paths. So needs to be there. I supposed this is the change to enable log with normal verbosity?

Bertk: Yes and some additional conditions for the platform "WIN32", "x64" are added as well.

guwirth:
@Bertk & @jmecosta please tell me when ready to merge.

Bertk:
@jmecosta Did you have some time to do some tests? I think we are now ready for the merge.

jmecosta:
@Bertk i was waiting for the review comments, i will do the tests today.

jmecosta:
seems to be running just fine. CR ok

guwirth:
@Bertk could you rebase and squash please

Bertk:
Sorry. I merged the updates from the master but this should also work for the CI.

guwirth:
@Bertk could you squash (merge commits) please. Much easier to handle in case of problems.

Bertk:
I spend some time to make this work "git rebase -i HEAD~4". Notepad++ had a nasty effect which saved the information without any user interaction and I use now notepad to get a change to modify the instructions - just wasted some time :expressionless:
Should be ready for the merge.

guwirth:
With TortoiseGit it's easy going.

guwirth:
But thx will merge it.

PR #541, opened 2015-06-30
additional checks for naming convention and complexity

guwirth:
ClassComplexityCheck
FunctionComplexityCheck
MethodNameCheck
FunctionNameCheck
ClassNameCheck
FileNameCheck

### DIFF
--- a/cxx-checks/src/main/java/org/sonar/cxx/checks/ClassComplexityCheck.java
+++ b/cxx-checks/src/main/java/org/sonar/cxx/checks/ClassComplexityCheck.java
@@ -19,62 +19,59 @@
  */
 package org.sonar.cxx.checks;
 
+import com.sonar.sslr.api.AstNode;
+import com.sonar.sslr.api.Grammar;
+import org.sonar.api.server.rule.RulesDefinition;
 import org.sonar.check.Priority;
 import org.sonar.check.Rule;
 import org.sonar.check.RuleProperty;
 import org.sonar.cxx.api.CxxMetric;
 import org.sonar.cxx.parser.CxxGrammarImpl;
-import org.sonar.squidbridge.api.SourceFunction;
-import org.sonar.squidbridge.checks.ChecksHelper;
-import org.sonar.squidbridge.checks.SquidCheck;
-import com.sonar.sslr.api.AstNode;
-import com.sonar.sslr.api.Grammar;
-import org.sonar.api.server.rule.RulesDefinition;
-import org.sonar.squidbridge.annotations.ActivatedByDefault;
 import org.sonar.squidbridge.annotations.SqaleLinearWithOffsetRemediation;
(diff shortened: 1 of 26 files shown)


cxx-checks/src/main/java/org/sonar/cxx/checks/FunctionComplexityCheck.java
```diff
@@ -46,7 +47,8 @@
   offset = "10min",
   effortToFixDescription = "per complexity point above the threshold")
 public class FunctionComplexityCheck extends SquidCheck<Grammar> {
-  private static final int DEFAULT_MAX = 10;
+
+  private static final int DEFAULT_MAX = 8;
```
jmecosta: any good reason to change this to 8... we usually use 12

guwirth: My understanding is:
1-10 = Low risk program
11-20 = Moderate risk
21-50 = High risk
>50 = Most complex and highly unstable method
Especially in OO development inside of methods 8 is a good value.
Maybe we should allow to define the rule more than once? So someone can e.g. set 8 as minor and 21 as critical?

jmecosta: That's a good idea, so you need multiple rules

jmecosta: But everything bellow 10 imo should not raise any issues

guwirth: My idea was to set @TemplateRule for this.
Comparison is always
   create message if CC > limit
So it is no problem to set different ascending values.

jmecosta: Is it not more difficult for end user to use? It would be nice that the plugin would already define the rules and the default threadsholds ready to use. Using templates likely no one will notice them

guwirth: Yes that is the drawback. You have to create a custom rule from the template: http://docs.sonarqube.org/plugins/servlet/mobile#Rules-RuleTemplatesandCustomRules.

Will set value back to 10. Any other opinions e.g. @Bertk.

guwirth:
Ready to merge:  additional checks for naming convention and complexity

* ClassComplexityCheck
* FunctionComplexityCheck: default function complexity 10
* MethodNameCheck
* FunctionNameCheck
* ClassNameCheck
* FileNameCheck
* MagicNumberCheck: set "0,1,0x0,0x00,.0,.1,0.0,1.0,0u,1u,0ul,1ul,1.0f,0.0f,0LL,1LL,0ULL,1ULL" as default

PR #550, opened 2015-07-04
remove default report paths

### LINKED ISSUE #267: permission denied on reports
Hi folks,

Is possible put a ERROR log to inform the user that the plugin failed to read the reports because permission issues? I had to lost several hours until discover that my report analysis have been failing to submit because permission issues, not because it was not found! My files were in a directory created by the root user and sonar is running with a normal user. :'(

I'm running the sonar-runner with -X and I didn't see any log relating this. Is possible include this?

Thanks!

### LINKED ISSUE #331: Get rid of default values for the sonar.cxx.<reportsource>.reportPath
I consider those values to be of no real help, introducing them was probably a dumb idea. Also, once we have no default values anymore, we could print warnings into the log if a value is set and yields no files, which is valuable user feedback.

guwirth:
* remove  DEFAULT_REPORT_PATH from all sensors
* don't execute sensor if key does not exist in configuration (exception is coverage: sensor set coverage to 0 in this case)
* improve error messages and warnings in case of wrong report settings or access problems
* close #267
* close #331

### DIFF
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
@@ -51,10 +51,6 @@ public class CxxCoverageSensor extends CxxReportSensor {
   public static final String REPORT_PATH_KEY = "sonar.cxx.coverage.reportPath";
   public static final String IT_REPORT_PATH_KEY = "sonar.cxx.coverage.itReportPath";
   public static final String OVERALL_REPORT_PATH_KEY = "sonar.cxx.coverage.overallReportPath";
-  private static final String DEFAULT_REPORT_PATH = "coverage-reports/coverage-*.xml";
-  private static final String IT_DEFAULT_REPORT_PATH = "coverage-reports/it-coverage-*.xml";
-  private static final String OVERALL_DEFAULT_REPORT_PATH = "coverage-reports/overall-coverage-*.xml";
-
   public static final String FORCE_ZERO_COVERAGE_KEY = "sonar.cxx.coverage.forceZeroCoverage";
 
   private static List<CoverageParser> parsers = new LinkedList<CoverageParser>();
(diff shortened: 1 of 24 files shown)


guwirth:
Ready to review.

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
```diff
@@ -72,40 +68,55 @@ public CxxCoverageSensor(Settings settings, FileSystem fs, ProjectReactor reacto
     parsers.add(new VisualStudioParser());
   }
 
+  @Override
+  public boolean shouldExecuteOnProject(Project project) {
+    return fs.hasFiles(fs.predicates().hasLanguage(CxxLanguage.KEY));
```
jmecosta: does this works in multi language?

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
```diff
@@ -72,40 +68,55 @@ public CxxCoverageSensor(Settings settings, FileSystem fs, ProjectReactor reacto
...
    * {@inheritDoc}
    */
   @Override
   public void analyse(Project project, SensorContext context) {
 
-    CxxUtils.LOG.debug("Parsing coverage reports");
-    List<File> reports = getReports(conf, reactor.getRoot().getBaseDir().getAbsolutePath(), REPORT_PATH_KEY, DEFAULT_REPORT_PATH);
-    if (reports.isEmpty()) {
-        reports = getReports(conf, fs.baseDir().getPath(), REPORT_PATH_KEY, DEFAULT_REPORT_PATH);
+    Map<String, CoverageMeasuresBuilder> coverageMeasures = null;
```
jmecosta: exactly this is actually going to slow down the local analysis

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
guwirth: This is a copy of the already existing CxxReportSensor.java / shouldExecuteOnProject: means no change to old version.

jmecosta:
yep this does unnecessary stuff. but once this is merged than i can prepare a pr for the other issue.

looks good to me

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
guwirth: All other sensor with reports are looking first if there is the report key in the sonar-project.properties if not the sensor is not executed. Coverage is different for two reasons:
1) There are three sonar-project.properties keys and not only one.
2) If no report is available there is still the 'sonar.cxx.coverage.forceZeroCoverage'

Inside of analyse I changed things. Each key is checked and if not available nothing is will be done. If also sonar.cxx.coverage.forceZeroCoverage is false then analyse is doing nothing and should be fast.

jmecosta: It was fast that's true. So I guess the other can also be close as it does not affect performance that much

PR #553, opened 2015-07-05
TabCharacterCheck per line

### LINKED ISSUE #390: tabcheck should report issue against each line instead of file
- vera works this way
- more often than not more than one person is to be able

guwirth:
Add an option to select if violation is for each line or only once per file (default).
close #390


guwirth:
@jmecosta Especially for you ;-) please review

jmecosta:
thanks... looks good

PR #555, opened 2015-07-05
use boost 1.55 for integration test (old was 1.46)

guwirth:
This will also test grammar & parser.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,18 @@
 language: java
 
 jdk:
-  - oraclejdk8
-  - openjdk7
+        - oraclejdk8
+        - openjdk7
 
 env:
-  - SONARHOME=/tmp/sonarqube-4.5.4
-  - SONARHOME=/tmp/sonarqube-5.0.1
-  - SONARHOME=/tmp/sonarqube-5.1.1
+        - SONARHOME=/tmp/sonarqube-4.5.4
+        - SONARHOME=/tmp/sonarqube-5.0.1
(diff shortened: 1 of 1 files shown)


guwirth:
Will create a Travis CI ticket

guwirth:
**Travis CI support:**

Thanks for getting in touch.

Unfortunately, I don't know exactly which boost versions does Ubuntu 14.04 support.

From boost's mailing list I see there was somebody with an issue similar to yours: http://lists.boost.org/Archives/boost/2015/04/221774.php, have you tried to install those dependencies in your script? 

Hope this helps, please let us know how it goes and if you got any other questions.

guwirth:
**Travis CI request:**

Ubuntu is saying that version 1.54 should be supported, but it is still 1.46
http://packages.ubuntu.com/trusty/libboost-all-dev

Could you verify that please.

guwirth:
I'm sorry for confusion, our default VM image is precise at the moment, not trusty. On precise the newest version of libboost is 1.48. That said, I've found a ppa repository, which also have newer versions: https://launchpad.net/~boost-latest/+archive/ubuntu/ppa

You may try installing boost 1.55 with:

sudo apt-add-repository -y ppa:boost-latest/ppa
sudo apt-get update
sudo apt-get install boost1.55

If you need newer versions, I'm afraid you'll have to install that from source.

Another way of dealing with it would be to use Ubuntu Trusty, although it's a beta image. Would you like to try it?

Issue #560, opened 2015-07-07
C++11: attributes


guwirth:
Not all valid combinations of attributes are working.

Example:
```C++
[[attr1]] void [[attr2]] func(int [[attr3]] p) [[attr4]]
{
 //@todo
    [[attr4(arg1, arg2)]] if (cond)
    {
        [[vendor::attr5]] return i;
    }
}

//@todo
[[attr1]] class C [[attr2]]
{

   C::C [[attr6]] () [[attr7]];

} [[attr3]] c [[attr4]], d [[attr5]];
// attr1 applies to declarator-ids c, d
// attr2 applies to the definition of class C
// attr3 applies to type C
// attr4 applies to declarator-id c
// attr5 applies to declarator-id d
```

PR #571, opened 2015-07-08
support attribute deprecated with user defined message

### LINKED ISSUE #565: C++14: attribute deprecated with user defined message
Attributte deprecated with user defined message not supported:

```C++
//@todo
[[deprecated("g() is thread-unsafe. Use h() instead")]]
void g( int& x );
```

guwirth:
close #565


guwirth:
implement C++11 Attributes according to N2761
support attribute deprecated with user defined message
close #565
close #560

PR #581, opened 2015-07-10
Generate public API documentation violation for alias declaration

### LINKED ISSUE #577: Type alias (using =) does not generate public API documentation violation
Hi,

I have a class with a type alias
```
class A
{
public:
   using X = int;
   typedef int Y;
};
```
This does not generate a public API documentation violation, although the typedef does.

lcintrat:
Fix #577

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/visitors/AbstractCxxPublicApiVisitor.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/visitors/AbstractCxxPublicApiVisitor.java
@@ -47,6 +47,7 @@ import com.sonar.sslr.impl.ast.AstXmlPrinter;
  * <li>enumerations</li>
  * <li>enumeration values</li>
  * <li>typedefs</li>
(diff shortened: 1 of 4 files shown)


guwirth:
@lcintrat Looks good. Only hint: you check not all getNext..., getFirst... for null. But don't know if this is always needed?

lcintrat:
@guwirth You are right, I rely on the parser. I assume that if the parsing passed, then the searched node is present (according to the grammar).
I can add checks for null, but what is the rule on error? Error message or is there a mean to create a SQ issue for that?

guwirth:
@lcintrat it should at least not crash.
Some grammar parts are optional, error recovery could be enabled, ...
Can speak only how I do it. If return value could be null I always check it.

lcintrat:
:+1:

guwirth:
@lcintrat if finished please rebase & squash. I will merge it then.

lcintrat:
Done

PR #603, opened 2015-08-01
stringinize with string prefix

Bertk:
The preprocessor will not identify an operator if a string prefix is used and this will create parser errors. 
This PR enhances the preprocessor to accept e.g. 'L#' as operator.


guwirth:
adding a source code example would be helpful

Bertk:
I extended the unit tests and maybe this sample could be added to one of the source files to check the pre-processor e.g. http://stackoverflow.com/questions/20003366/stringizing-argument-of-macro-to-be-unicode.

PR #606, opened 2015-08-05
Added constexpr check to magic number checker.

jens-auer:
Added constexpr check to MagicNumberCheck to not report violations on constexpr definitions.

### DIFF
--- a/cxx-checks/src/test/resources/checks/magicNumber.cc
+++ b/cxx-checks/src/test/resources/checks/magicNumber.cc
@@ -5,6 +5,8 @@ class Program
 {
     int _tmain(int argc, _TCHAR* argv[])
     {
+        constexpr int compliant1 = 10;             // Compliant
+        static constexpr int compliant2 = 10;      // Compliant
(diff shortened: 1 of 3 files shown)


jens-auer:
Could somebody provide some detail why the build fails? The log just says "Couldn't start SonarQube", which is not very helpful. I would fix the issue, but without any information this will be hard.

cxx-checks/src/test/resources/checks/magicNumber.cc
```diff
@@ -43,13 +45,19 @@ int someData[] = {
 
 static const int i1 = 10;     // Compliant
 int i2 = 11;                  // Non-Compliant
+constexpr int compliant1 = 10;
```
guwirth: Maybe you can add //compliant -or- // non-compliant also to your new lines?

guwirth:
Think this was an Travis CI problem. We had this also in the past.
Just make a change rebase, squash and try again.

jens-auer:
Hi,

I've added the Compliant comments to the cpp test file.

Issue #600, opened 2015-07-27
PC-lint report may create SonarException in decorator


Bertk:
A rare scenario will create a java.lang.ArrayIndexOutOfBoundsException in decorator.
This will happen whenever a issue is detected after the complete file was processed e.g.
```number = "766" desc = "Header file 'some-header.h' not used in module 'src\mysource.cpp'"``` and there is no empty line at the end of the file. The line number for the issue in the PC-lint report file is incremented by 1 compared to the number of lines in the source file.

This was observed with SQ 5.1.1 and the C++ community plug-in.

Possible Solutions:
a) Add an empty line at the end of the source files
b) Correct the line number of the issue from the Report and set it to line=0 or the value for the last line.

Bertk:
The SonarException will be observed for **RATS** and **PC-lint** reports. Both analysis tools create line numbers greater than the available *max line of the source file* and for me this looks like some "special feature". The C standard anyway demands a single empty line at the end of every file and I modified the Cxx plug-in to correct the report line number and limit it to the *max line of the source file* (4d4efdcb4b36c8e1dd04bb744c4a91b64a02ae94).
The exception is created only with SQ 5.x and not with the SQ 4.5.5 (LTS) version.

Bertk:
see also SQ issue 6792 (https://jira.sonarsource.com/browse/SONAR-6792)

PR #611, opened 2015-08-16
limit report line number

Bertk:
report line number shall not be greater than the existing max. line number
(see also #600).

### DIFF
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxReportSensor.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxReportSensor.java
@@ -52,6 +52,7 @@ public abstract class CxxReportSensor implements Sensor {
   private ResourcePerspectives perspectives;
   private Set<String> notFoundFiles = new HashSet<String>();
   private Set<String> uniqueIssues = new HashSet<String>();
+  private HashMap<String, Integer> maxLineFiles = new HashMap<String, Integer>();
   private final Metric metric;
   private int violationsCount;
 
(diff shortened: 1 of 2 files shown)


sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxUtils.java
```diff
@@ -78,8 +82,32 @@ public static String normalizePathFull(String filename, String baseDir) {
...
+   * @return returns true for Reactor Projects
+   */
   public static boolean isReactorProject(Project project) {
     return project.isRoot() && !project.getModules().isEmpty();
   }
+  
+  /**
+   * @return returns number of Lines of a File
+   */
+  public static int countLines(String normalPath) {
```
guwirth: Is there no core metric we can use to get the line number?

Bertk: There is a line number in the size section of the metrics.
http://docs.sonarqube.org/display/SONAR/Metric+definitions#Metricdefinitions-Size
I am not sure whether the line metric is available when the report is consumed.

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxReportSensor.java
```diff
@@ -239,6 +240,21 @@ private boolean saveViolation(Project project, SensorContext context, String rul
           = org.sonar.api.resources.File.fromIOFile(new File(normalPath), project);
```
guwirth: fromIOFile is deprecated, see http://javadocs.sonarsource.org/4.5.2/apidocs/deprecated-list.html. So we should use something like:
```Java
InputFile file= fs.inputFile(fs.predicates().is(new File(filename)));
int maxLine = file.lines();
```
InputFile has method lines.

jmecosta: performance needs to be measured...

Issue #426, opened 2015-02-14
Relative paths in Cobertura reports are treated relative to the current directory instead of the projectBaseDir


abbeyj:
When using reports in Cobertura format that contain relative paths and running sonar-runner from a directory other than the projectBaseDir, the paths are canonicalized incorrectly.  This leads to the relevant files not being found.

For instance, say my project base directory is ./base_dir and I also have ./sonar/sonar-project.properties.  After building the project, I switch to the ./sonar directory and then run sonar-runner.  The current directory while running is ./sonar but relative paths in the Cobertura report should be taken relative to the projectBaseDir.  Making them relative to the current directory isn't going to find anything as the only thing in the sonar directory is the sonar-project.properties file.

I have sample project to demonstrate this at http://files.stardrifter.org/pub/cobertura_paths.tar.gz

To use it:
$ tar xzvf cobertura_paths.tar.gz
$ cd cobertura_paths/base_dir
$ make
$ cd ../sonar
$ sonar-runner

You'll see a message like
WARN  - Cannot find the file '&lt;snip&gt;/cobertura_paths/**sonar**/main.cpp', ignoring coverage measures

It should be looking instead for .../cobertura_paths/**base_dir**/main.cpp which is where the file is actually located.

I believe the problem occurs at the normalizePath call here:
https://github.com/wenns/sonar-cxx/blob/79894beea01bd829275c8c6e5502855bcf063cb1/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CoberturaParser.java#L70

This needs to be modified to take into account the projectBaseDir from the .properties file and not just use the current directory as CxxUtils.normalizePath does.

guwirth:
Maybe wiki helps https://github.com/wenns/sonar-cxx/wiki/Path-and-path-separator-issues

abbeyj:
That link doesn't help me.  I'm going by the instructions at 
http://docs.sonarqube.org/display/SONAR/Analyzing+with+SonarQube+Runner#AnalyzingwithSonarQubeRunner-AdvancedUsage
and
http://docs.sonarqube.org/display/SONAR/Analysis+Parameters#AnalysisParameters-projectBaseDir

[...]

PR #613, opened 2015-08-21
Treat relative paths in Cobertura reports as relative to the project …

abbeyj:
…base directory as opposed to the current directory.

This is intended to address #426.  That issue is currently marked closed but it is not fixed I've requested that it be reopened.

### DIFF
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CoberturaParser.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CoberturaParser.java
@@ -36,7 +36,7 @@ import org.sonar.plugins.cxx.utils.CxxUtils;
 /**
  * {@inheritDoc}
  */
-public class CoberturaParser extends CoverageParserBase implements CoverageParser {
+public class CoberturaParser extends CxxCoverageParser {
 
   public CoberturaParser(final String baseDir)
   {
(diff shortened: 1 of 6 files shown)


sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CoberturaParser.java
```diff
@@ -37,6 +37,14 @@
  * {@inheritDoc}
  */
 public class CoberturaParser implements CoverageParser {
```
guwirth: @abbeyj looks like a specific solution for CoberturaParser. Could you move this ```baseDir``` to ```CoverageParser``` to make it more generic please.

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
```diff
@@ -63,7 +63,7 @@ public CxxCoverageSensor(Settings settings, FileSystem fs, ProjectReactor reacto
     super(settings, fs, reactor);
 
     this.reactor = reactor;
-    parsers.add(new CoberturaParser());
+    parsers.add(new CoberturaParser(fs.baseDir().getAbsolutePath()));
```
guwirth: Would forward baseDir to all parsers in the ctor.

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CoberturaParser.java
abbeyj: CoverageParser is an interface so I can't move it there.  Do you want me to create a base class and put baseDir there?  If so, do you have a preference for a name?

guwirth:
@abbeyj Sorry for the late answer. Would do it like here https://github.com/wenns/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxReportSensor.java

New class CxxCoverageParser which implements CoverageParser. Current location of your class in coverage folder is ok.

guwirth:
Thx

PR #617, opened 2015-08-25
Remove module restriction on CxxXunitSensor

Zetten:
Allow the test execution report sensor to run in the same way as other
sensors, rather than limited to "non-modules" in simple mode.

### DIFF
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/xunit/CxxXunitSensor.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/xunit/CxxXunitSensor.java
@@ -110,21 +110,6 @@ public class CxxXunitSensor extends CxxReportSensor {
   public Class<?> dependsUponCoverageSensors() {
     return CoverageExtension.class;
   }
-  
-  /**
-   * {@inheritDoc}
-   */
-  @Override
-  public boolean shouldExecuteOnProject(Project project) {
-    if (conf.hasKey(reportPathKey())) {
-      if (!conf.getBoolean(PROVIDE_DETAILS_KEY)) {
-        return !project.isModule();
-      } else {
-        return fs.hasFiles(fs.predicates().hasLanguage(CxxLanguage.KEY));
-      }
-    }
-    return false;
-  }
 
   /**
    * {@inheritDoc}


guwirth:
PR is failing and can't be merged.

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/xunit/CxxXunitSensor.java
```diff
@@ -110,21 +110,6 @@ protected String reportPathKey() {
...
-  /**
-   * {@inheritDoc}
-   */
-  @Override
-  public boolean shouldExecuteOnProject(Project project) {
-    if (conf.hasKey(reportPathKey())) {
-      if (!conf.getBoolean(PROVIDE_DETAILS_KEY)) {
-        return !project.isModule();
-      } else {
-        return fs.hasFiles(fs.predicates().hasLanguage(CxxLanguage.KEY));
```
guwirth: Not sure if removing shouldExecuteOnProject is a good idea? Think it should test at least ... hasLanguage(CxxLanguage.KEY)?

Zetten:
The default shouldExecuteOnProject in the superclass (CxxReportSensor) is to check for the presence of the language and that the report path property is configured.

But now to find out why the ITs are failing, as everything seemed fine locally...

jmecosta:
Did you run the it tests locally?

Zetten:
I did, but I think I was accidentally running the ITs against an existing (running) instance of SQ which was running an older sonar-cxx version so everything looked fine.

So with a totally clean SQ instance, I can reproduce the failure. Two issues:

1. We need to pass both module and reactor paths to `getReports` rather than `""` (as other sensors already have). This was breaking the Travis build as no reports were being found so the test metrics were all 0.
2. After fixing that, the multimodule project adds the same test cases/suites repeatedly by parsing the same project-global report for each module. So in the test project we get 12 tests instead of the expected 3 (3 tests * 4 modules).

I'm not sure how to deal with that second problem. The other sensors don't seem to have any issue with parsing the same report several times in a multimodule project, I guess maybe because issues and coverage data are line-level measures and SQ deals with the duplication.

Any ideas?

Maybe we could try to identify somehow whether a given report has been parsed already, or add a config option to the CxxXunitSensor which enables the existing "top level project only" behaviour by default and make "per module test reports" optional.

jmecosta:
I think that thing was there because of exactly that problem. I don't mind
having a option to optionally enable that call.

Modules really suck

Bertk:
We should have a solution for this problem anyway. Some plug-ins use conditions to reduce repeated analysis of reports. Can we do the same for the test results?
```
public static boolean isReactorProject(Project project) { 
     return project.isRoot() && !project.getModules().isEmpty(); 
 } 
```

Bertk:
I spend some time and found a solution which should be checked. The shouldExecuteOnProject function is used to check whether a test report exist or not (see b4e7893fcf94963968aff649495e8679d1f6a9ec).
In addition you need to define the test reports by module e.g. in sonar-project.properties use ```modul1.sonar.cxx.xunit.report=module1-test-report.xml.```
This will work with the setting ```sonar.cxx.xunit.provideDetails=false```

guwirth:
@jmecosta & @Bertk & @Zetten : Shell we close this PR and create a new one with improved solution?

jmecosta:
I am OK with it

PR #619, opened 2015-08-27
support team builds

Bertk:
There is only one log file for a VS solution for multiple projects (or modules) which might be located in the parent folder of the project root folder.


guwirth:
Bigger change. @jmecosta could you verify it please? Think you are also using TFS.

jmecosta:
Nahh, to expensive for us :). However I will need to use the msbuild
runner, so I might need to check that.

guwirth:
@jmecosta & @Bertk: How should we proceed with this one?

Bertk:
Let us wait until Jorge provides the feedback. This update enables the usage of absolute paths for the report files and reuses this report for all modules. If the file does not exist (not a absolute path) the regular search rules are used to detect the report file(s).

jmecosta:
i will tomorrow do a regression test in this pr, if that is ok i think i can be merged

guwirth:
@Bertk sorry for the inconvenience. But could you rebase it please...

Bertk:
@guwirth I did the rebase without squash but don't know why CI fails.

guwirth:
@Bertk think is a Travis problem.  Sometimes one of the builds is red because of internal Travis problems. In the past I did a small change and tried again an upload. I can try to write the Travis support again. Maybe they can help this time to avoid this in future.

guwirth:
Travis:
  "Thanks for getting in touch, and sorry to hear about the troubles.

  We have had a look at some of the failing jobs on your repository, and it is seems pretty hard to tell for us what might be causing this.

  Have you observed any kind of pattern in these failures? It seems sometimes an expected plugin cannot be found, but sometimes it just logs an unspecific message like "cannot start SonarCube".
May I suggest to add additional debug output around these failures so it might be easier for you to spot potential causes?"

guwirth:
@jmecosta did you have time for testing?

jmecosta:
Nop, I will build today and check the results tomorrow.

jmecosta:
results were ok. for me it can be merged

guwirth:
@jmecosta Thx

PR #635, opened 2015-09-23
Fix cpplint's NOLINT() with new rule names

cbagwell:
NOLINT() only supports original rules list from _ERROR_CATEGORIES
but that list was left unmodified by cpplint_creatrules.py script.
This meant that any existing code that developers marked as false
positives using NOLINT method began reporting errors when switching to
modified cpplint and also it was impossible to re-mark using new rule names.

This modifies NOLINT() behavior to work like cpplint's --filter option
which masks any rule that starts with specified pattern.

This allows, for example, "// NOLINT(build/c++11)" to mask build/c++11
rule from non-modified cpplint as well as build/c++11-0 in modified
cpplint.

### DIFF
--- a/sonar-cxx-plugin/src/tools/cpplint_createrules.py
+++ b/sonar-cxx-plugin/src/tools/cpplint_createrules.py
@@ -31,6 +31,14 @@ def WriteUpdateFile(filename):
     global _FILE_LINES
     filetowrite = open(filename, 'w')   
     for linei in _FILE_LINES:                   
+        if linei.startswith('        if category in _ERROR_CATEGORIES:'):
+            filetowrite.write('        for value in _ERROR_CATEGORIES:\n')
+            filetowrite.write('          if category.startswith(value):\n')
+            filetowrite.write('            category = value\n')
+        if linei.startswith('  return (linenum in _error_suppressions.get(category, set()) or'):
+            filetowrite.write('  for key in _error_suppressions.keys():\n')
+            filetowrite.write('    if category.startswith(key):\n')
+            filetowrite.write('      category = key\n')
         filetowrite.write(linei)
                                                         
     filetowrite.close()


jmecosta:
Thanks for the PR, I will try this as soon as possible. There are so check broken it might just be travis . you can try to push something to see if it passes

cbagwell:
Thanks for the quick feedback.  I looked for a while in travis logs to feel comfortable that the boost test failure is not likely to be related to PR.  I'll go read up best way to restart travis job.

In mean time, here is a small test C++ file to run with cpplint.py and cpplint_mod.py to demonstrate the issue before this PR.  The rule for using unapproved <thread> library becomes build/c++11-2 in modified version of cpplint.  2nd line is correctly disabled when ran with cpplint.py but all three show up as issues with cpplint_mod.py.

```
#include <thread>
#include <thread>  // NOLINT(build/c++11)
#include <thread>  // NOLINT(build/c++11-2)
```

jmecosta:
OK now I got it... Haven't come across this since usually we don't like
polluting code with additional NO lint in code. Thanks for the example

jmecosta:
i cannot get this thing to work, // NOLINT alone works, but with the patched version i cannot get // NOLINT(readabilitiy/casting-1) to be ignored

what version are u using of cpplint?

jmecosta:
@cbagwell ok, now it works, just realize my version is really old...need to update then.

this can be merged. and thanks for the fix

cbagwell:
To be safe, I looked at cpplint.py commit log and looks like code this PR applies to was added in Apr 30, 2010.  I tested PR on that version and test worked.  Seems reasonably long enough time period to not worry about asking people to upgrade.

guwirth:
@cbagwell final version ready to merge?

cbagwell:
@guwirth Yes, its final version and ready to merge.

PR #637, opened 2015-10-05
add rule tags

Bertk:
Add tags for some rules of CppCheck, RATS and PC-lint (initial
effort).


guwirth:
Looks good to me :+1: 
Any other feedback?

PR #644, opened 2015-10-17
change colourizer 

jmecosta:
with this pr i get this error: 

ERROR: Caused by: Unable to highlight file [moduleKey=com.fmrco.ist.fot:tws, relative=CppProjectMultiModule/CppProjectMultiModule.h, basedir=D:\SonarVersions\MultLanguageProject] from offset 1 to offset 13
ERROR: Caused by: 13 is not a valid offset for file [moduleKey=com.fmrco.ist.fot:tws, relative=CppProjectMultiModule/CppProjectMultiModule.h, basedir=D:\SonarVersions\MultLanguageProject]. Max offset is 12

for this file:
#pragma once

if include \r\n then it does not fail. 

this pr is just to help debug what file is failing so users can add the new line


guwirth:
@jmecosta wouldn't it be better to fix the wrong offset calculation instead of creating an error message?

jmecosta:
Sure, i did try. But did not went far because all the tests are throwing
exceptions not sure why?

But anyway, even if fixing this case there might be more and the logging
should help

jmecosta:
@guwirth i will try to still understand what that thing, seems to me that we can use another approach to highlight code. for example : 
https://github.com/SonarCommunity/sonar-csharp/blob/master/src/main/java/org/sonar/plugins/csharp/CSharpSourceCodeColorizer.java

that does not rely on the AST

guwirth:
:+1:  PR looks good to me. My only question is if this interface is the most up to date one?

jmecosta:
perhaps not, however i did a test with 5.2 RC 2 and with the previous
version on top of the problem i add during normal analysis there was
another issue in server side.

see: https://groups.google.com/forum/#!topic/sonarqube/lupyII6ATvo

once i used the colorizer everything went ok. anyway, as long as we follow
what the c# plugin does i think we are ok. at least i took the code from
there

Issue #648, opened 2015-10-18
Add SQ 5.2 to Travis build


guwirth:
Add SQ 5.2 to Travis build:
https://sonarsource.bintray.com/Distribution/sonarqube/sonarqube-5.2-RC2.zip

See #649

jmecosta:
Curious to see if this works. At least I was not able to run c++ analysis without the colorizer

guwirth:
Also forceZeroCoverage seems no more to work?

jmecosta:
Does that thing do some processing that access db? This is no longer
possible in plugin side. We need to update to 5.2 rc2 to see what
components are broken. Problem I can't find those artifacts anywhere

guwirth:
Currently known issues:
* additional warning: WARN  - Unable to get a valid mac address, will use a dummy address
* Seems there is a problem to access the analysis log. This is different with SQ 5.2? Think SQ will create it on server side and provide it delayed? Blocker to use sonar runner in build automation. 
* Also forceZeroCoverage seems no more to work?

guwirth:
> Does that thing do some processing that access db?

Code is here https://github.com/wenns/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java. Maybe context.GetMeasure? How do I know which API is using the DB?

> We need to update to 5.2 rc2 to see what components are broken.

Depends. If there are breaking changes in SQ 5.2 we should first release a final 0.9.4. For my understanding plugin should work with SQ 4.5 to 5.2. 

> Problem I can't find those artifacts anywhere

What are you looking for?

jmecosta:
On Sun, 18 Oct 2015 at 18:33 Günter Wirth <notifications@github.com> wrote:

> Does that thing do some processing that access db?
>
> Code is here
> https://github.com/wenns/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java.
> Maybe context.GetMeasure? How do I know which API is using the DB?
>
yep, it looks like its that: changes are here:
https://groups.google.com/forum/#!topic/sonarqube/ZCIsvd41Esk but unclear
to me what needs changing

> We need to update to 5.2 rc2 to see what components are broken.
>
> Depends. If there are breaking changes in SQ 5.2 we should first release a
[...]

PR #649, opened 2015-10-18
add sonarqube-5.2-RC2 to Travis

guwirth:


guwirth:
Getting this new warning with SQ 5.2. Don't know why?
```
 Assertion Failed: Found following errors and/or warnings lines in the logfile:
      10:44:00.604 WARN  - Unable to get a valid mac address, will use a dummy address
      For details see _coverage_project_.log
```

guwirth:
Seems there is a problem  to access the analysis log. This is different with SQ 5.2? Think SQ will create it on server side and provide it delayed?

jmecosta:
Yep but you need to access the ui, I think there is also some rest api for
this. At least I've seen something I'm the api. I've raise a question about
this in the forum that we now don't know if analysis is OK or not even if
its ok locally

guwirth:
Natural solution would be to have an option in sonar runner to wait for the result and provide it on client side.

jmecosta:
Right, but I guess the only possible workaround is to poll the server after
each test

guwirth:
* update links to https://sonarsource.bintray.com/Distribution/sonarqube
* add '.*WARN.*Unable to get a valid mac address, will use a dummy address' to integration tests
* mark SQ 5.2 as allowed to fail
* build will fail as soon as a job has failed, or when the only jobs left allows failures
* close #648

PR #652, opened 2015-10-19
change paths in report so they dont randomly clash with folders that exist in e drive

jmecosta:
ok, got it... this getCanonicalPath() is kind of evil. my problem was that i had a folder in e:\ with e:\Test then when this is called with e:\test\askdsa.c it will retrieve E:\Test\askds.c and the tests fail in my environment.  

in real life, we should run the coverage in same machine as the analysis or at least make sure everything is the same across machines. 

@guwirth can you verify this in your environment?


guwirth:
We have different servers running in parallel which:
- all have same folder structure
- or we patch reports afterwards before we forward it to sonar runner

PR #655, opened 2015-10-21
Fix NOLINT(*) regression

cbagwell:
After recent commit was tested over wider existing code
base, it was reported cpplint_mod.py would crash on
source code that contained "// NOLINT(*)".

This regression was fixed and tested against this fuller
(fullest) set of cases.  First two are "mask all lint reports"
wildcards.

    #include <thread>
    #include <thread>  // NOLINT
    #include <thread>  // NOLINT(*)
    #include <thread>  // NOLINT(build/c++11)
    #include <thread>  // NOLINT(build/c++11-2)


guwirth:
@jmecosta I don't use lint: can you give your ok to merge it please

jmecosta:
Seems all good

Issue #527, opened 2015-06-02
Custom SquidCheck


schiean:
Hi,

At our company we have defined some custom checks in java classes (extends SquidCheck<Grammar>). For now I use a locally forked cxx. However we also implement the same checks for Java. And the sonar-java plugin allowed us to separate those checks in a separate plugin (so we don't have to use a local fork). I would prefer to just install default sonar-java and default sonar-cxx in combination with a company-checks plugin which contains our checks for cxx and java.

I created a class that implements RulesDefinition and does the following:
  NewRepository repo = context.createRepository(REPOSITORY_KEY, KEY);
  repo.setName(REPOSITORY_NAME);
 RulesDefinitionAnnotationLoader annotationLoader = new RulesDefinitionAnnotationLoader();
 annotationLoader.load(repo, checks);
 repo.done();

This seems to work for Cxx rules to get them in the sonar-rule repository. Also they are activated in my profile, yet they don't seem to be included by CXX-squid when sonar-runner is being run.
(The rule does work in a local-cxx-checks fork..)

How can i provide my own rules to the CxxSquidSensor ?

Kind Regards,
Arjen

guwirth:
See #467 and #521. 

Which additional checks do you have? Can you share them?

schiean:
>> Which additional checks do you have? Can you share them?

Most of them are really specific to our company coding conventions, they are adjusted versions of function complexity check and function size check. The only one I think that might be of interest, is one that 'counts the number of parameters/arguments to a function and above a certain threshold triggers an issue'. If you want I could provide that check to this repo.

As for my question, I don't really understand the relation to the tickets you linked. But I am pretty new to Cxx so maybe it should have been obvious, however could you explain a little more ?

Below is an example of one of my rules/checks. The system.out is there for debug purpose. It does get printed when it is packaged in cxx-checks, but not when it is packaged in my company-plugin.

	@Rule(key = "...", name = "..", description = "..", tags = { ".." })
	@NoSqale
        @ActivatedByDefault
	public class ParamCountCheck extends SquidCheck<Grammar> {

		@Override
		public void init() {
			subscribeTo(CxxGrammarImpl.functionDefinition);
		}

		@Override
		public void leaveNode(AstNode node) {
			int nrOfParams = count(node);
			System.out.println("Debug, "+printName(node)+" has params:"+nrOfParams);
			// create issues
		}
}

guwirth:
[...]

PR #667, opened 2015-11-03
Support plugins with custom rules

felipebz:
As I mentioned in https://github.com/SonarOpenCommunity/sonar-cxx/issues/527#issuecomment-153189197, these are the changes necessary to allow custom checks in external plugins.

The grammar classes had to be moved to org.sonar.plugins.cxx.api due the SonarQube plugin classloader (http://docs.sonarqube.org/display/DEV/Plugin+Class+Loader).

The CustomCxxRulesDefinition class is the extension point. All plugins must extends this class to add new checks.

Instances of CustomCxxRulesDefinition are injected automatically by SonarQube in CxxSquidSensor.

I consider this a WIP because:
- needs more discussion, of course
- I don't like to hardcode the sonar-cxx-plugin version in the example plugin. Can I make this project a submodule of sonar-cxx (to compile with the project) or should I create a new kind of integration tests for it?

### DIFF
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/api/CustomCxxRulesDefinition.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/api/CustomCxxRulesDefinition.java
@@ -1,0 +1,59 @@
+/*
+ * Sonar C++ Plugin (Community)
+ * Copyright (C) 2010 Neticoa SAS France
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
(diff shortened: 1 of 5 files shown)


guwirth:
@felipebz please rebase and update PR, Travis CI is working again

felipebz:
It should be ok now.

Btw, I reverted the rename from org.sonar.cxx.api to org.sonar.plugins.cxx.api. The sample plugin uses CxxGrammarImpl (which is not in org.sonar.plugins.cxx.api) and it seems to work fine. I need to check why.

guwirth:
Hello @felipebz thanks for providing the PR. Will have a look the next days.

guwirth:
I like the idea to have an extension point for custom plugins to add their own checks. From my point of view the minimum requirements to add it to the cox plugin are:

1. Should have no negative impact to the other use cases. 
2. Unit tests for your extension exists. 
3. Documentation how to use it exists. 

With the example code within our plugin I'm not sure. Here I see the risk that others add more "examples" which contradicts the idea of this extension point. So I would prefer to have the sample outside if unit testing is without possible. But like to here also other opinions...

In general :+1:

jmecosta:
I agree with @guwirth , i dont think having the sample in the repository brings much value. imo its best to just create a new repository in the https://github.com/SonarOpenCommunity to demonstrate as a example. Having it properly documented i think its the best option.

felipebz:
I agree with you. The example plugin is not included in this PR anymore (but the code still available here for reference: https://github.com/felipebz/sonar-cxx/commit/ce6d5024dfce22213e116279a4dc7c2ea5ecd177).

I was concerned with the binary compatibility of CxxGrammarImpl because it's an enum and I wasn't sure about what would happen if a new value was added between the existing ones (in C# this is a very dangerous idea). But it works fine in Java.

So, is there anything to do? Please let me know if I need to change something.

guwirth:
@felipebz thanks. I think there is a unit test for CxxChecks missing?

guwirth:
@felipebz should we merge this now?

felipebz:
Sorry, I've added some tests for CxxChecks

guwirth:
:+1:  Any other opinions?

Issue #670, opened 2015-11-03
Update Center links are incorrect


patrickjmccarty:
Bug in plugin version 0.9.4:
Within SonarQube, going to System > Update Center > and clicking to expand C++ (Community) shows a Homepage link and Issue Tracker link.

The Homepage link is incorrect; it currently goes to http://docs.codehaus.org/display/SONAR/CXX+PLUGIN which dumps me at SQ's generic Documentation page.

The Issue Tracker link works, but it ends up getting redirected to a different location, so it should probably be updated as well so it doesn't rely on the redirect?
Link is: https://github.com/wenns/sonar-cxx/issues?state=open
Ends up at: https://github.com/SonarOpenCommunity/sonar-cxx/issues?q=is%3Aopen

guwirth:
This plugin us not supported by the update center. For SonarSource this plugin is a competitor to their own commercial C++ plugin and therefore excluded.

jmecosta:
I think this was because of the move to an organization. Once you install
the plugin it will be in update center. We need to fix the poms to point to
the correct place now.

jmecosta:
You can assign to me, I will check

patrickjmccarty:
@jmecosta Regarding the homepage link, searching the source for CXX+Plugin, I found it two files:
[sonar-cxx-plugin\pom.xml](https://github.com/SonarOpenCommunity/sonar-cxx/blob/a1b1d8797dbfb9d00701d2b357fd4d4b7e59101c/sonar-cxx-plugin/pom.xml#L17)
[sonar-cxx-plugin\src\samples\SampleProject2\Readme](https://github.com/SonarOpenCommunity/sonar-cxx/blob/a1b1d8797dbfb9d00701d2b357fd4d4b7e59101c/sonar-cxx-plugin/src/samples/SampleProject2/Readme#L29)

Also found a number of instances of github.com/wenns that could be updated since you were replacing those with github.com/SonarOpenCommunity.

jmecosta:
right, ive updated all those now, and in my test server everything works properly.... thanks

guwirth:
@jmecosta can we close this?

PR #674, opened 2015-11-06
Update urls in pom.xml, wiki and plugin (SonarOpenCommunity)

jmecosta:
handles #670 , still not sure about the home page. need to try this


patrickjmccarty:
Looks like you missed updating the main <url> element on line 17?

PR #675, opened 2015-11-08
remove final and override from keyword list

### LINKED ISSUE #673: Failure to parse Botan header file Botan/build/include/botan/buf_comp.h
Using Botan v1.10.10 on Mac OS X I get a parse error when running analysis:

17:00:47.052 ERROR - Parse error at line 80 column 11:

   65: data()), str.size());
   66:          }
   67: 
   68: 
   69: 
   70: 
   71: 
   72:       void update(byte in) { add_data(&in, 1); }
   73: 
   74: 
   75: 
   76: 
   77: 
   78: 
   79: 
  -->        void final(byte out[]) { final_result(out); }
   81: 
   82: 
   83: 
   84: 
   85: 
   86: 
   87:       SecureVector<byte> final()
   88:          {
   89:          SecureVector<byte> output(output_length(

I commented out line 80, and it fails at line 87 instead.  Commenting out line 72 doesn't help either.

I'm unclear on what the issue is that is causing the parse error.

guwirth:
* final and override have a special meaning: http://en.cppreference.com/w/cpp/keyword and are no keywords
* close #673

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/api/CxxKeyword.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/api/CxxKeyword.java
@@ -24,8 +24,13 @@ import com.sonar.sslr.api.TokenType;
 
 /**
  * C++ Standard, Section 2.12 "Keywords"
+ *
+ * In this list are only C++ keywords allowed.
+ * All other extensions must be handled as identifiers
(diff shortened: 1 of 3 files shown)


guwirth:
Can I merge this?

jmecosta:
Think so.

PR #680, opened 2015-11-15
make non test code compile

jmecosta:
@guwirth , @Bertk started trying to upgrade this to 5.2 seems all our tests are going to failed because they started using java.nio.file.Path for DefaultInput that use in testing. we have seen earlier that this will result in files not found if the drive for example does not exist. 

do you have any ideas how we can proceed with this?


sonar-cxx-plugin/src/test/java/org/sonar/plugins/cxx/TestUtils.java
```diff
@@ -171,16 +172,16 @@ public static void addInputFile(DefaultFileSystem fs,
         temp.setCharAt(0, Character.toUpperCase(temp.charAt(0)));
         path = temp.toString();
       }
-      inputFile = new DefaultInputFile(path);
-      inputFile.setAbsolutePath(path);
+      inputFile = new DefaultInputFile("key", path);
```
jmecosta: @Bertk @guwirth this seems to be a problem for us? now they require DefaultInputFile to behave as a module project. this is going to brings many issues to us

jmecosta: and what about if we just mock everything related with filesystem, in the end we want to test the parser not really the file system. if doing it like this then test passes fine

sonar-cxx-plugin/src/test/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensorTest.java
```diff
@@ -132,17 +137,20 @@ public void shouldReportNoCoverageWhenInvalidFilesInvalid() {
   @Test
   public void shouldReportCoverageWhenVisualStudioCase() {
```
jmecosta: this is the one that filesystem gets mocked, it also simplifies a bit the windows/linux issues

sonar-cxx-plugin/src/test/java/org/sonar/plugins/cxx/TestUtils.java
jmecosta: for reference
https://github.com/SonarSource/sonarqube/blob/0ae5ddd039e29094eee818370fe34810d3a225f7/plugins/sonar-xoo-plugin/src/test/java/org/sonar/xoo/lang/SyntaxHighlightingSensorTest.java

pom.xml
```diff
@@ -129,10 +129,10 @@
   </ciManagement>
 
   <properties>
-    <license.title>Sonar C++ Plugin (Community) </license.title>
+    <license.title>Sonar C++ Plugin (Community)</license.title>
     <maven.test.redirectTestOutputToFile>true</maven.test.redirectTestOutputToFile>
-    <sonar.version>4.5.2</sonar.version>
-    <sslr.version>1.20</sslr.version>
+    <sonar.version>5.2</sonar.version>
```
guwirth: This change means we are no more compatible to LTS version.

sonar-cxx-plugin/src/test/java/org/sonar/plugins/cxx/TestUtils.java
guwirth: > and what about if we just mock everything related with filesystem

In the past we had a lot of issues with differences on Windows and Linux file system. Not sure if it is a good idea to mock this all away?

pom.xml
jmecosta: Yeah, this is more to give an idea what we will be facing when moving to next lts. So far loads of changes. Anyway we should not merge this until lts is available

sonar-cxx-plugin/src/test/java/org/sonar/plugins/cxx/TestUtils.java
jmecosta: I don't know? Should we not trust the platform will do its job and import all files accordingly?

Probably all this file import could be simplified. And the good thing is that we don't have any maintenance when they decide to change all the filesystem interfaces as they did for 5.2

pom.xml
guwirth: Looking to other plugins they seems to be alll still 4.5.2 compatible **and** working with SQ 5.2?
So this could still be 4.5.2?

guwirth:
@jmecosta would keep it 4.5 LTS compatible. If someone wanna see what's wrong with 5.2 he can change this locally. If you are aware of things we should change I would create issues for it.

jmecosta:
ok... sounds good.

Issue #677, opened 2015-11-09
SonarQube 5.2: configured rules not downloaded for analysis (using sonar-runner)


BobSilent:
Hi,

after switching to SonarQube 5.2 no c++ rules get downloaded from quality profile:
- installed plugin
- enabled additional Microsoft Compiler Rules 

when starting the Runner with -X option, and some additional debug information in plugin 
sonar-cxx-plugin\src\main\java\org\sonar\plugins\cxx\compiler\CxxCompilerSensor.java
```java
  public boolean shouldExecuteOnProject(Project project) {
	CxxUtils.LOG.info("shouldExecuteOnProject super: {}", super.shouldExecuteOnProject(project));
	
	CxxUtils.LOG.info("shouldExecuteOnProject getActiveRules size: {}", profile.getActiveRules().size());
	CxxUtils.LOG.info("shouldExecuteOnProject rulesRepositoryKey:  {}", getCompilerParser().rulesRepositoryKey());
	CxxUtils.LOG.info("shouldExecuteOnProject rulesRepositoryKey size:  {}", profile.getActiveRulesByRepository(getCompilerParser().rulesRepositoryKey()).size());
	boolean flag = !profile.getActiveRulesByRepository(getCompilerParser().rulesRepositoryKey()).isEmpty();
	CxxUtils.LOG.info("shouldExecuteOnProject flag:  {}", flag);
	
    return super.shouldExecuteOnProject(project)
      && flag;
  }
```

I get the following output in sonar 5.2
```
shouldExecuteOnProject super: true
shouldExecuteOnProject getActiveRules size: 510
shouldExecuteOnProject rulesRepositoryKey:  compiler-vc
shouldExecuteOnProject rulesRepositoryKey size:  0
shouldExecuteOnProject flag:  false
```

With Sonar 5.1.2 the output is the following:
```
shouldExecuteOnProject super: true
shouldExecuteOnProject getActiveRules size: 1041
shouldExecuteOnProject rulesRepositoryKey:  compiler-vc
shouldExecuteOnProject rulesRepositoryKey size:  531
shouldExecuteOnProject flag:  true
```

The difference in getActiveRules size is the number of c++ rule activated.

In addition I see in the Debug output the following information:
```
Download: http://localhost:9000/api/rules/search.protobuf?f=repo,name,severity,lang,internalKey,templateKey,params,actives&amp;activation=true&amp;qprofile=c++-12310&amp;p=1&amp;ps=500
[...]

PR #681, opened 2015-11-16
Feature/change language key

jmecosta:
covers #677, now this is working with 5.2

change to 5.2 api its a all different thing and we should not bother with it until actually some of the core plugins update.

also notice, that if a absolute path is used in reports path the lookup fails.


jmecosta:
@guwirth how can we get logs for the integration tests from travis?

guwirth:
@jmecosta What are the side effects of changing key **c++** to **cxx**?

jmecosta:
Probably to many, at least need to figure out why integration tests fail. For sure profiles need to be patched back into this version . but if we don't so this we definitely will not have a 5.2 compatible version

jmecosta:
Nice at least the ita are running on windows now :) I will take the logs as arrifacts

jmecosta:
@guwirth so made progress in the integration tests, just to find that running sonar as a service does not work properly. anyway we are now able to collect all logs when we have a failed build. at least older versions things should work better in travis. so those are what i found:

1. language=c++ needs to be removed, not a problem since its deprecated.
2. profiles must be restore and patched, that we new already
3. history at the point of transition will show -1000 remove +1000 added issues.

we could wait for 5.3 however we will not be supporting 5.2 at all. anyway we should use some of the things in this pr, at least the integration tests on windows part. also i want to update the python scripts to get the status of the analysis after the local analysis but all versions of sonar dont like to run in windows service so i need to do it locally only.

guwirth:
Hello @jmecosta thanks for testing. 

> history at the point of transition will show -1000 remove +1000 added issues.

Means your are loosing all user defined settings in the database, for example 'false positive'.

jmecosta:
@guwirth ive took the stuff that is important for now, its in appveyour and the support of absolute paths  for reports.

ive build 5.2 and apply the fix from 5.3 and now i can use 0.9.4 in there. so i am not stuck... this can be closed

PR #688, opened 2015-11-24
Improve the comment regarding the public api limitations

wichtounet:
The previous comment seemed to indicate that inline comments were not supported, but they are perfectly. This update makes the comment clearer by indicating that only comments by name in other files are not supported.


guwirth:
:+1:  Any other comments?

wichtounet:
No, the rest seem fine :)

lcintrat:
:+1:

Issue #693, opened 2015-11-30
CxxReportSensor.java: file not found for issue


BobSilent:
Hi,

the visualStudio c++ compiler writes the compiler warnings without absolute path information into the log file, if the source file (which has an issue) belongs to my vcxproj. 
Example
```
  aaa.cpp
  bbb.cpp
  ccc.cpp
C:\FULLPATH_TO_GLOBAL_INCLUDE\Include\ddd.h(xx): warning C4239: nonstandard extension used : 'argument' : conversion from 'CString' to 'CString &' (ccc.cpp)
          A non-const reference may only be bound to an lvalue
ccc.cpp(yyy): warning C4239: nonstandard extension used : 'argument' : conversion from 'CString' to 'CString &'
          A non-const reference may only be bound to an lvalue
```

unfortunately the source file ccc.cpp is not found, as in
CxxReportSensor.java (sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxReportSensor.java)

```java
  private void saveViolation(Project project, SensorContext context, String ruleRepoKey,
    String filename, String line, String ruleId, String msg) {
    Issuable issuable = null;
    int lineNr = 0;
    // handles file="" situation -- file level
    if ((filename != null) && (filename.length() > 0)) {
      String root = reactor.getRoot().getBaseDir().getAbsolutePath();
```

you are using 
```java
      String root = reactor.getRoot().getBaseDir().getAbsolutePath();
```
for resolving the root directory,
if I change it to 

```java
	  String root = fs.baseDir().getAbsolutePath();
```
everything can be resolved.
also other sensors, e.g. java use fs.baseDir
 sonar-java/java-surefire/src/main/java/org/sonar/plugins/surefire/api/SurefireUtils.java
 sonar-java/java-jacoco/src/main/java/org/sonar/plugins/jacoco/JaCoCoOverallSensor.java

Some other side aspects: i am using the following directory structure
```
\sonar-project.properties (*)
├───module1
│       module1.vcxproj (**)
│       sonar-project.properties
│
├───module2
│       module2.vcxproj
│       sonar-project.properties
│
├───module3
│       module3.vcxproj
│       sonar-project.properties
│
└───module4...
```
(*) one root project properties
(**) module properties, in parallel to the vcxproj

reactor.getRoot().getBaseDir().getAbsolutePath() is always "\" same as location of (*)
[...]

PR #683, opened 2015-11-19
allow to specify absolute paths for reports. to support msbuild runner

jmecosta:
support absolute path of report paths

PR #695, opened 2015-12-02
improved multi module support

guwirth:
test 1 for #693

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,4 +51,4 @@ script:
 
 after_failure:
         - cat $SONARHOME/logs/sonar.log
-        - find . -name _cpp-multimodule-project-2_.log | xargs cat
+        - find . -name "*.log" | xargs cat
--- a/integration-tests/testdata/cpp-multimodule-project/sonar-project.properties
+++ b/integration-tests/testdata/cpp-multimodule-project/sonar-project.properties
@@ -1,18 +1,22 @@
 sonar.host.url=http://localhost:9000
 
 sonar.modules=cli,lib,package1,package2
-cli.sonar.projectBaseDir=cli
-cli.sonar.sources=.
+
+cli.sonar.projectBaseDir=/home/travis/build/SonarOpenCommunity/sonar-cxx/integration-tests/testdata/cpp-multimodule-project/
+cli.sonar.sources=cli
 cli.sonar.projectName=cli
 cli.sonar.cxx.includeDirectories=../../googletest_project/tests/gtest-1.6.0/include/gtest
-lib.sonar.projectBaseDir=lib
-lib.sonar.sources=.
+
+lib.sonar.projectBaseDir=/home/travis/build/SonarOpenCommunity/sonar-cxx/integration-tests/testdata/cpp-multimodule-project/
+lib.sonar.sources=lib
 lib.sonar.projectName=lib
-package1.sonar.projectBaseDir=package1
-package1.sonar.sources=.
+
+package1.sonar.projectBaseDir=/home/travis/build/SonarOpenCommunity/sonar-cxx/integration-tests/testdata/cpp-multimodule-project/
+package1.sonar.sources=package1
 package1.sonar.projectName=package1
-package2.sonar.projectBaseDir=package2
-package2.sonar.sources=.
+
+package2.sonar.projectBaseDir=/home/travis/build/SonarOpenCommunity/sonar-cxx/integration-tests/testdata/cpp-multimodule-project/
+package2.sonar.sources=package2
 package2.sonar.projectName=package2
 
 sonar.projectKey=cpp-multimodule-project
@@ -41,7 +45,7 @@ sonar.cxx.xunit.reportPath=sonarcpp/reports-xunit/*.xml
 sonar.cxx.coverage.reportPath=sonarcpp/bullseye-coverage-results/bullseyecoverage-result-0.xml
 sonar.cxx.coverage.itReportPath=sonarcpp/bullseye-coverage-results/bullseyecoverage-result-1.xml
 sonar.cxx.coverage.overallReportPath=sonarcpp/bullseye-coverage-results/bullseyecoverage-result-2.xml
-                             
+
 sonar.cxx.defines=__cdecl, \n\
                 __declspec(a), \n\
                 __pragma(a), \n\
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/compiler/CxxCompilerSensor.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/compiler/CxxCompilerSensor.java
@@ -35,7 +35,6 @@ import org.sonar.api.resources.Project;
 import org.sonar.plugins.cxx.utils.CxxMetrics;
 import org.sonar.plugins.cxx.utils.CxxReportSensor;
 import org.sonar.plugins.cxx.utils.CxxUtils;
-import org.sonar.api.batch.bootstrap.ProjectReactor;
 
 /**
  * compiler for C++ with advanced analysis features (e.g. for VC 2008 team edition or 2010/2012/2013 premium edition)
@@ -56,8 +55,8 @@ public class CxxCompilerSensor extends CxxReportSensor {
   /**
    * {@inheritDoc}
    */
-  public CxxCompilerSensor(ResourcePerspectives perspectives, Settings settings, FileSystem fs, RulesProfile profile, ProjectReactor reactor) {
-    super(perspectives, settings, fs, reactor, CxxMetrics.COMPILER);
+  public CxxCompilerSensor(ResourcePerspectives perspectives, Settings settings, FileSystem fs, RulesProfile profile) {
+    super(perspectives, settings, fs, CxxMetrics.COMPILER);
     this.profile = profile;
 
     addCompilerParser(new CxxCompilerVcParser());
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
@@ -29,7 +29,6 @@ import java.util.Map;
 import javax.xml.stream.XMLStreamException;
 
 import org.sonar.api.batch.SensorContext;
-import org.sonar.api.batch.bootstrap.ProjectReactor;
 import org.sonar.api.batch.fs.FileSystem;
 import org.sonar.api.batch.fs.FilePredicates;
 import org.sonar.api.batch.fs.InputFile;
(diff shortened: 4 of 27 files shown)


guwirth:
Failing:
Feature: cpp-multimodule-project
Test multimodule project with reports at root of the project

guwirth:
@all: had a closer look to this yesterday. Currently we are using the root folder as baseDir.(https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Path-and-path-separator-issues)

> Import reports in multi module configuration
From version 0.9.3 all reports are looked from the project root, instead of the module base dir. This change makes the plugin behavior consistent between all types of different configurations.

This is more or less a copy from the C# plugin pattern.

Looking to https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxReportSensor.java there is different behaviour for ```analyse``` and ```saveViolation```:

```Java
public void analyse(Project project, SensorContext context) 
...
List<File> reports = getReports(conf,
          reactor.getRoot().getBaseDir().getCanonicalPath(),
          fs.baseDir().getPath(),
...
```

```Java
private void saveViolation(Project project, SensorContext context, String ruleRepoKey,
    String filename, String line, String ruleId, String msg) 
...
String root = reactor.getRoot().getBaseDir().getAbsolutePath();
...
```

Think we have to unify this behaviour for both cases. I also think the search order is also wrong:
1) should be first ``` fs.baseDir().getPath()```
2) then ```reactor.getRoot().getBaseDir().getCanonicalPath()```

What are you thinking?

jmecosta:
I was already thinking that it must be some thing different between reports
and save because the report contains both search. This is in the PR for the
absolute paths.

I can apply the module first and root second on that PR, maybe it's worth
to apply the file search there also and create its.

It would be great someone could prepare a sample project that uses the
multi modules in different way that we have the ir

guwirth:
Think we have two support the following cases:
* absolute and relative paths for reports in sonar-project.properties
* absolute and relative file items in report files
* in case of relative paths & multi module project:
 * relative to module baseDir
 * relative to module baseDir

==> 8 combinations

jmecosta:
@guwirth the reactor needs to be used imo, there is no other way of finding what is the project root without using it. so the case where we look for reports from the root needs to be supported since it was the only supported case so far and the most users that use multi modules are using this approach,

to be is loads for setup if you need to generated reports for each modules, its much easier to generate reports for all the project and then just put them in a single location in root of the project

guwirth:
> the reactor needs to be used imo, there is no other way of finding what is the project root without using it.

@jmecosta In case we are supporting absolute paths for defining report file locations what's the difference?

```INI
# baseDir = x:/root
sonar.cxx.XXX.reportPath=x:/root/reports/xxx-*.xml
sonar.cxx.XXX.reportPath=x:/outside/reports/xxx-*.xml
```

jmecosta:
@guwirth usually its not a good idea to hard code absolute paths in props files, it will only work in the machine where it was created. for example our ci system uses Base Paths that change according to git revision... so hard coding those in props will not work for us. we would need to pass those via arguments so that they overwrite what exist in props file

i mean absolute paths are ok, but it must be a machine passing those to the analysis via -D parameters. imo we are supporting this case to enable such flexibility and not to make it a default padron that everyone uses

guwirth:
@jmecosta with the *flat* modell the paths in the report files has to be absolute. Means reports working only if they fit to a specific machine. So have one more absolute path doesn't matter?

jmecosta:
We use relative paths for reports in flat mode, no problems. Perhaps we
talk about differwnt things?

guwirth:
> We use relative paths for reports in flat mode, no problems

On #697 you said:  flat (relative to root + absolute)

In case flat is relative then there is still the question in case of multi module:

> #697 ... reactor... In case of using always root baseDir: If there are realtive paths in report files are they relative to root baseDir or current baseDir?

jmecosta:
I'm refering to the patterns we use to search reports, not the content of
the report it self. The patterns are hard coded in property file and making
it absolute will not work in ci system.

The path inside report indeed have to be absolute, but those are generated
during analysis and we have our tools setup to report only in absolute paths

jmecosta:
To be honest I would not support cases where reports contain only relative
paths there are to many ambiguities. It's easy to tell a static analyser to
produce issues with absolute paths . we should however support absolute
paths for report search path

Bertk:
I agree the absolute paths shall be supported for reports but we should also check the performance and maybe introduce a file/path interface for the plug-in. I found this comment and I have low confidence that this will be improved: 
```
FilePredicate org.sonar.api.batch.fs.FilePredicates.hasAbsolutePath(String s)

    Predicate that gets a file by its absolute path. The parameter accepts forward/back slashes as separator and non-normalized values (/path/to/../foo.txt is same as /path/foo.txt). 

Warning - not efficient because absolute path is not indexed yet.
```
We should also consider to support a simple, comma separate list of paths for a report configuration item. The ant style search pattern is not a common way.
By the way, I did a small update to support multiple test reports with absolute paths for a component (see 24ed1e24f8df25ec8ca31a70fc545e29c6283e78).

jmecosta:
so we agree on the need for absolute paths for reports, #683 should handle it.  24ed1e24f8df25ec8ca31a70fc545e29c6283e78 looks to me that it does not support /abc/def/*.xml  however the "," can be added later once the main lookup is clarified and merged to master.

now for allowing relative paths inside reports to raise issues, the only hard rule that we can apply is disallow it (the performance i think it should not be a big issue because we have been using this so far no one as complained?). if we really want to support it then imo would be, relative to project root and if not found relative to module however this would not use the same algorithim as for looking reports but use only the functions provided by the plaform (so that it does not make it slower)

another thing regarding the relative paths is that, we dont know what users are using as base dir to generate reports. file x can be relative to anything, even to a folder outside project. soon they will be asking why my relative path does not work if we support it.

integration-tests/features/multimodule_analysis.feature
```diff
@@ -1,52 +1,51 @@
...
-               # violations
-               | violations               | 28    |
-               # test execution statistics
-               | test_success_density     | 33.3  |
-               | test_failures            | 2     |
-               | skipped_tests            | 1     |
-               | test_errors              | 0     |
-               | tests                    | 3     |
-
+# Feature: cpp-multimodule-project
```
jmecosta: i am ensure about this, with this change we are telling people that they must to have their reports always per module and not once at the root of the repo as it has been supported until now. as far as i know this work is mostly manual and you need to generate reports per module instead of generating the reports once for the all repo.

although this might be ok for some, i think we should also support the lazy people that just put reports on root of the repository and let the plugin deal with it. as far as i know the only way of doing it is with the reactor, so i think we should not remove it

guwirth:
@jmecosta this PR is only to play around and see what is needed to support the [SQ default behavior](http://docs.sonarqube.org/plugins/servlet/mobile#content/view/6948272) for multi module.

What I like on this solution is that the code is very simple, SQ is supported as described on there page and also absolute and relative path support is included in this solution. What is not working is "our" reactor mode having the reports all relativ to the root folder. This could be solved by setting absolute paths in the root BaseDir config (I know you don't like it because path could be different on different servers). Here I would like to come back to a suggestion I made already in the past: why not supporting environment variables in the config file? (I know that there are plans to support this in sonar runner 2.5 anyway). 

What I don't like on your approach #683 currently: it sounds to much like a msbuild runner solution only. I think there are still many other users like to use this plugin without VS and without msbuild runner. We should not reduce this plugin to a msbuild runner extension. (What I like on your approach: code and structure is better as existing one.)

Anyway: as long as your solution is still supporting the "old" use cases we should merge it as soon as the existing unit and integration tests are green (**before this one**). We can improve/generalize things always later one.

I will also try to add some multi module IT tests in a separate PR to see if default SQ modes are supported.

jmecosta:
the reactor is only used for looking up reports from root of the project, its of no use to the msbuild runner. 

For the msbuild runner the only thing required is to support absolute paths (and those need to be hacked). the msbuild changes this ideas quite a lot, the reports are looked up by default like this.
<ROOT>\.sonarqube\out\a1a__Win32_Debug_635855202494571037\analysisresults.xml

12:31:04.669 INFO  - Base dir: D:\Development\SonarQube\cxx\sonar-cxx\integration-tests\testdata\cpp-multimodule-project-2
12:31:04.669 INFO  - Working dir: D:\Development\SonarQube\cxx\sonar-cxx\integration-tests\testdata\cpp-multimodule-project-2\.sonarqube\out\.sonar

At least when i tried the first time, all reports that were under 
<ROOT>\cxxresults\*.xml were picked up from the .sonarqube folder above. this is the reason i started to implement the support of absolute paths.

... just to be sure i really dont use the reactor myself, but i assume that some users do. because its kind of easy to keep those there. (an example, cppcheck works better if you run on root of the directory)

but anywhy i like your suggestion to support this special variable, on the file. but this definetly can be done once the absolute report path is done.

i will get #683 ready, requires unit tests to be re-factored. and then we can review it and if its ok to merge then we can discuss the removal of the rector?

guwirth:
> i will get #683 ready, requires unit tests to be re-factored. and then we can review it and if its ok to merge 

:+1:

jmecosta:
Caused by: java.lang.IllegalStateException: You must define the following mandatory properties for 'cpp-multimodule-project-2:A:1:a': sonar.sources

not sure what is this, works fine in my machine in 5.1.2, 

12:54:58.708 INFO  - Base dir: D:\Development\SonarQube\cxx\sonar-cxx\integration-tests\testdata\cpp-multimodule-project-2\A\1\a
12:54:58.708 INFO  - Working dir: D:\Development\SonarQube\cxx\sonar-cxx\integration-tests\testdata\cpp-multimodule-project-2\.sonar\cpp-multimodule-project-2_A_1_a
12:54:58.708 INFO  - Source paths: .

supose need to simplify the project for this.

guwirth:
This PR is now ready to merge. The reactor is removed but with a modified configuration file the 'reactor multi project' is still working.

This solution is now much closer to the [SQ default behavior](http://docs.sonarqube.org/display/SONAR/Analyzing+with+SonarQube+Scanner#AnalyzingwithSonarQubeScanner-Multi-moduleProject)  and there is no more 'magic' functionality inside the plugin. The user has to define projectBaseDir explicit - in reactor use case absolute.

One thing is a little bit disappointing: It's not possible to set a placeholder for projectBaseDir. The sonar runner is checking the value in the [EmbeddedRunner](https://github.com/SonarSource/sonar-runner/blob/master/sonar-runner-api/src/main/java/org/sonar/runner/api/EmbeddedRunner.java) long before the plugin can see it (see https://github.com/SonarSource/sonar-runner/blob/master/sonar-runner-api/src/main/java/org/sonar/runner/api/Dirs.java). Don't know if this is possible to modify? Will create a issue in the SQ newsgroup.

@Bertk you are using this PR already?
@jmecosta ok?

jmecosta:
Ok

PR #698, opened 2015-12-05
remove CompileIncludePathNotFoundOrInvalide Rule

jmecosta:
this rule is out of the scope of the plugin, its trying to raise issues at project level taking the information from the vc++ buildlog. this is even failing the analyis in 5.2 and has nothing to do with C++ code

there are alternatives for this, for example https://github.com/TeklaCorp/sonar-msbuild-plugin contains already this rule and will raise these issues against the msbuild file that is defining the compilation


jmecosta:
@guwirth i think @Bertk agrees with me that this should not be in the plugin. right?

guwirth:
Think there are two reasons a include can't be found:
* invalid include: compiler can find it
* invalid cxx plugin configuration, wrong sonar.cxx.includeDirectories setting

jmecosta:
This is a include folder used by msbuild. We already have another rule to
detect cxx plugin misconfiguration.

guwirth:
:+1:

Bertk:
The SQ 5.2 problem is fixed with this PR :+1:

jmecosta:
ready to merge?

Issue #700, opened 2015-12-06
Travis build for Java 7 failing


guwirth:
Travis build with Java 7 is failing since 5.12.2015.

jmecosta:
do we have no access to sonar.log to check why sonar is not starting?

guwirth:
Restarted some of the older successful builds to see if this is an Travis issue or in our code,

> do we have no access to sonar.log to check why sonar is not starting?

No, We should add something like this to get more info
```
after_failure:
  - cat target/surefire-reports/*.txt
```

jmecosta:
also cat /tmp/sonarqube-4.5.6/logs.sonar.log since its the one that is failing

guwirth:
Seems to be a Travis problem with Java 7. Also older revisions which were successful before are failing?

guwirth:
```
mail to Travis support
```
Dear support team,

Starting last week we have problems with our CI on Travis. Tests with Java 7 are no more working:
https://travis-ci.org/SonarOpenCommunity/sonar-cxx

Is there a known issue/changes with the Java 7 support?

What we already tried:
We did already a test with old green PRs: they are red now.

Error log:
https://travis-ci.org/SonarOpenCommunity/sonar-cxx/jobs/95241426 (at the end)
```
The command "RAILS_ENV=production PATH=$PATH:/tmp/sonar-runner-2.4/bin behave" exited with 255.
after_failure

0.00s$ cat $SONARHOME/logs/sonar.log
--> Wrapper Started as Daemon

Launching a JVM...
Wrapper (Version 3.2.3) http://wrapper.tanukisoftware.org
  Copyright 1999-2006 Tanuki Software, Inc.  All Rights Reserved.
*** buffer overflow detected ***: java terminated

======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(__fortify_fail+0x37)[0x7f4dd9355e57]
/lib/x86_64-linux-gnu/libc.so.6(+0x108d50)[0x7f4dd9354d50]
/usr/lib/jvm/java-7-openjdk-amd64/jre/lib/amd64/libnet.so(Java_java_net_Inet4AddressImpl_getLocalHostName+0x190)[0x7f4dd4213960]
[0x7f4dcd012d98]

======= Memory map: ========
00400000-00401000 r-xp 00000000 08:01 271208                             /usr/lib/jvm/java-7-openjdk-amd64/jre/bin/java
[...]

PR #702, opened 2015-12-06
add sonar log file content to Travis log after failure

guwirth:
related to #700


jmecosta:
looks good, usually the it also create logs during analysis. i think we could also report those. but maybe  its going to flood the log? and only the failed log should be reported

jmecosta:
cat: /tmp/sonarqube-5.1.2/logs/sonar.log: No such file or directory 

is it possible that sonar is not being donload/unziped? or are we remo

PR #707, opened 2015-12-12
add SQ 5.3-RC1 to Travis CI testing

guwirth:


jmecosta:
5.2 will never support the current plugin, so we can remove that. if they release 5.2.1 then it can be added. 

for 5.3 the current its will require a modification to wait for server analysis to end before checking the analysis, but otherwise it should be ok

jmecosta:
and it seems it requires bigger refactoring to support the new rest api.

        File "integration-tests/features/steps/test_execution_statistics.py", line 206, in assert_measures
          json_measures = json.loads(response.text)[0].get("msr", None)
      KeyError: 0

guwirth:
@jmecosta merge this anyway to see the current state. 

> and it seems it requires bigger refactoring to support the new rest api.

Have less knowledge here. Can you do this?

jmecosta:
Yep, its OK to merge. I will take a look during next week

PR #716, opened 2015-12-19
Support environment variables and system properties in configuration files

guwirth:
* introduce new class CxxSettings
* overload Settings.getClearString to replace environment variables in values
* replace all Settings with CxxSettings

todo: unit test

@all: any feedback to this approach?

### DIFF
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxSettings.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxSettings.java
@@ -1,0 +1,110 @@
+/*
+ * Sonar C++ Plugin (Community)
+ * Copyright (C) 2010 Neticoa SAS France
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package org.sonar.plugins.cxx;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.sonar.api.config.Settings;
+
+/**
+ * CxxSettings supports variables in configuration files
+ *
+ * - CxxSettings clones Settings in ctor
+ * - format for placeholder is ${xxx} - supported are environment variables,
+ *   Java system properties and SonarQube properties
+ * - backslashes in values from environment variables and Java system properties
+ *   are replaced with slashes to support Windows paths
+ */
+public class CxxSettings extends Settings {
+
+  private final HashMap<String, String> vars = new HashMap<>();
+  private final Pattern regex = Pattern.compile("\\$\\{(.+?)\\}");
+
+  /**
+   * Clone settings.
+   */
+  public CxxSettings(Settings other) {
+    super(other);
+
+    Map<String, String> envMap = System.getenv();
+    for (Entry<String, String> entry : envMap.entrySet()) {
+      String key = entry.getKey();
+      String value = entry.getValue();
+      value = value.replace("\\", "/");
+      vars.put(key, value);
+    }
+
+    Properties props = System.getProperties();
+    for (String key : props.stringPropertyNames()) {
+      String value = props.getProperty(key);
+      value = value.replace("\\", "/");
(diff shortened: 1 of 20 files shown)


jmecosta:
Try to to support also vars that are passed as arguments.

For example

-DRoot=c:\src

Usually those are more useful than environment variables.  Both should be coded. Suppose the you can use get clear property to retrieve the the props that are ${} before looking into the environment

guwirth:
@jmecosta good idea will try. Do you know how sonar runner is handling this defines? 
E.g. - DRoot=c:\src is this an environment variable afterwords or can I use an API to get them?

jmecosta:
I dont think it matters, we they reached the plugin I think they are already parsed properly

jmecosta:
I think you can use the settings class to get those.

guwirth:
This PR is always crashing on server side.
* first try was to replace all Settings with CxxSettings: server crash during start up
* next try to move and use CxxSettings only internally: crash during configuration creation

@all someone experience with this? Is it possible to extend Settings? Is there something I have to register on server side?

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxSettings.java
```diff
@@ -0,0 +1,65 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package org.sonar.plugins.cxx;
+
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Properties;
+import org.sonar.api.config.Settings;
+
+public class CxxSettings extends Settings {
```
jmecosta: wonder if the internals of Settings will be in proper state when it arrives to the plugin. i would assume core to create its own instance of Settings class, likely when CxxSettings gets into the plugin it comes without any information about the settings.

perhaps this class cannot be used like this, and might explain the issue in server side. Could this be done in a helper class that receives Settings and then returns the value as per you implementation below?

guwirth:
This version is working. Maybe someone more familiar with Java can have a look to it.

jmecosta:
Looks good to me, I don't know what are the java patterns that should be used in this case but for me it looks functional

guwirth:
The point with this solution is, that the `CxxSettings(Settings settings)` Ctor creates a **copy** of Settings. If someone is changing the values of the original Settings afterwards this will not affect CxxSettings. I'm not sure if this is relevant (except unit tests) because after starting an analysis the Settings have to be stable anyway?

Other approach could be to use inside CxxSettings a reference and the pimpl pattern. But this means all methodes must be duplicated.

guwirth:
Ready to merge :+1: 

Support environment variables and system properties in configuration files: use ${xxx} as placeholder

jmecosta:
Let me test it first. Try to do it before end of year.

guwirth:
Ok. Happy xmas. 

Von meinem iPhone gesendet

> Am 24.12.2015 um 10:02 schrieb Jorge Costa <notifications@github.com>:
> 
> Let me test it first. Try to do it before end of year.
>

jmecosta:
Merry Xmas also..

guwirth:
@jmecosta did you have some time for testing? Can I merge this?

jmecosta:
i will do it today.

jmecosta:
@guwirth few things i spotted:

if passing -DAnalysisReportsDir=D:\SonarVersions\MultLanguageProject\.cxxresults i dont get the parser to replace the AnalysisReportsDir

environment variables work, however if using 
set AnalysisReportsDir=D:\SonarVersions\MultLanguageProject\.cxxresults

what i get in the plugin is:
D:SonarVersionsMultLanguageProject.cxxresults

so it does not support the windows separators.

guwirth:
😕 thx for testing. Will try to improve it.

guwirth:
> if passing -DAnalysisReportsDir=D:\SonarVersions\MultLanguageProject.cxxresults i dont get the parser to replace the AnalysisReportsDir

There are three types to replace:
1. SQ properties from configuration file or -D parameter
2. Java system properties
3. environment variables

Currently I replace only 2 & 3. Will add also 1. 

> environment variables work, however if using 
> set AnalysisReportsDir=D:\SonarVersions\MultLanguageProject.cxxresults
> what i get in the plugin is:
> D:SonarVersionsMultLanguageProject.cxxresults

In case of 2 & 3 I have to replace \ with \\\ or / to behave like defined in config file.

jmecosta:
Ok so its possible to support the windows paths, I know that using the
sonar runner this would not work. If you are able to get it working then
would be really nice

guwirth:
> -DAnalysisReportsDir=D:\SonarVersions\MultLanguageProject

For this case you have to define it like in config file. Can't replace \ with / because I don't know which properties are from config file and which are from command line.

works now:
```
-DAnalysisReportsDir=D:/SonarVersions/MultLanguageProject
-DAnalysisReportsDir="D:/SonarVersions/MultLanguageProject"
```

> environment variables work, however if using set AnalysisReportsDir=D:\SonarVersions\MultLanguageProject.cxxresults what i get in the plugin is: D:SonarVersionsMultLanguageProject.cxxresults

It's working now.

guwirth:
CxxSettings supports variables in configuration files …

- CxxSettings clones Settings in ctor
- format for placeholder is ${xxx}
- supported are environment variables, Java system properties and SonarQube properties
- backslashes in values from environment variables and Java system properties are replaced with slashes to support Windows paths

jmecosta:
Ok I will test again in a bit. Thanks

jmecosta:
works fine now, thanks. for me it can be merged

PR #719, opened 2015-12-27
refactoring

guwirth:
* unify formatting
* add @todo for deprecated APIs
* add @Override, diamond operator


guwirth:
@jmecosta & @Bertk is it ok for you to merge this? It's only beautifying.

jmecosta:
Ok

Bertk:
:+1:
I recommend to create a SQ profile which can be used to check the format of the sources. Ideally this can be used within CI but until now we do not have a SQ instance for this. Nevertheless the SQ profile can be used locally to check the updates before a PR is created.

jmecosta:
We can setup the github plugin in appveyour?

Issue #725, opened 2015-12-29
CUDA kernel function parser error


Bertk:
@guwirth , @jmecosta 
I want to analyze CUDA source code and look for a solution to handle the CUDA C syntax extension for kernel functions e.g. ```kernel<<< gridDim, blockDim, 0 >>>(d_data, height, width);```.
The other CUDA extensions could be handled by pre-processor defines but this one need to be handled by the CXX grammar and I need some hints where to start (className rule???). 

I use some sources from https://en.wikipedia.org/wiki/CUDA (https://github.com/Bertk/sonar-cxx/commit/953873eae31fd64733027f2b548dd396235ae492) and analyze the *Nvidia CUDA Samples v7.5*. 

Now I am stuck and any recommendations are appreciated.

additional information: http://docs.nvidia.com/cuda/cuda-c-programming-guide/#kernels

guwirth:
```C++
    // Kernel invocation with N threads
    VecAdd<<<1, N>>>(A, B, C);
```
This CUDA extension looks like a template with additional brackets. As we had already a lot of trouble with the double >> I think it's not easy. My gut feeling:

Try to introduce '<<<' and '>>>' as own rule (but not sure if this will not conflict with templates: e.g. A<B<C<c,1>2>>>), e.g.:
```
b.rule(cuda).is(b.sequence("<<<", IDENTIFIER, ">>>"));
```
Afterwards it looks like a function call: `cuda(a,b,c)`?
```
b.rule(postfixExpression).is(...)
```

Bertk:
@guwirth Yes, I followed the discussion on the template parsing. I am not sure where the ```b.rule(postfixExpression).is(...)``` should be added within the ```void expressions(LexerfulGrammarBuilder b)``` grammar. I guess the parser uses ```simpleTemplateId``` instead of ```IDENTIFIER```. Do you know a way to how to debug this?

By the way, the C++ standard (n4527) defines 
```
class-name:
          identifier
          simple-template-id
```

but we use : 
```java
      b.firstOf(
        simpleTemplateId,
        b.sequence(b.optional("::"), IDENTIFIER)
        )
      );
```

guwirth:
Reading a little bit more about CUDA: http://docs.nvidia.com/cuda/cuda-c-programming-guide/#um-language-integration ...access to additional language integration features, such as shared symbol names and inline kernel launch via the <<<...>>> operator.

[...]

PR #727, opened 2016-01-01
support cudaKernel grammar

Bertk:
- added optional rule to postfixExpression
- "b.rule(cudaKernel).is(b.sequence("<<", "<",
b.optional(expressionList), ">>", ">"));"

see also #725


guwirth:
:+1:

Issue #734, opened 2016-01-02
SQ-5.2 background task: java.lang.IllegalArgumentException: Rule 'cxx:CycleBetweenPackages' 


Bertk:
I observed an issue with SQALE remediation function for 'cxx:CycleBetweenPackages'. I disabled the rule and the results were loaded successfully to SQ DB.
@guwirth, @jmecosta: Did you also see this issues?
```
2016.01.02 16:33:47 ERROR [o.s.s.c.t.CeWorkerRunnableImpl] Failed to execute task AVIC-GwUbIrdtFH5k1IO
java.lang.IllegalStateException: Fail to process issues of component 'CudaSamples:CudaSamples:29A96CA4-8088-3B7B-E13E-540C8DB178CC:boost/config/abi_prefix.hpp'
	at org.sonar.server.computation.issue.IntegrateIssuesVisitor.processIssues(IntegrateIssuesVisitor.java:74) ~[sonar-server-5.2.jar:na]
	at org.sonar.server.computation.issue.IntegrateIssuesVisitor.visitAny(IntegrateIssuesVisitor.java:59) ~[sonar-server-5.2.jar:na]
	at org.sonar.server.computation.component.TypeAwareVisitorWrapper.visitAny(TypeAwareVisitorWrapper.java:83) ~[sonar-server-5.2.jar:na]
	at org.sonar.server.computation.component.VisitorsCrawler.visitNode(VisitorsCrawler.java:97) ~[sonar-server-5.2.jar:na]
	at org.sonar.server.computation.component.VisitorsCrawler.visit(VisitorsCrawler.java:80) ~[sonar-server-5.2.jar:na]
	at org.sonar.server.computation.component.VisitorsCrawler.visitChildren(VisitorsCrawler.java:90) ~[sonar-server-5.2.jar:na]
	at org.sonar.server.computation.component.VisitorsCrawler.visit(VisitorsCrawler.java:77) ~[sonar-server-5.2.jar:na]
	at org.sonar.server.computation.component.VisitorsCrawler.visitChildren(VisitorsCrawler.java:90) ~[sonar-server-5.2.jar:na]
	at org.sonar.server.computation.component.VisitorsCrawler.visit(VisitorsCrawler.java:77) ~[sonar-server-5.2.jar:na]
	at org.sonar.server.computation.component.VisitorsCrawler.visitChildren(VisitorsCrawler.java:90) ~[sonar-server-5.2.jar:na]
[...]

PR #735, opened 2016-01-03
fix rule CycleBetweenPackages Sqale remediation parameters

Bertk:
see #734


guwirth:
Wait for

> The cxx-checks sqale definitions should be removed from sonar-cxx-plugin/src/main/resources/com/sonar/sqale/cxx-model.xml.

before I merge this?

Bertk:
The cxx-model.xml file does not contain any cxx-checks definitions. This was removed before with https://github.com/guwirth/sonar-cxx/commit/b49577b556ab5bd72e205a6205dcf7a1629646ab.
I also deleted the obsolete cxx-model-project files because they are not in sync with cxx-model.xml.

jmecosta:
Ok that's good, so all should be oknow ?

Bertk:
I am not sure. There are other checks which use ```@SqaleConstantRemediation``` as well and this is a potential background tasks exception (SQ 5.2 or 5.3). I have not all checks activated but some of them and they did not create an exception.
So for now - I assume the issue is fixed and only 'cxx:CycleBetweenPackages' created an exception. Unfortunately I do not know why this is limited to 'cxx:CycleBetweenPackages'
@guwirth: When will you create the next RC? We should now ready to target SQ5.3 aren't we?

guwirth:
@Bertk over AppVeyor there is always the latest plugin available. I did no tests with 5.3 up to now and integration tests are also red. The issue list is also still very long :-(

guwirth:
PR :+1:

PR #741, opened 2016-01-04
improve variables support in configuration file

guwirth:
add CxxProjectBuilder to replace placeholder earlier
its only one single place to do this then


guwirth:
improve #716

PR #753, opened 2016-01-09
build with SQ 5.3 final version

guwirth:


jmecosta:
seems we can give another try to appveyour its,
https://jira.sonarsource.com/browse/SONAR-7140 its included in this version

please merge so i can rebase

Issue #775, opened 2016-02-11
Cannot find reports with relative Ant globbing


mjdetullio:
Environment:
- SQ 5.3
- 2ba787e23cb791a414391150fa269b45fa38e454 (latest from master at time of writing)
- Multi-module project

This multi-module project is analyzed using the SQ Gradle Scanner.  When all said and done, the settings end up looking something like this:

```
sonar.modules=module1,module2
...
module2.sonar.sources=dir1,dir2,dir3,dir4
module2.sonar.exclusions=**/build/**,**/cppcheck.xml
module2.sonar.cxx.cppcheck.reportPath=**/cppcheck.xml
```

The reports most certainly exist in the tree (and they have correct ownership and 0644 perms):

```
$ find . -name cppcheck.xml
./module2/dir1/build/cppcheck.xml
./module2/dir2/cppcheck.xml
./module2/dir3/build/cppcheck.xml
./module2/dir4/build/cppcheck.xml
```

But here's what I get in the analysis:

```
-------------  Scan module2
Base dir: /path/to/project/module2
Working dir: /path/to/project/build/sonar/com.mycompany.project_module2
Source paths: dir1, dir2, dir3, dir4
Source encoding: ISO-8859-1, default locale: en_US
Index files
Excluded sources: 
  */build/**
  **/cppcheck.xml
82 files indexed
...
omitted for brevity
...
Sensor CxxCppCheckSensor
Using pattern '**/cppcheck.xml' to find reports
Unprocessed root directory '/path/to/project/module2'
Unprocessed report file '**/cppcheck.xml'
Processed root directory '/path/to/project/module2'
Processed report file '**/cppcheck.xml'
Cannot find a report for 'sonar.cxx.cppcheck.reportPath=**/cppcheck.xml'
Sensor CxxCppCheckSensor (done) | time=29ms
```

As you can see, the property makes its way to the sensor correctly but something goes wrong when actually searching for the files.  There are no exceptions when walking the tree.  I don't know exactly where the problem in the code is without attaching a debugger, but it's within this stack: https://github.com/SonarOpenCommunity/sonar-cxx/blob/2ba787e23cb791a414391150fa269b45fa38e454/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxUtils.java#L119 with ```CxxFileFinder.FindFiles``` turning up an empty Collection.

jmecosta:
Hi

[...]

PR #780, opened 2016-02-16
Fix #766: Handle relative report paths with backticks outside basedir.

### LINKED ISSUE #766: NullPointerException looking for CppCheck report
I'm getting a NPE when I attempt to run analysis on my project, with sonar-cxx 0.9.3.

```
INFO: Sensor CxxCoverageSensor (done) | time=1526ms
INFO: Sensor CxxCppCheckSensor
DEBUG: Using pattern 'null' to find reports
INFO: Analysis report generated in /home/gitlab-runner/builds/580bbdff/0/hidden/path/src/.sonar/batch-report
INFO: ------------------------------------------------------------------------
INFO: EXECUTION FAILURE
INFO: ------------------------------------------------------------------------
INFO: Total time: 15.928s
INFO: Final Memory: 41M/110M
INFO: ------------------------------------------------------------------------
ERROR: Error during SonarQube Scanner execution
org.sonar.api.utils.SonarException: Cannot feed the data into sonar, details: 'java.lang.NullPointerException'
        at org.sonar.plugins.cxx.utils.CxxReportSensor.analyse(CxxReportSensor.java:134)
        at org.sonar.batch.phases.SensorsExecutor.executeSensor(SensorsExecutor.java:58)
        at org.sonar.batch.phases.SensorsExecutor.execute(SensorsExecutor.java:50)

.... snip ....

Caused by: java.lang.NullPointerException
        at org.apache.tools.ant.DirectoryScanner.setIncludes(DirectoryScanner.java:569)
        at org.sonar.plugins.cxx.utils.CxxReportSensor.getReports(CxxReportSensor.java:166)
        at org.sonar.plugins.cxx.utils.CxxReportSensor.analyse(CxxReportSensor.java:104)
        ... 31 more
```
I _am_ setting the report path in my properties file:
```
sonar.cxx.cppcheck.reportPath=../out/doc/cppcheck/*.xml
```

It looks like there's been a lot of refactoring going on in your repository here so the code that caused this may no longer exist, but maybe this problem is still there. In the code that went into 0.9.3, I'm not sure what would cause reportPath to be null, because (1) I'm setting it in my config, and (2) it's getting set to the default if it is null.

mjdetullio:
It was possible to have an NPE when normalizing backticks on relative
paths if the backticks moved the directory too far back.  This change
will handle invalid normalization of relative paths.

Normalization is reattempted against the absolute path, and if it is
still invalid then it is not given to the directory scanner, which does
not accept null elements.

### DIFF
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxReportSensor.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxReportSensor.java
@@ -25,9 +25,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.ArrayList;
 import java.util.Set;
-import javax.annotation.Nullable;
-import com.google.common.base.Function;
-import com.google.common.collect.Lists;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.tools.ant.DirectoryScanner;
 import org.sonar.api.batch.Sensor;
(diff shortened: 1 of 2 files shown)


sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxReportSensor.java
```diff
@@ -152,24 +149,30 @@ protected String getStringProperty(String name, String def) {
...
-          return FilenameUtils.normalize(moduleBaseDir.getAbsolutePath() + File.separator + path);
+    List<String> reportPaths = Arrays.asList(settings.getStringArray(reportPathPropertyKey));
+    if (!reportPaths.isEmpty()) {
+      List<String> includes = new ArrayList<>();
+      for (String reportPath : reportPaths) {
+        // Normalization can return null if path is null, is invalid, or is a path with back-ticks outside known directory structure
+        String normalizedPath = FilenameUtils.normalize(reportPath);
+        if (normalizedPath != null && new File(normalizedPath).isAbsolute()) {
+          includes.add(normalizedPath);
+          continue;
```
guwirth: @mjdetullio I'm not a fan of continue in loops. Think if / else if / else would be the same ....

guwirth:
fix #766 together with #775

PR #788, opened 2016-02-27
 add metric COMPLEXITY_IN_CLASSES and COMPLEXITY_IN_FUNCTIONS

### LINKED ISSUE #747: SQ 5.4: now requires the metrics complexity_in_classes  and complexity_in_functions
**Problem: value of class complexity is different with SQ 5.2+**

SQ support: Thanks for the project, we've been able to investigate and could understand your issue. As we're computing the complexity per class in a different way, the value can now be different but this come in fact from another issue : the plugin doesn't compute the measure 'complexity_in_classs' so there's a fallback mechanism to use the complexity instead. 

But this is not the right way to do : we should only rely on the complexity_in_classes and not anymore rely on the complexity. This will be done by https://jira.sonarsource.com/browse/SONAR-7169.

* link https://groups.google.com/forum/#!topic/sonarqube/8hV4EUauzJE
* related to #721

guwirth:
close #747

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,8 @@ env:
   
 matrix:
     fast_finished: true
-    allow_failures:
-        - env: SONARHOME=/tmp/sonarqube-5.4
+#    allow_failures:
+#        - env: SONARHOME=/tmp/sonarqube-5.4
 
 before_install:
         - cat /etc/hosts # optionally check the content *before*
--- a/integration-tests/features/multimodule_analysis.feature
+++ b/integration-tests/features/multimodule_analysis.feature
@@ -44,7 +44,7 @@ Feature: cpp-multimodule-project
                | complexity               | 7     |
                | function_complexity      | 1.4   |
                | file_complexity          | 0.9   |
-               #| class_complexity         | 7     | -> to enable when 5.x LTS is released, changes in core makes the value from 6 to 7 in 5.x
+               | class_complexity         | 6     |
                # violations
                | violations               | 28    |
                # test execution statistics
--- a/integration-tests/features/smoketest.feature
+++ b/integration-tests/features/smoketest.feature
@@ -39,7 +39,7 @@ Feature: Smoketest
                | complexity               | 7     |
                | function_complexity      | 1.4   |
                | file_complexity          | 0.9   |
-               #| class_complexity         | 7     |  -> to enable when 5.x LTS is released, changes in core makes the value from 6 to 7 in 5.x
+               | class_complexity         | 6     |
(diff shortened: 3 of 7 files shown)


guwirth:
@all someone familiar with class and function metrics and can give some hints? Still not working...

jmecosta:
@guwirth do we really need it?

the c# plugin does not feed this metric to sonar.

https://github.com/SonarSource/sonar-csharp/blob/master/src/main/java/org/sonar/plugins/csharp/CSharpSensor.java

guwirth:
@jmecosta if I understand this right then yes?
https://jira.sonarsource.com/browse/SONAR-7169

jmecosta:
we dont have  a check "class should not be to complex", monitoring a
complexity of a function is imo is enough. as of now i dont think we use
this metric for anything.

would be better to for example just raise issues for a class that has high
complexity. we dont even have class distribution so really dont know if its
taht important

guwirth:
With COMPLEXITY_IN_CLASSES you are maybe right, but still think something is wrong:
Current branch with SQ5.4-RC2: Assertion Failed: function_complexity is actually None
https://travis-ci.org/SonarOpenCommunity/sonar-cxx/jobs/112379604

jmecosta:
ah ok, tought it was only the class that was broken. in this case then i suppose the problem is related and both can be fixed.

      | file_complexity          | 2.0   |
      | class_complexity         | 4     |
      | violations               | 19    |
      | line_coverage            | 100   |
      | branch_coverage          | 50    |
      | test_failures            | 1     |
      | test_errors              | 0     |
      | tests                    | 2     |
      Assertion Failed: 
      	class_complexity is actually 0.0
      	function_complexity is actually None

at least the class_complexity was changed so something done in this pr has made it 0, perhaps need to wait until the official plugins update this?

guwirth:
add metric COMPLEXITY_IN_CLASSES and COMPLEXITY_IN_FUNCTIONS: should work now with all versions

jmecosta:
Good news :) time for a new release?

guwirth:
ready to review and merge

guwirth:
> new release

@jmecosta there are still some open issues: https://github.com/SonarOpenCommunity/sonar-cxx/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22M+0.9.5%22.
There is also still the erroneous indentation check: fix or remove?

jmecosta:
Disable for now? Perhaps later it will be easier to fix it. We use clang
format, so we are not I'm hurry to validate if code is correctly indented

Issue #684, opened 2015-11-20
CxxCompilerVcParser: Default Regex not working


BobSilent:
Default Regex does not work for me

```java
public static final String DEFAULT_REGEX_DEF = "^.*[\\\\,/](.*)\\((\\d+)\\)\\x20:\\x20warning\\x20(C\\d+):(.*)$";
```

I have to adapt it to 
```
sonar.cxx.compiler.regex=^.*[\\\\,/](.*)\\((\\d+)\\)\\x20*:\\x20warning\\x20(C\\d+):(.*)$
```

as there is no space before the ": warning" 
filename.c(276): warning C4334: '<<' : result of 32-bit shift implicitly converted to 64 bits

Bertk:
Please check the source.
https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/compiler/CxxCompilerVcParser.java

Microsoft modified the compiler warning output in VC 9.0, 10.0, 12.0 and therefore the regular expression is configurable.

guwirth:
Think we should document this on https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxPlugin.java?

BobSilent:
I would suggest to at least go on always with latest Conpiler Warnings or with none and force/inform the user to set the value (maybe with some examples in wiki).

Otherwise it is really tricky to find why no findings are found, especially as the example regex on 
https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Supported-configuration-properties
is not so self explained to know what should be matched.

jmecosta:
documented here #806

PR #806, opened 2016-03-13
#684 document vs compiler  reg ex property

jmecosta:
handles #684


jmecosta:
wiki updated also:

https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Compilers

guwirth:
:+1:

PR #810, opened 2016-03-13
Remove methods marked with "override" from public API

### LINKED ISSUE #761: Undocumented API improvements
We were introducing the *Undocumented API* metric in our company. I got the following feedback from the developers:

* Feature does not support **inherited documentations** like doxygen with the switch `INHERIT_DOCS`. There it is possible to document an API only in the interface or base class. Developers don't want to maintain (same) comments at two places.
* Typically it makes less sense to document **default ctor** and **default dtor**. There should be a switch to suppress messages for such methods.
* Typically it makes less sense to document **operators** (=, +, -, ...) the meaning is normally clear. Only if a operator behaves different they should be documented.
* **Getter and Setter**: Should be sufficient to document corresponding member variable. Setter is setting it and getter is returning the current value.

What is your experience with the feature?

lcintrat:
Fix #761


guwirth:
:+1:

PR #813, opened 2016-03-22
Creating a GCC+MAKE log analyser

rlucca:
GCC+Make allows to filter for Entering/Leaving directory to have a better location of files.

### DIFF
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxPlugin.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxPlugin.java
@@ -28,6 +28,7 @@ import org.sonar.api.config.PropertyDefinition;
 import org.sonar.api.resources.Qualifiers;
 import org.sonar.plugins.cxx.compiler.CxxCompilerGccParser;
 import org.sonar.plugins.cxx.compiler.CxxCompilerGccRuleRepository;
+import org.sonar.plugins.cxx.compiler.CxxCompilerGccMakeParser;
 import org.sonar.plugins.cxx.compiler.CxxCompilerSensor;
 import org.sonar.plugins.cxx.compiler.CxxCompilerVcParser;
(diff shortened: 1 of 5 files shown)


sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/compiler/CxxCompilerGccMakeRuleRepository.java
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * {@inheritDoc}
+   */
+  public CxxCompilerGccMakeRuleRepository(ServerFileSystem fileSystem, RulesDefinitionXmlLoader xmlRuleLoader, Settings settings) {
+    super(fileSystem, xmlRuleLoader, settings, KEY, NAME, CUSTOM_RULES_KEY);
+  }
+
+  @Override
+  protected String fileName() {
+    return "/compiler-gcc.xml";
```
jmecosta: you are using the same rules as in gcc parser, cant you use only that rule repository instead of duplicating rules across repositories?

guwirth:
@rlucca thanks for providing this. I'm not familiar with gcc. What's the difference between gcc and gcc-make?

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/compiler/CxxCompilerGccMakeRuleRepository.java
rlucca: Hum... yes, it is possible. I'll change.

rlucca:
@guwirth, the difference between gcc and gcc-make is that the last one can interpret log of GNU Make to track directories. For example, gcc gives a warning in "zipfile.c" file and the directory previously entered was "/home/user". So the issue will be created at "/home/user/zipfile.c".

guwirth:
@rlucca please rebase and upload again. There a bigger changes in the plugin to be compatible with SQ 5.6. I close this one.

rlucca:
OK.

Issue #720, opened 2015-12-28
Add SQALE metadata to XML rule format (SQ5.3+)


guwirth:
Extended rule file format supporting SQALE metadata in same file: add support to cxx plugin

* https://jira.sonarsource.com/browse/SONAR-6591
* https://github.com/SonarSource/sonarqube/blob/master/sonar-plugin-api/src/main/java/org/sonar/api/server/rule/RulesDefinitionXmlLoader.java

Bertk:
Should the SQ-API of the C++ plug-in also be updated ? e.g. use  ```    <sonar.version>5.2</sonar.version> ```
I checked the new format (https://github.com/Bertk/sonar-cxx/commit/a41d22158911565d2cac2f9849e9ee0579347843) and I plan to rework the techn. debt settings but there are some more tasks to complete this effort:
- remove file com/sonar/sqale/cxx-model.xml
- remove <priority> from default-profile.xml
- use RulesDefinitionXmlLoader in class CxxRuleRepository
- remove SqaleXmlLoader from class CxxRuleRepository
We should divide the effort ... (cppCheck, PC-lint, Vera++, RATS, VC-compiler, ...)

guwirth:
@Bertk yes we should do this after this release. Goal with this release is to be still compatible with last LTS 4.5.

PR #816, opened 2016-03-27
support 5.3 rule format

Bertk:
- Extent rules format with remediation information
- Rename tags priority to serverity
- Rename tags configkey to internalkey
- Introduce xml schema with fixed sequence

partial solution for #720


guwirth:
:+1:

PR #822, opened 2016-03-30
cache coverage data

jmecosta:
@guwirth do you know a way to get this cache thing working. caches for each module are always empty

### DIFF
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxPlugin.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxPlugin.java
@@ -52,6 +52,9 @@ import org.sonar.plugins.cxx.tests.dotnet.CxxUnitTestResultsProvider.CxxUnitTest
 import org.sonar.plugins.cxx.utils.CxxMetrics;
 
 import com.google.common.collect.ImmutableList;
+import org.sonar.api.config.Settings;
+import org.sonar.plugins.cxx.coverage.CxxCoverageCache;
+import org.sonar.plugins.dotnet.tests.CoverageAggregator;
 
 /**
  * {@inheritDoc}
@@ -407,6 +410,7 @@ public final class CxxPlugin extends SonarPlugin {
     l.add(CxxRatsSensor.class);
     l.add(CxxXunitSensor.class);
     l.add(CxxCoverageSensor.class);
+    l.add(CxxCoverageAggregator.class);
     l.add(CxxCppCheckRuleRepository.class);
     l.add(CxxCppCheckSensor.class);
     l.add(CxxPCLintRuleRepository.class);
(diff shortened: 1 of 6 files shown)


guwirth:
> do you know a way to get this cache thing working. caches for each module are always empty 

Hi @jmecosta  no idea. What exactly should this PR do/improve?

jmecosta:
in multiple modules configurations, for example when using the msbuild sonar runner if you have a solution with multiple projects then the parser will parse every report again and again, we have 100 MB reports and solutions with 60-70 projects. so the benifict in this case will be good. ive follow the same approach as here, https://github.com/SonarSource/sonar-dotnet-tests-library/pull/19

guwirth:
Gotcha. Is it ready to merge from your side?

jmecosta:
But anyway leave it open I try to figure out how it's donw when I have a
bit of time

guwirth:
> you know a way to get this cache thing working

I guess it's a path issue. Different baseDir or slash/backslash ...

jmecosta:
@BobSilent notice that you have proposed a similar thing to the .net side. have you tested the solution implemented there? i have tested it and i get some behaviour has this one, the batchextension gets instantiated for each module, rather than only once for all modules.

jmecosta:
nevermind, its working now. @guwirth ready for review. next to have cache are reports and build log parser

guwirth:
@jmecosta code looks good but also failing with SQ 5.5-RC1. Start of SQ not possible.

guwirth:
@jmecosta SQ 5.5-RC1 is another problem. 
:+1:

guwirth:
@jmecosta merge conflicts. Please rebase and try again.

jmecosta:
@guwirth rebase and squashed. ready to go

jmecosta:
@guwirth whats the issue with 5.5, i try with a local installation and it also fails with this:

2016.04.09 15:44:18 ERROR web[o.a.c.c.C.[.[.[/]] Exception sending context initialized event to listener instance of class org.sonar.server.platform.PlatformServletContextListener
org.sonar.api.utils.MessageException: The rule 'rats:drand48' is invalid : Linear functions must only have a non empty gap multiplier 
2016.04.09 15:44:18 ERROR web[o.a.c.c.StandardContext] One or more listeners failed to start. Full details will be found in the appropriate container log file
2016.04.09 15:44:18 ERROR web[o.a.c.c.StandardContext] Context [] startup failed due to previous errors

BobSilent:
@jmecosta i proposed a very simple solution there by caching the information for which coverage file the information was already loaded. With caching the information in a static way.
Now the solution is to merge the coverage data into a large coverage information.

If you would have asked me, i would not have reimplemented it here in cxx, as you already have the dotnet library dependency I would have reused the coverage importer from there, by just adding an adapted CxxSensor and CxxParser entrypoint here and using maybe have a dedicated implemention for the bullseye and cobertura parsers here in cxx.
This would also enable cxx for ncover, dotcover and opencover as well (which is implemented in sonar-dotnet-tests-library already)

jmecosta:
@BobSilent thanks, i am ensure how much should we depend on the dot test test library since its mostly for manage side. anyway the implementation that ive done its simpler than in the tests library, so it just saves each report coverage(it does not merge). also we need to do some caching on the issues report and buildlog parsing so it was a good exercise to see how to put it together.

PR #824, opened 2016-04-03
use latest sonar runner 2.5.1

guwirth:
http://docs.sonarqube.org/plugins/servlet/mobile#content/view/6954796

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,8 @@ install:
         - unzip -qq sonarqube-5.3.zip
         - travis_retry wget -q https://sonarsource.bintray.com/Distribution/sonarqube/sonarqube-5.4.zip
         - unzip -qq sonarqube-5.4.zip		
-        - travis_retry wget -q http://repo1.maven.org/maven2/org/codehaus/sonar/runner/sonar-runner-dist/2.4/sonar-runner-dist-2.4.zip
-        - unzip -qq sonar-runner-dist-2.4.zip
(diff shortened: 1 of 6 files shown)


guwirth:
@all: seems to work except reading sonar-runner log. Does someone know what's different with sonar-runner 2.5.1?

```
ut the analysis log contains a line matching
      """
      .*ERROR - Report .* cannot be parsed
      """
      Traceback (most recent call last):
        File "c:\python27\lib\site-packagesehave\model.py", line 1456, in run
          match.run(runner.context)
        File "c:\python27\lib\site-packagesehave\model.py", line 1903, in run
          self.func(context, *args, **kwargs)
        File "integration-tests      eatures\steps	est_execution_statistics.py", line 267, in step_impl
          assert contains_line_matching(context.log, context.text)
      AssertionError
```

jmecosta:
have they change encoding on the files?

jmecosta:
na, some reports are parsed. the google test fails as expected bu with different warnigs supose the tests can be updated also?

guwirth:
**use latest sonar runner 2.5.1**
* warning / errors are different now:
  - old: ERROR - The property 'sonar.tests' is unset. Please set it to proceed
  - new ERROR: The property 'sonar.tests' is unset. Please set it to proceed
* some WARN are INFO only now

@jmecosta are these INFO ok?

jmecosta:
i think so. ERROR: is not us that sets it.

PR #826, opened 2016-04-04
CxxCompilerSensor: Fix check for valid input handling

BobSilent:
By default the CxxReportSensor saveUniqueViolation(..) handles 
correctly issues without file and line information as Project Level Findings

unfortunately IsValidInput is also checked here in CxxCompilerSensor and 
issues without file and line information get ignored


guwirth:
@BobSilent thanks for this.

To merge it you have to rebase it. Would be also good to have a sample/ test to avoid regression.

Bertk:
Nice catch. Can you please also add a compiler message without file name in the unit tests. I am not sure whether any of these messages exists.
Why is the fxcop sqale model included ( ```sonar-cxx-plugin/src/main/resources/com/sonar/sqale/fxcop.xml```) ???
This file should be handled with the C# plugin and not with the C++ Community plug-in.

BobSilent:
fxcop.xml was added by accident.

I added a testcase as requested. i also added the code header, but the checks fail. Am I missing something?

guwirth:
@BobSilent Please rebase to latest master and copy one of the headers from there

guwirth:
PR :+1: 
@Bertk what do you think?

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/compiler/CxxCompilerSensor.java
```diff
@@ -128,20 +128,18 @@ protected void processReport(final Project project, final SensorContext context,
...
       }
     } catch (java.io.FileNotFoundException|java.lang.IllegalArgumentException e) {
       CxxUtils.LOG.error("processReport Exception: {} - not processed '{}'", report, e);
     }
   }
 
-  private boolean isInputValid(String file, String line, String id, String msg) {
-    return !StringUtils.isEmpty(file) && !StringUtils.isEmpty(line)
-      && !StringUtils.isEmpty(id) && !StringUtils.isEmpty(msg);
+  private boolean isInputValid(CompilerParser.Warning warning) {
```
Bertk: ```isInputValid()``` shall block ```null``` and empty strings. I recommend to check at least id and msg. This 2 strings are also used in line 134 ```CxxUtils.LOG.warn("C-Compiler warning: '{}''{}'", w.id, w.msg);``` and a NullPointerException() will occur if ```msg==null```

BobSilent:
@Bertk have you testes the nullref?
because if would have been thrown also today here:
https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/compiler/CxxCompilerVcParser.java

Line 90 to Line 95

I added a null check in Warning Class

Bertk:
@BobSilent Sorry but the update of the comment didn't make it and I just noticed this now. 
I am not sure whether the NullPointerException might occur if the correct number of groups for the regular expression is used. 
:+1:

guwirth:
Ok ready to merge :+1:

PR #834, opened 2016-04-11
Visual studio coverage fix null ref exception

BobSilent:
fix nullref exception for VisualStudio Coverage, where a complete file is not covered by tests:
CxxCoverageSensor
entry.getValue().createMeasures()
entry.getValue() would be be null.

VisualStudioParser:
therfore ignore files which are not in coverageData.

CoverageMeasuresBuilder builder = coverageData.remove(id);
would return null if id does not exists in coverageData

### DIFF
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
@@ -156,7 +156,9 @@ public class CxxCoverageSensor extends CxxReportSensor {
       String filePath = entry.getKey();
       InputFile cxxFile = fs.inputFile(fs.predicates().hasPath(filePath));
       if (cxxFile != null) {
-        Collection<Measure> measures = entry.getValue().createMeasures();
+        CxxUtils.LOG.debug("Saving for file '{}'", filePath);
+        CoverageMeasuresBuilder builder = entry.getValue();
+        Collection<Measure> measures = builder.createMeasures();
         CxxUtils.LOG.debug("Saving '{}' coverage measures for file '{}'", measures.size(), filePath);
         for (Measure measure : measures) {
           Measure convertedMeasure = measure;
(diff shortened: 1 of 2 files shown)


sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
```diff
@@ -165,7 +165,9 @@ private void saveMeasures(SensorContext context,
       String filePath = entry.getKey();
       InputFile cxxFile = fs.inputFile(fs.predicates().hasPath(filePath));
       if (cxxFile != null) {
-        Collection<Measure> measures = entry.getValue().createMeasures();
+        CxxUtils.LOG.debug("Saving for file '{}'", filePath);
```
jmecosta: does it need this additional debug information? there is another 2 lines below

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
```diff
@@ -165,7 +165,9 @@ private void saveMeasures(SensorContext context,
...
+        CoverageMeasuresBuilder builder = entry.getValue();
```
jmecosta: why do we need to introduce this here?

BobSilent: I had the case where an exception was thrown 
entry.getValue().createMeasures();
and it is easier to analyze if there is no chained method call

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
BobSilent: it is in connection with the change below, an exception was thrown and i did not had any file information, therefore i added this line to debug log

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
jmecosta: Sure but can't we print it in the exception handler. Now for nominal cases logs are going to be really big. We have done some changes to the plugin to keeps things to a minimal and this seems going the other team around

guwirth:
@BobSilent please rebase and upload again. There a bigger changes in the plugin to be compatible with SQ 5.6. I close this one.

PR #833, opened 2016-04-10
change base to gap

jmecosta:
should handle startup issues in 5.5

PR #837, opened 2016-04-16
Implements Dr Memory analysis

arnaudsylvestre:
Dr. Memory is a memory monitoring tool capable of identifying memory-related programming errors such as accesses of uninitialized memory, accesses to unaddressable memory.

Here is a Pull Request to support this tool in sonar-cxx.

Hope it helps !


arnaudsylvestre:
This is for the wiki:
### Dr. Memory
SonarQube can be fed with results of a Dr. Memory analysis.

Just launch your program with Dr. Memory. The actual call should look something like:

```BASH
drmemory.exe -logdir c:/logs -- c:/path/to/my/app
```
Now, you just have to fill the sonar.cxx.drmemory.reportPath:

```BASH
sonar.cxx.drmemory.reportPath=c:/logs/**/results.txt
```

jmecosta:
this seems good, however just out of curiosity why did you not use the external plugin. for example intel inspector we use it in same way. see https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Extending-the-code-analysis#resources

arnaudsylvestre:
The resultst.txt file generated by Dr. Memory is not easy to transform into a xml file. I must developed a tool to parse this file (based on RegEx). So, I would like to share it with the community.

In addition, it's easier to use for other people if it is natively integrated into SonarQube (through sonar-cxx)

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/drmemory/CxxDrMemorySensor.java
```diff
@@ -0,0 +1,112 @@
...
+						saveUniqueViolation(project, context, CxxDrMemoryRuleRepository.KEY,
+								errorLocation.file,	errorLocation.line.toString(),
+								error.type.getId(), error.message);
+						break;
+					}
+
+				}
+			}
+		} catch (IOException e1) {
+			throw new EmptyReportException();
```
guwirth: Error handling is different. All IO failures are handled as EmptyReportException. Other sensor are doing this only if really empty.

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/drmemory/package-info.java
```diff
@@ -0,0 +1,27 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+/**
+ * Package with sensor to evaluate Dr Memrory specific results files.
+ */
+@ParametersAreNonnullByDefault
+package org.sonar.plugins.cxx.drmemory;
+
+import javax.annotation.ParametersAreNonnullByDefault;
```
guwirth: no newline at end of file

sonar-cxx-plugin/src/main/resources/drmemory.xml
```diff
@@ -0,0 +1,177 @@
...
+<h2>References</h2>
+<p><a href="http://drmemory.org/docs/page_unaddr.html" target="_blank">Unaddressable Access</a>
+</p>
+     ]]>
+    </description>
+    <internalKey>UnadressableAccess</internalKey>
+    <severity>CRITICAL</severity>
+    <type>BUG</type>
+    <remediationFunction>LINEAR</remediationFunction>
+    <remediationFunctionBaseEffort>15min</remediationFunctionBaseEffort>
```
jmecosta: This needs to be gap, see my pr that fixes those

guwirth:
@arnaudsylvestre thanks for providing this. Looks good beside some smaller points.
* error handling (see comment above)
* wiki item:
  * would be good to have a link to Dr. Memory homepage
  * would be good to add a small sample of the input file format

otherwise :+1:

jmecosta:
By the way how do you handle the stacktrace, single issue or multiple issues?

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/drmemory/CxxDrMemorySensor.java
arnaudsylvestre: What kind of Exception do you want to be raised ? Can I create my own Exception type ?

guwirth: @arnaudsylvestre please have a look to https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxReportSensor.java. EmptyReportException  is a special case.

```Java
  public void analyse(Project project, SensorContext context) {
    try {
      List<File> reports = getReports(settings, fs.baseDir(), reportPathKey());
      violationsCount = 0;
      
      for (File report : reports) {
        int prevViolationsCount = violationsCount;
        CxxUtils.LOG.info("Processing report '{}'", report);
        try {
          processReport(project, context, report);
          CxxUtils.LOG.debug("{} processed = {}", metric == null ? "Issues" : metric.getName(),
             violationsCount - prevViolationsCount);
        } catch (EmptyReportException e) {
          CxxUtils.LOG.warn("The report '{}' seems to be empty, ignoring.", report);
        }
      }

      CxxUtils.LOG.info("{} processed = {}", metric == null ? "Issues" : metric.getName(),
        violationsCount);
          
      if (metric != null) {
        Measure measure = new Measure(metric);
        measure.setIntValue(violationsCount);
        context.saveMeasure(measure);
      }
    } catch (Exception e) {
      String msg = new StringBuilder()
        .append("Cannot feed the data into sonar, details: '")
        .append(e)
        .append("'")
        .toString();
      throw new IllegalStateException(msg, e);
    }
  }
```

arnaudsylvestre:
I parse the stack trace until I found a file that is handled by the analysis. After that, I produce only one issue.

jmecosta:
So you can have multiple issues that are related to the same leak?

arnaudsylvestre:
No, because when I produce an issue, I go to the next leak without analyzing the rest of the stacktrace

jmecosta:
So that means if the leak happens in a file included in the analysis but
after the first is found then you loose that information?

jmecosta:
I suppose that the lowest you can raise the issue the best is to report the
issue? And suppose if you report the all stack in the issue message that is
also good. At least that's the approach I follow for the Intel inspector

arnaudsylvestre:
Dr. Memory generates an entry by leak. The stacktrace is just here to inform in which context the leak happened. So, if there is another leak in the stacktrace, it will be another entry ! So, there is no problem.

In a first version, I displayed the full stack trace in the message but it was not easily readable...

jmecosta:
Ok, i understand what you mean, in previous versions of sq it was easy to
display those. What we found usefull was also to to include the binary that
cause the leak. So that devs would know what test to run to check the leak

sonar-cxx-plugin/src/main/resources/drmemory.xml
jmecosta: see https://github.com/SonarOpenCommunity/sonar-cxx/pull/833/files#diff-0bb24978ba30c6dafd8ed29ceabdd8a2L34 remediationFunctionGapMultiplier

jmecosta:
@arnaudsylvestre but other than the remediationFunctionGapMultiplier i am ok with the pr

arnaudsylvestre:
Thank you for your comments and thank you for this great plugin ! If there is anything else, please tell me !

arnaudsylvestre:
Do you want that I add the documentation in the Wiki ? If yes, where ? Maybe in the **Code checkers** part ?

jmecosta:
yep seems a good place to put it

Issue #828, opened 2016-04-08
Metric lines is an internal metric computed by SonarQube. Provided value is ignored.


guwirth:
LOG file item:
```
DEBUG: Metric lines is an internal metric computed by SonarQube. Provided value is ignored.
```
Remove lines metric calculation from plugin it's part of core now.

guwirth:
@all does someone know how to fix this?

guwirth:
Solved with #913

PR #839, opened 2016-04-17
remove CxxMetric.LINES

guwirth:
This is only a test. Unit tests are failing. CoreMetrics.LINES returns wrong value now. Seems that number is number of code lines now (without comment and empty lines)?


jmecosta:
How can that be, without core knowing anything about how parser works or how comments are.

guwirth:
Good question. 😄. Do you know another plugin doing this right without creating the message in the LOG file?

jmecosta:
But why should we even check lines of code, if that's handle in core they should check all that. If we try to set expectations on that tests will brake every time they change the core implementation

guwirth:
Initial idea was to remove #828 by removing CxxMetric.LINES from our code. I also thought that the core is calculating this then. But as you see the unit test are failing afterwards. Means there seems to be something wrong but I don't know what?

jmecosta:
I would let the test run until the integration tests just to confirm this
suspicions.

guwirth:
will be fixed with #913

PR #846, opened 2016-04-26
log only errors for parameter matching

Bertk:
- MismatchException is used to consume all separators and this is not an
error
- only create an error message if the closing bracket token is not
following an comma token (might not happen)

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
@@ -102,16 +102,19 @@ public class CxxPreprocessor extends Preprocessor {
   }
 
   static class MismatchException extends Exception {
+    private static final long serialVersionUID = 1960113363232807009L;
 
-    private final String why;
-
-    MismatchException(String why) {
-      this.why = why;
+    MismatchException(String message) {
+      super(message);
     }
-
-    @Override
-    public String toString() {
-      return why;
+    MismatchException(Throwable cause) {
+      super(cause);
+    }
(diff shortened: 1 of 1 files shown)


cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
```diff
@@ -743,19 +743,20 @@ private int matchArguments(List<Token> tokens, List<Token> arguments) {
           break;
         }
       } while (true);
-
+    } catch (MismatchException me) {
+      LOG.debug("Mismatch: expected ','got: '" + rest.get(0).getValue() + "'");
```
jmecosta: should this then be error?

Bertk: Typically this message is: 

> Mismatch: expected ',' got: ')' 

This is the reason for the exception but not an error. Anything different than ',' or ')' is an error. This message could be an error if the ')' case is not reported.

SonarOpenCommunityAdmin: ok, but then do we even need to do the debug print. if running with -X it will just print loads of not needed stuff. right?

jmecosta:
ok, but then do we even need to do the debug print. if running with -X it will just print loads of not needed stuff. right?

Bertk:
Yes, the debug information might be huge. I removed this code line. Please check the update.

guwirth:
Code review only without testing with sample :+1:

jmecosta:
Try still to run this before merging

jmecosta:
looks good: :+1:

PR #841, opened 2016-04-21
sonar-api-5.6-RC2

guwirth:
- org.codehaus.sonar => org.sonarsource.sonarqube
- use API version 5.6-RC2
- org.sonarsource.dotnet:sonar-dotnet-tests-library:1.3.2
  - support: sonar.cxx.xunit.reportsPaths

PR #850, opened 2016-04-30
use @Properties for SonarPlugin

guwirth:


guwirth:
This PR is to prepare #841 
* subcategory is no more available: don't know how to handle this?
* property settings **project = true, module = true** are not unique: seems sometimes random if set or not?

guwirth:
https://groups.google.com/forum/#!topic/sonarqube/9fJQW3eum-Y
```
Hi Günter,

You should keep on using the PropertyDefinition API, because it's more easy to
maintain an java API than an annotation API. But if you really want to use annotation,
you can create your own one based on the java API.

Regards,
Julien LANCELOT | SonarSource
http://sonarsource.com
```

PR #861, opened 2016-05-08
remove StringUtils

guwirth:
remove all StringUtils functions
disable check for Java 7 (temporary only)


guwirth:
this one close #849

PR #867, opened 2016-05-20
revert java.nio.file.Files.readAllLines usage

guwirth:
`java.nio.file.Files.readAllLines` behaves different as `com.google.common.io.Files.readLines`, sample:
* read `/* ü */` with US-ASCII
* `com.google.common.io.Files.readLines` returns `/* ? */`, no exception
* `java.nio.file.Files.readAllLines`: exception & list = null

Behavior: replacing invalid character and will allow to continue the check.


jmecosta:
@guwirth please merge and rebase the remove the getchild so i can give another test there and confirm this is ok also

PR #874, opened 2016-05-28
remove DependencyAnalyzer

### LINKED ISSUE #509: Make "Dependency Structure Matrix" work for multi-module projects

guwirth:
* metrics are deprecated or calculated by core
* close #509


guwirth:
@all think this ist no more needed all metrics are marked deprecated or calculated by the core?
@Typz do you also agree?

jmecosta:
We were using the duplicate header check. It was useful.. Seems to me it can exist regardless of the dependency check

guwirth:
> We were using the duplicate header check.

Think would be better to use an static code analyser for that? If we wanna keep this rule we have to move the check to e.g. the preprocessor.

How is the rule working today? Check for two direct `#include` in the same CPP file?

jmecosta:
but if we have the rule already? 

yes it detects when the same file is included twice in same file

guwirth:
> but if we have the rule already? 

If we are removing the DependencyAnalyzer class then also this check is no more working (see comment in https://github.com/SonarOpenCommunity/sonar-cxx/blob/4db72204a105b81b2897439a7cfb6b47ad5ca481/cxx-checks/src/main/java/org/sonar/cxx/checks/DuplicatedIncludeCheck.java.

Keeping the whole thing for this rule only is too expensive: https://github.com/SonarOpenCommunity/sonar-cxx/pull/874/files#diff-9320b8c32912c240d697b4309c33166eL102.

For this rule we have to find another way then.

jmecosta:
ok, but the dependency check is also usefull on its own, right? I mean we dont need to have fancy core diagram with dependencies however if we are able to detect those is already a very good improvement. 

why are we removing this anyway? core drop this concept and does not show anymore the diagram. Thats fine, however cycle concent is a general thing. and if the plugin can detect it and even better improve it accross modules, that would make it a extremely valuable feature for the plugin.

Typz:
As far as I know, the "raw" dependencies reporting is not deprecated, only all the metrics that we compute from them : so it is best to keep the dependency analyzer, both to compute these (which may be used again by core of/when they find a nice way to present the info, eg a dependency matrix v2) and generate issues (cyclic dependencies, ...)

guwirth:
My intention was to delete it because most parts inside `DependencyAnalyzer` are deprecated. On the main branch `addFile` is commented out to make it running: https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/squid/CxxSquidSensor.java#L191. If the remaining parts are functional and deprecated parts are removed I also agree it's better to keep it.

@Typz Do you have some time to clean it up?

guwirth:
will be fixed with #913

PR #894, opened 2016-06-23
support Visual C++ throw(...)

### LINKED ISSUE #892: parser fails with "throw(...)" exception specification
The parser fails with exception specification like 
```cpp
int myfunc() throw(...){}
```
Without "..." everything is fine

```cpp
int myfunc1() throw(){
}
int myfunc2() throw(int){
}
```
see also https://msdn.microsoft.com/en-us/library/wfa0edys.aspx

Bertk:
close #892


guwirth:
Looks good. Think doesn't matter at the moment where we add the "...".
typeIdList is only used by dynamicExceptionSpecification.

I wait with merging until version number is increased.

PR #902, opened 2016-07-02
exclude sonar api  from sslr

jmecosta:
#898


jmecosta:
merged to #903 . open just to investigate #898

PR #917, opened 2016-07-29
Catch PC-Lint MISRA C++ messages

dkt01:
Added check for default PC-Lint MISRA C++ 2008 message identifier.  Rules already exist.


guwirth:
Hi @dkt01 , thanks a lot for your contribution. Would like to ask you to add a testcase in https://github.com/dkt01/sonar-cxx/blob/master/sonar-cxx-plugin/src/test/java/org/sonar/plugins/cxx/pclint/CxxPCLintSensorTest.java to avoid regression. Please also squash your commit for final merge then.  Regards

PR #930, opened 2016-08-17
Fix issue #928

### LINKED ISSUE #928: Problem with alias declarations inside template class
Hi, 

It's me again and my template issues ;)

I think there is an issue with 7a5e47a8eebb9715205e6041346c0b87b8eb51e2

The problem occurs when there is an alias declaration (non-template) inside a class template. For instance: 

```cpp
template <std::size_t T_T = 30>
struct binarize_layer_desc {
    /*!
     * Blah
     */
    using parameters = cpp::type_list<>;
};
```

In that case, "parameters" is shown as undocumented. If I comment the class and the not the member:
    
```cpp
/*!
 * Blah
*/
template <std::size_t T_T = 30>
struct binarize_layer_desc {
    using parameters = cpp::type_list<>;
};
```

The member "parameters" is not marked as undocumented (no issue generated). 

In my opinion this comes from the commit 7a5e47a8eebb9715205e6041346c0b87b8eb51e2 because it gets the first template ancestor of the alias declaration and it gets the template declaration of the class itself and then looks for the documentation of the class and not of the alias declaration. It should check if the template is related to the class or to the alias declaration. I can be mistaken though since I don't really know the AST functions of sonar-cxx, but it seems logical. 

Thanks guys :)

P.S. I'm using version 0.9.5.

wichtounet:
This PR fixes #928 

The fix was simpler than I thought, it was already done for member declarations. I just used the same for alias declarations. This avoid that the wrong template declaration is used.

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/visitors/AbstractCxxPublicApiVisitor.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/visitors/AbstractCxxPublicApiVisitor.java
@@ -511,11 +511,19 @@ public abstract class AbstractCxxPublicApiVisitor<GRAMMAR extends Grammar>
       if (aliasDeclIdNode == null) {
         LOG.error("No identifier found at {}", aliasDeclNode.getTokenLine());
       } else {
-        // check if this is a template specification to adjust
+        // Check if this is a template specification to adjust
         // documentation node
-        AstNode template = aliasDeclNode
-            .getFirstAncestor(CxxGrammarImpl.templateDeclaration);
-        AstNode docNode = (template != null) ? template : aliasDeclNode;
+        AstNode container = aliasDeclNode.getFirstAncestor(
+          CxxGrammarImpl.templateDeclaration,
+          CxxGrammarImpl.classSpecifier);
+
+        AstNode docNode;
+        if (container == null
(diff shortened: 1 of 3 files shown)


wichtounet:
I don't think the travis-ci build failure is of my doing: "Cannot start SonarQube from '/tmp/sonarqube-5.6', exiting"

jmecosta:
yep, looks like it. in windows everything is fine. can you just push a dummy commit to force travis to rebuild

wichtounet:
It does not change anything. Seems your Travis is broken.

guwirth:
@wichtounet I tried to rebuild a working commit with same result: it's failing now. Will ask Travis support what's wrong.

guwirth:
@wichtounet please rebase & squash to merge this (see #932)

wichtounet:
Not better

guwirth:
After there is a Travis problems which seems to be not easy to fix (on Travis side) I would also merge it with only green AppVeyor? Someone else wanna review the changes?

guwirth:
@wichtounet CI is up and running again, please rebase and try again.

wichtounet:
Done :)

Congrats on fixing the CI issues.

guwirth:
@wichtounet thx for the fix.

PR #943, opened 2016-09-10
Update highlighter.cc

jmecosta:


guwirth:
@jmecosta thx will try this evening.

PR #949, opened 2016-09-15
Alias declaration in function is not public API

wichtounet:
This commit prevents alias declaration declared at block-scope to be
identified as public API. It also adds a test to ensure that it
works.

### DIFF
--- a/cxx-squid/src/test/resources/metrics/alias_in_template_func.h
+++ b/cxx-squid/src/test/resources/metrics/alias_in_template_func.h
@@ -1,0 +1,25 @@
+template<typename Second>
+struct A {
+    template<typename Third>
+    void foo(){
+        using type = void;
+    }
+};
+
+template<typename Fourth>
+void bar(){
+    using type = void;
+}
+
+/*!
+ * \brief Commented
+ */
+template<typename Fourth>
+void foobar(){
+    /*!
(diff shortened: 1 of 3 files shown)


wichtounet:
It seems that this time it's AppVeyor that is broken: 

> 'mvn' is not recognized as an internal or external command

cxx-squid/src/test/resources/metrics/alias_in_template_func.h
```diff
@@ -0,0 +1,12 @@
...
+    template<typename Third>
+    void youpee(){
+        using intrinsic_type = void;
+    }
+};
+
+template<typename Fourth>
+void youpee(){
+    using intrinsic_type = void;
+}
```
guwirth: @wichtounet would add also a positive test: same with comments.

guwirth:
@wichtounet thx looks good. @jmecosta maybe you can have a look to the AppVeyor issue. 
:+1:

cxx-squid/src/test/resources/metrics/alias_in_template_func.h
wichtounet: @guwirth Good idea

wichtounet:
I improved the test and pushed a new squashed commit.

wichtounet:
The build should be fixed with #950

PR #950, opened 2016-09-16
Improve AppVeyor MAVEN path handling

wichtounet:
This should fix the appveyor build by handling more cases. 

This should handle more cases:

1) If maven is not installed at all
2) If maven is in path but not configured (M2_HOME/MAVEN_HOME unset)
3) If Maven is in path and configured (M2_HOME/MAVEN_HOME set)


jmecosta:
Yep, this was needed, thanks for the fix. You can merge this

Issue #842, opened 2016-04-21
Coverage fails with wrong number of lines


Jonathan34:
Hi,

Following up on https://github.com/SonarOpenCommunity/sonar-cxx/issues/840, i have changed my configuration to debug to see how optimization could affect the coverage.

I have deleted and recreated the project in SonarQube.
During the POST step in SonarQube, i have an exception:
`
2016-04-21T18:40:32.1475520Z 13:40:32.085 DEBUG - Saving '7' coverage measures for file 'C:\TFS\Builds\Agent404_01\1\s\Acidizing\View\AcidizingCoiledTubingView.cpp'
2016-04-21T18:40:32.1475520Z 13:40:32.085 ERROR - Saving cov measure 'coverage_line_hits_data' = 'null' failed
2016-04-21T18:40:32.1475520Z 13:40:32.085 ERROR - Obtained 'coverage_line_hits_data' = '8=0;9=0;10=0;11=0;14=0;15=0;16=0;31=0' failed
2016-04-21T18:40:32.1475520Z 13:40:32.085 ERROR - Ctype : 'UT_COVERAGE'     Exception 'Can't create measure for line 31 for file 'C:/TFS/Builds/Agent404_01/1/s/aaa/View/MyClass.cpp' with 17 lines'
2016-04-21T18:40:32.1475520Z 13:40:32.100 INFO  - Analysis report generated in C:\TFS\Builds\Agent404_01\1\.sonarqube\out\.sonar\batch-report

...

2016-04-21T18:40:33.0694300Z ##[error]ERROR: Error during Sonar runner execution
2016-04-21T18:40:33.0694300Z ##[error]org.sonar.runner.impl.RunnerException: Unable to execute Sonar
2016-04-21T18:40:33.0694300Z ##[error]at org.sonar.runner.impl.BatchLauncher$1.delegateExecution(BatchLauncher.java:91)
2016-04-21T18:40:33.0694300Z ##[error]at org.sonar.runner.impl.BatchLauncher$1.run(BatchLauncher.java:75)
2016-04-21T18:40:33.0694300Z ##[error]at java.security.AccessController.doPrivileged(Native Method)
2016-04-21T18:40:33.0694300Z ##[error]at org.sonar.runner.impl.BatchLauncher.doExecute(BatchLauncher.java:69)
2016-04-21T18:40:33.0694300Z ##[error]at org.sonar.runner.impl.BatchLauncher.execute(BatchLauncher.java:50)
[...]

Issue #758, opened 2016-01-16
Vote: default behavior invalid issues and messages


guwirth:
To speed up analysis the different reports could be generated on different workstation. One is doing the static code analysis, another one the unit tests and a third one the coverage measurement and so one. To forward the results there are two possibilities then:
* Wait unit all results for a revision are available and forward them all together to SonarQube. This is the safe way and everything fits together. In case there is an issue or measure where no sources are available this is for sure an error. Drawback of this method is that you have to wait always until the last report is available.
* Second possibility is to run the different analysis asynchronous, collect them all in a folder and use always the latest results only. Every time if there is at least one new report available a new SonarQube update can start. This method provides typically much faster results but with the disadvantage that there could be temporary invalid measures and issues because machines are working on different revision numbers. In this case it must be possible to ignore invalid issues and measures and continue. (See also #757)
* Third possibility is to mix this two approaches: do a complete new analysis once a week (e.g. over the weekend) and only incremental analysis during the week. For the switch from one mode to the other it must be ensured that all reports has been generated at least once otherwise you could run into problems like loosing false positives.

To provide a generic solution the plugin should support configuration settings like:
* ignore invalid issues and metrics and continue
* break if no report is available

I'm not sure if these are global settings or each sensor should have this? Solution of PR #757 is a partial solution doing this for code coverage.

What is your experience with this? Are there any other use cases?
Looking forward to get your feedback.

guwirth:
How should the plugin behave, please vote (see also discussion in #757):

(1) invalid issue in a report (e.g. cppcheck, external, ...)
 a) create an error and stop analysis
 b) create an error and continue analysis
 c) create a warning and continue analysis

(2) invalid measure in a report (e.g. unit tests, coverage, ...)
 a) create an error and stop analysis
 b) create an error and continue analysis
 c) create a warning and continue analysis

(3) no report available (configuration file contains a property but no file found)
 a) create an error and stop analysis
 b) create an error and continue analysis
 c) create a warning and continue analysis

jmecosta:
We are voting for default behaviour.

1. a
2. a
3. a if report is not one of the *.xml

Bertk:
My vote is:

1. b (there are also tooling issues and stopping the analysis will might mike a tool useless. I will not modify the source to adapt it for any tool)
[...]

PR #959, opened 2016-10-01
ignore invalid coverage measures

jmecosta:
@guwirth @Bertk  lets ignore the coverage measures?

### DIFF
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
@@ -37,6 +37,7 @@ import org.sonar.api.config.Settings;
 import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
 import org.sonar.plugins.cxx.CxxLanguage;
+import org.sonar.plugins.cxx.CxxPlugin;
 import org.sonar.plugins.cxx.utils.CxxReportSensor;
 
 /**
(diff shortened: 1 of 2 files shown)


guwirth:
@jmecosta looks good to me, closes #842.
Later on I would add a config setting to decide continue/break (#758)

jmecosta:
@guwirth i will create the flag now, and make it disable by default.

jmecosta:
should now cover #842 and #758

guwirth:
@jmecosta I would not introduce a switch to the coverage sensor, I would introduce a general switch, see my last proposal in #758. Doing it as you did it would mean in consequence that we have to (can) add settings to each sensor individual. Not sure if these are not too much choices for a normal user?

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/coverage/CxxCoverageSensor.java
```diff
@@ -163,8 +164,16 @@ private void saveMeasures(SensorContext context,
           }                             
         }
         
-        newCoverage.save();
-        
+        try
+        {
+          newCoverage.save();
+        } catch(Exception ex) {
+          LOG.error("Cannot save measure '{}' : '{}', ignoring measure", filePath, ex.getMessage());
```
guwirth: @jmecosta setting name and flag name do not match? ignoreInvalidMeasures and FAIL_ON_INVALID_MEASURES have invers logic?

jmecosta: yep, forgot to change that. now done

guwirth:
@bertk we need a third opinion ;-). One setting for all sensors to ignore invalid input or individual settings for each sensor?
@all opinions from others are also welcome.

Bertk:
I am fine to ignore invalid input data as long we have error messages for this.

guwirth:
@Bertk question was, use
* `sonar.cxx.coverage.failOnInvalidMeasures` for coverage sensor only
* or one global setting `sonar.cxx.errorRecoveryEnabled` to put all sensor in a tolerant mode

Bertk:
@guwirth 
I missed this detail - sorry :grin:
I think the analysis should not stop if a sensor has an issue e.g. invalid or outdated data. Until now I observed the wrong line number issue with code coverage files only. The code coverage reports might be created on special test systems and not on the build systems which execute the static code analysis.
We should have a global setting for the _tolerant mode_. This is also easier to handle by users and it should be the default.

guwirth:
@jmecosta hi, what should we do with this now?  I still would prefer to have only one setting to put the whole plugin into a tolerant mode.

jmecosta:
Yep, I will make the change soon. As per vote

jmecosta:
should be ready

guwirth:
@jmecosta thx for fixing this.

PR #971, opened 2016-10-10
ignore measures for tests that can be saved because they have been added already

jmecosta:
…ded already


guwirth:
@jmecosta ready to merge?

jmecosta:
Yep

PR #972, opened 2016-10-11
fail if coverage conditions are invalid

jmecosta:
hald made fix, in fact the exceptions are also throw when conditions are set.


guwirth:
@jmecosta ready to merge?

jmecosta:
Yep

PR #988, opened 2016-10-30
fix quality flaws

Bertk:
- different license headers
- method needlessly boxes a boolean constant
- concatenates the result of a toString()
- passes constant String of length 1 to character overridden method 
- string literal on the left side of this string comparison
- Assertions should be complete

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
@@ -1,18 +1,18 @@
 /*
  * Sonar C++ Plugin (Community)
- * Copyright (C) 2011-2016 SonarOpenCommunity
+ * Copyright (C) 2010-2016 SonarOpenCommunity
  * http://github.com/SonarOpenCommunity/sonar-cxx
- * 
+ *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
  * License as published by the Free Software Foundation; either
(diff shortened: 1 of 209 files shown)


guwirth:
@Bertk looks good, thx for providing this. Would be good a second one (maybe @jmecosta) will also have a look.

guwirth:
Will merge this because a lot of files are changed. Risk to get merge conflicts if I delay it. @jmecosta if there are problems we have to fix it later on.

cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java
```diff
@@ -1,493 +1,496 @@
...
-    }
-    AddMacro("_MSC_VER=1900", fileElement);
-    // VS2015 RC
-    AddMacro("_MSC_FULL_VER=190022816", fileElement);
-    //_MFC_VER Defines the MFC version. For example, in Visual Studio 2013, _MFC_VER is defined as 0x0C00.
-    AddMacro("_MFC_VER=0x0C00", fileElement);
-    AddMacro("_ATL_VER=0x0C00", fileElement);
-  }
-}
+/*
```
jmecosta: Seems to be only line endig fixing. Seems ok

jmecosta: Seems to be only line endig fixing. Seems ok

jmecosta: Seems to be only line endig fixing. Seems ok

Issue #616, opened 2015-08-24
Support for TFS team builds (MSBuild .SonarQube.Runner)


Bertk:
@guwirth, @wenns, @jmecosta: Shouldn't we add this to the Wiki?  
The new SQ tool “MSBuild SonarQube Runner” simplifies the TFS Team build integration and I could successfully analyze the CppCheck C++ sources using the C++ Community plug-in. I followed the installation description form Microsoft ALM rangers [SonarQube Setup Guide For .NET Users] (https://github.com/SonarSource/sonar-.net-documentation/releases/download/1.1.0/SonarQube-Setup-Guide-For-Net-Users-V-1-1-0.pdf) and used for my tests the TFS 2015 Express edition. 
My goal was to use MSVC++, PC-lint, CppCheck, RATS and Vera++ reports for a daily build scenario. All tool reports are created and stored in the SQ DB with one build and after a small modification to support the MSVC++ solution log (df042081f34925bc79ddf7b52b5d898d12a084a9) also the compiler warnings showed up. The SQ instance has now a multi-module project based on the Visual Studio solution file :grin:
I used the following C++ VS property file for the MS-Build configuration “SonarQube.StaticAnalysis.props”:

*SonarQube.StaticAnalysis.props was replaced with updated version - see below*

I added this property file to the folder with the solution file and added the reference to the usual import groups settings of the visual studio project:
```
...
  <ImportGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="PropertySheets">
    <Import Project="$(SolutionDir)\SonarQube.StaticAnalysis.props" />
    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" />
  </ImportGroup>
...
```

Final statement -> 
# This works also for C++ 
and the TFS build report shows this:
![capture](https://cloud.githubusercontent.com/assets/2315215/9450492/2907087e-4aa9-11e5-9a45-021785a2389b.PNG)

guwirth:
Sounds also interesting for others and should be documented.

jmecosta:
I have a few ideas on how to improve this, I will comment later as soon as
I find the time to work on this on our projects

Bertk:
[...]

Issue #421, opened 2015-02-06
Needs to support compile_commands.json


LegalizeAdulthood:
sonar-cxx only lets me configure macros and include paths at the project level.  This is an over-simplification of how things work in a C++ code base.

Instead, it should support [`compile_commands.json`](http://clang.llvm.org/docs/JSONCompilationDatabase.html) which allows me to define includes and preprocessor symbols appropriately for each and every file in my build.

LegalizeAdulthood:
Oh, and this would come for free if you were based on clang libtooling.

jmecosta:
How would you produce the compile commands file?

LegalizeAdulthood:
CMake can generate it directly if you have a CMake-based build.  If you had a SCons based build you could generate it from SCons.

There are also approaches of providing a "build system interceptor" so that you run your build system and when it invokes the compiler, it's actually invoking your interceptor which captures the compile command for each translation unit before passing the command onto the real compiler.  Clang's static analyzer uses the latter approach, but only as a way of integrating the static analyzer into real-world builds that can't generate a `compile_commans.json` file.

Typz:
The documentation recommends Bear https://github.com/rizsotto/Bear in case the file is not generated by cmake/scons.

-- 
Francois

jmecosta:
Yes this is cool. We use msbuild, and ive created pull request to support a
similar approach see #399. Ive tough it would be a good idea to include
already something to use only macros and includes that are used in each
file. But that requires a bigger change. But I like the idea

guwirth:
Are these tools especially Bears also running on Windows?

> Am 06.02.2015 um 20:44 schrieb Jorge Costa <notifications@github.com>:
> 
> Yes this is cool. We use msbuild, and ive created pull request to support a
> similar approach see #399. Ive tough it would be a good idea to include
> already something to use only macros and includes that are used in each
> file. But that requires a bigger change. But I like the idea
>

jmecosta:
Should be quite portable in unix operating systems according to readme.
Anyway what it requires from the plugin side is the ability define includes
and macros per file. Once that is done we can have multiple parsers
supporting multiple operating systems

LegalizeAdulthood:
[...]

PR #993, opened 2016-11-08
JSON Compilation Database support

vesajaaskelainen:
This pull request adds support for use compilation unit specification file (xml) to specify defines and includes for compilation units (source files) and for global files which are not compilation units (include files).

Example usage is to use in example Klocwork to statically analyse source code. Use its captured include and defines and from those generate this compilation unit specification file.

With this compiler includes and defines are automatically set correctly for analysis thus giving better analysis results.

### DIFF
--- a/sonar-cxx-plugin/pom.xml
+++ b/sonar-cxx-plugin/pom.xml
@@ -95,6 +95,16 @@
       <version>2.6</version>
       <type>jar</type>
     </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-core</artifactId>
+      <version>2.8.6</version>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>2.8.6</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/cxx-squid/src/main/java/org/sonar/cxx/CxxCompilationUnitSettings.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/CxxCompilationUnitSettings.java
@@ -1,0 +1,44 @@
+/*
+ * Sonar C++ Plugin (Community)
+ * Copyright (C) 2010-2017 SonarOpenCommunity
+ * http://github.com/SonarOpenCommunity/sonar-cxx
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.cxx;
+
+import java.util.List;
+import java.util.Map;
+
+public class CxxCompilationUnitSettings {
+  private Map<String, String> defines = null;
+  private List<String> includes = null;
+
+  public Map<String, String> getDefines() {
+    return defines;
+  }
+
+  public void setDefines(Map<String, String> defines) {
+    this.defines = defines;
+  }
+
+  public List<String> getIncludes() {
+    return includes;
+  }
+
+  public void setIncludes(List<String> includes) {
+    this.includes = includes;
+  }
+}
--- a/cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
@@ -25,6 +25,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 
@@ -48,6 +49,9 @@ public class CxxConfiguration extends SquidConfiguration {
   private boolean errorRecoveryEnabled = true;
   private List<String> cFilesPatterns = new ArrayList<>();
   private boolean missingIncludeWarningsEnabled = true;
+  private String specFile = null;
+  private CxxCompilationUnitSettings globalCompilationUnitSettings = null;
+  private HashMap<String, CxxCompilationUnitSettings> compilationUnitSettings = new HashMap<>();
 
   private final CxxVCppBuildLogParser cxxVCppParser;
 
@@ -214,6 +218,41 @@ public class CxxConfiguration extends SquidConfiguration {
 
   public boolean getMissingIncludeWarningsEnabled() {
     return this.missingIncludeWarningsEnabled;
+  }
+
+  public String getSpecFile() {
+	return specFile;
+  }
+
+  public void setSpecFile(String specFile) {
+    this.specFile = specFile;
+  }
+
+  public CxxCompilationUnitSettings getGlobalCompilationUnitSettings() {
+    return globalCompilationUnitSettings;
+  }
+
+  public void setGlobalCompilationUnitSettings(CxxCompilationUnitSettings globalCompilationUnitSettings) {
+    this.globalCompilationUnitSettings = globalCompilationUnitSettings;
+  }
+
+  public CxxCompilationUnitSettings getCompilationUnitSettings(String filename) {
+    return compilationUnitSettings.get(filename);
+  }
+
+  public void addCompilationUnitSettings(String filename, CxxCompilationUnitSettings settings) {
+    compilationUnitSettings.put(filename, settings);
+  }
+
+  public List<File> getCompilationUnitSourceFiles() {
+    List<File> files = new ArrayList<File>();
+
+    for (Iterator<String> iter = compilationUnitSettings.keySet().iterator(); iter.hasNext(); ) {
+      String item = iter.next();
+      files.add(new File(item));
+    }
+
+    return files;
   }
 
   public void setCompilationPropertiesWithBuildLog(List<File> reports,
--- a/cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
@@ -50,6 +50,7 @@ public class CxxConfiguration extends SquidConfiguration {
   private List<String> cFilesPatterns = new ArrayList<>();
   private boolean missingIncludeWarningsEnabled = true;
   private String jsonCompilationDatabaseFile = null;
+  private boolean scanOnlySpecifiedSources = false;
   private CxxCompilationUnitSettings globalCompilationUnitSettings = null;
   private HashMap<String, CxxCompilationUnitSettings> compilationUnitSettings = new HashMap<>();
 
@@ -226,6 +227,14 @@ public class CxxConfiguration extends SquidConfiguration {
 
   public void setJsonCompilationDatabaseFile(String jsonCompilationDatabaseFile) {
     this.jsonCompilationDatabaseFile = jsonCompilationDatabaseFile;
+  }
+
+  public boolean isScanOnlySpecifiedSources() {
+    return scanOnlySpecifiedSources;
+  }
+
+  public void setScanOnlySpecifiedSources(boolean scanOnlySpecifiedSources) {
+    this.scanOnlySpecifiedSources = scanOnlySpecifiedSources;
   }
 
   public CxxCompilationUnitSettings getGlobalCompilationUnitSettings() {
--- a/cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java
@@ -49,7 +49,7 @@ public class CxxConfiguration extends SquidConfiguration {
   private boolean errorRecoveryEnabled = true;
   private List<String> cFilesPatterns = new ArrayList<>();
   private boolean missingIncludeWarningsEnabled = true;
-  private String specFile = null;
+  private String jsonCompilationDatabaseFile = null;
   private CxxCompilationUnitSettings globalCompilationUnitSettings = null;
   private HashMap<String, CxxCompilationUnitSettings> compilationUnitSettings = new HashMap<>();
 
@@ -220,12 +220,12 @@ public class CxxConfiguration extends SquidConfiguration {
     return this.missingIncludeWarningsEnabled;
   }
 
-  public String getSpecFile() {
-	return specFile;
+  public String getJsonCompilationDatabaseFile() {
+	return jsonCompilationDatabaseFile;
   }
 
-  public void setSpecFile(String specFile) {
-    this.specFile = specFile;
+  public void setJsonCompilationDatabaseFile(String jsonCompilationDatabaseFile) {
+    this.jsonCompilationDatabaseFile = jsonCompilationDatabaseFile;
   }
 
   public CxxCompilationUnitSettings getGlobalCompilationUnitSettings() {
--- a/cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
@@ -26,6 +26,7 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Deque;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
@@ -47,6 +48,7 @@ import com.sonar.sslr.api.TokenType;
 import com.sonar.sslr.api.Trivia;
 import com.sonar.sslr.impl.Parser;
 
+import org.sonar.cxx.CxxCompilationUnitSettings;
 import org.sonar.cxx.CxxConfiguration;
 import org.sonar.cxx.lexer.CxxLexer;
 import org.sonar.squidbridge.SquidAstVisitorContext;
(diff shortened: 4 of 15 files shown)


guwirth:
@vesajaaskelainen thanks a lot for contributing to this plugin. 

I have some questions and hints to your pull request:
* What is different compared to setting the defines and includes in sonar-project.properties?
* xml file with the defines and includes: is this somehow a standard format? if yes from which tool?
* Could you provide an integration test sample in https://github.com/SonarOpenCommunity/sonar-cxx/tree/master/integration-tests/testdata?

vesajaaskelainen:
@guwirth: 

1) Problem of using global defines and includes is that they are not what is used during actual compilation process. When compiling some files there may be different defines or include directories specified. This gets fixed in our case by using Klocwork's build specification file that is captured during build of the software. There seems to be slight differences in defines and includes when compiling some tools or when compiling library part of the software or when compiling test suite.

As an example we have some code bases that gets compiled for multiple architectures. In this case defines and includes vary.

There is actually one problem left here; if I run analysis for project let's say for x86 then different set if defines and includes are in use than for arm target. Different set of files get compiled in. For this I added 'sonar.cxx.scanOnlySpecFileSources' to only scan files that are used in build. However it does not automatically add include files that is referenced in sources so that "full" analysis would be done for those with defines and includes in use when #include was specified. If there would be some mechanism to enable following to includes and adding then scanning them and performing normal analysis would be ideal. But so far didn't figure out good way for this.

Another example is that if one uses FreeRTOS or so which includes support lots of different architectures. Only one of those will be used by software being compiled. So we want to scan what we use and not what we do not. To workaround this we do not scan FreeRTOS folders at all as it gives too much noise.

2) About XML format. This is cooked format just for this case. Output produced in example by Klocwork is custom format which I think is not suitable to be integrated here. If you know some other format that would be suitable here I could adapt code for that format too. XML as format is used because it was actively used in project and I didn't want to add additional dependencies.

3) Could you give a tip how to implement this integration test sample? I was thinking on adding org.sonar.cxx.preprocessor.CxxPreprosessorTest that could perhaps exercise it a bit more? What kind of test are you expecting for?

guwirth:
@vesajaaskelainen think the easiest to understand it is to add a sample to https://github.com/SonarOpenCommunity/sonar-cxx/tree/master/integration-tests/testdata
* own subfolder
* global settings in root
* one subfolder with one source file and local settings 
* another subfolder with two source files and individual local settings per file

Think this helps to understand and test the feature. To integrate it into our tests we can help you then. 

@all someone else an opinion to this feature?

jmecosta:
@vesajaaskelainen this is actually interesting, we should have something like this for the msbuild log parsing. currently we extract all inforamtion and give all information to all files. If i understnad you only get the includes and defines for each file, so during the parsing only the correct includes/defines are used?

perhaps we could refactor this and make a common functionality for every method of extranction information. i see that we will have plenty of duplication if each method has its own implementation?

Bertk:
I appreciate this initiative but I think we really should enhance existing features instead adding proprietary configuration files. The Community C++ plug-in supports already **sonar.cxx.forceIncludes** and **sonar.cxx.defines** which allows to add compiler settings dynamically.
The MSBuild log parsing adds this kind of information for each compilation unit from the build log file and other compilers could be supported as well.
I also use MSBuild targets to feed the Community C++ plug-in with information from the build definition and this is another way to define the analysis scope (https://github.com/SonarOpenCommunity/sonar-cxx/issues/616).  This does not require any extension of the Community C++ plug because it uses existing interfaces e.g. available with **SonarQube Scanner for MSBuild** .

vesajaaskelainen:
@jmecosta Yes that is the idea. Proper defines and includes for source files as they were used during compilation process. Only problem is way sonar-cxx currently analyses header files as they also need include and defines to be available. I suppose right way would be to analyse only compiled source files and then record include/define state on entry with #include and then analyse all combinations of different combinations and give combined analysis report. I tried to make commits so that it could be common functionality and only the parsing part would depend on compiler backend being used.

vesajaaskelainen:
@Bertk Sadly sonar.cxx.defiens/forceIncludes is not flexible enough for the job. This is the reason why I have went extra mile to define a way to make it possible to configure different settings for different configuration units.

I would say this specification file is more or less internal file format for sonar-cxx. As it is not proprietary file format here (eg. not generated by any existing tools). I would even say it is open file format.

You can think this as way to integrate another toolchain support (vs. msbuild) for sonar-cxx usage that does not require built-in support from sonar-cxx plugin. Instead it tries to be generic way to provide the information. If you can think another way to extend the sonar-cxx to carry same information then that could be considered too here.

Anyway with this we can analyse our files without parsing errors and lot of complains about missing headers files. Before this change there were thousands of errors that didn't have anything to do with code quality but just tool not working properly. Now we can actually see what is wrong with the code as it is analysed properly. (except for header files where we have to cheat a bit due to current operation model of sonar-cxx)

jmecosta:
@vesajaaskelainen @Bertk meant is if you use msbuild sonar qube runner, in principle the settings that are set in the sonar.cxx.defiens/forceIncludes are per module/project. So for example if you are in MSBuild world those should be the tailored to the project and not globally. So in principle is able to do what described without introducing another external format.

also the support for compilers logs is likely the easiest as well accurate for end users, if we implement the stuff that is in this pull request (get compilation setting per unit) on top of the logs parsing then it should provide basically all we need. MSBuild log parsing is ready, others can also be added.

by the way, what compiler are you using?

vesajaaskelainen:
@jmecosta Somehow this module/project approach in sonar scanner is too restricted. I have troubles on getting it configured properly as our code based do not follow "Java" style of code bases. Eg. they are split a bit there and there. I ended up in some cases just using one setting for project and just selecting source folders for that.

Especially partly shared contents are hard to split properly. And they tend to be compiled a bit differently form where they are used.

You can think this like Linux kernel's __KERNEL__ define that differentiates user space and kernel code parts in headers.

Toolchains that we use range from (but not limited to) gcc 2.95, 4.x, 5.x, 6.x (most in cross compilation mode), msvc (different versions), IAR EWARM. Then there are multiple different built systems from autotools, cmake, scons, msbuild, iarbuild and plain hand crafted makefiles. Some of our code utilize Yocto Project's bitbake build system.

We have license for Klocwork C/C++ static source code analyser which we use to scan all those. It provides kwinject command that listens for different compilations and their records settings and then creates internal representation of that information for its own analysing phase. What I did was translate this Klocwork's internal representation of this information to more easily accessible XML representation that could be used with different tooling. With this I don't need to care what compiler is being used as kwinject works with many different compilers out-of-the-box.

I believe this is better approach vs log parsing as I don't need to take care whether target is ARM, x86, 68000 or something else. kwinject figures out target architecture settings eg. compiler's internal defines and includes and then it gets include/defines settings from compiler's command lines being executed (or however it listens what is happening).

Due to large amount of projects and different compilers handling this by hand is out-of-question. So there really needs to be automated support.

I assumed you do not want to have Klocwork specific support as that would be proprietary tool support and its internal format is subject to change between versions so I decided to make intermediate XML format to piggying the information.

jmecosta:
@vesajaaskelainen there was a request earlier to support similar thing, https://github.com/SonarOpenCommunity/sonar-cxx/issues/421 it was closed because no one showed interest in implement it. How would that sound to you? Considering that the compilation database format is more likely to be supported by different build systems, it might be a better approach then creating a new format? 

the build log parsing for msbuild, imo is good approach, since what goes in there are the real evaluated values for all properties so i dont need to wonder about if setting this property like this what will happen to the include paths etc. simplifies a lot the maintenance of sonar, the only thing missing is produce compilation units inside the plugin

vesajaaskelainen:
@jmecosta I believe this is more or less similar thing vs. #421. Expect of using JSON -- this PR uses XML files (as that was already used in project). If you prefer JSON that is fine, I can adapt for that too. 

In this PR compiler command line is processed off-line from sonar-cxx's point of view. Problem of analysing only compiler command line is that when you build cross compilation tool most of the compiler defines are hidden from command line. One would need to analyse compiler binary, or execute it with special arguments to capture default defines and includes. For us Klocwork's kwinject does this job. Kwinject internally has dozen of python files that analyse different compiler invocations and tries to capture most important details.

I did not spot any compilation database format on that ticket?

One could think this XML file as compilation database as it includes all needed details. I am fine on using something else as intermediate file format (or method of transmitting the information) as long as it contains the needed details. Would it help if I would use/create some example project and attach generated XML here to get better insights what is the idea here?

To be complete needed details here are:
- capture of compiler's defines and includes for target device (for most compilers target selection arguments affect what is being used, with gcc some are hidden inside binary and depends on gcc's compilation settings in use)
- there can be different compilers in project (as an example gcc and g++) -- they have different set of includes in place
- some files get compiled with different settings
- some files may even get compiled multiple times with different settings
- analysis would need to act like real compilation process

If intermediate format would support all that then you have all means to properly analyse your source code. With this PR sonar-cxx doesn't support multiple compilation nor does it act like real compilation process but others are handled assuming something provides compilation unit specification file. Nothing restricts on improving it later on to support all of those.

If you prefer to have support directly for Klocwork's compiler capture format, I can do that too. I was just thinking that you might prefer more generic solution inside (https://github.com/SonarOpenCommunity/sonar-cxx/pull/993/commits/1067d16999212f7855d45704e283d0cfe959651b) that then outer layer would support the file format allowing other types of compilation capture to be used.

jmecosta:
this is the format:  http://clang.llvm.org/docs/JSONCompilationDatabase.html, as far as i know its a more generic approach than own xml, i am not sure about the others but i think but using it would be more beneficial to everyone and not only for Klocwork's (assuming that they would need to implemet a interceptor and coverter to your format)

I am not sure about other compilers, however msbuild with diagnostic information on (the one we request /v:diag and the parser runs on top of that) sptis out all the needed information regarding build information.

What i think it will not be possible is to define a state in the plugin for every possible configuration/platform that you have defined in the compilation database. As this will likely bring 2 much complexity that we dont want to maintain. 

The idea of the plugin is that it analysis a single setup, the build information should be according with that setup. More than that the plugin woulnd work if not mistaken, but @guwirth can confirm this

vesajaaskelainen:
@jmecosta Problem with that file format is that it cannot provide need defines nor includes.

One could extend that format thou with "includes" keyword with data type of array of strings and "defines" with data type of object with key-value pairs. I can switch to this JSON format with needed local extensions if wanted.

I have implemented this state change in this PR, there are "quite" clear entry and exit points so from that point of view I do not think complexity is a problem. When compilation unit changes (file name changes) it gets updated based information for target compilation unit (lookup based on source file being analysed). Or if that is missing the use global settings.

jmecosta:
```
[
  { "directory": "/home/user/llvm/build",
    "command": "/usr/bin/clang++ -Irelative -DSOMEDEF=\"With spaces, quotes and \\-es.\" -c -o file.o file.cc",
    "file": "file.cc" },
  ...
]
```

in this case,
include are: -Irelative
Define: -DSOMEDEF
did i miss something?

Regarding the complexity, i think we are looking from different perspectives. Right? you handle all that from the external tool, what the cxx plugin consumes is a clearly defined set of includes, defines (or global defines if not availalbe in your format)

vesajaaskelainen:
@jmecosta: That expression says that this file will be compiled in this directory with this command line arguments. It does not provide view what internal compiler definitions are in use. This is highly important when using cross compilers especially there are 50-100 different defines being set by compiler internally which affect system headers processing and others to code analysis so that proper code is analyzed. 

System includes are not under /usr/include they are in sysroot directory somewhere. Compilers have multiple internal include directories that has effect on compilation. Then settings like soft/hard floating point compiler settings generate different defines thus having effect on what include files to process. What compiler extensions were used when compiling code. Then there are some settings that define what data types mean eg. what does int8_t mean. How many bytes does long consume. What are value ranges for long and so on.

vesajaaskelainen:
Here is example XML file for simple hello world C application that is cross compiled.
[hello.xml.txt](https://github.com/SonarOpenCommunity/sonar-cxx/files/589583/hello.xml.txt)

This is a bit simple example so it does not show flexibility what format provides but gives you an idea what kind of data there is.

I anonymised data a bit.

jmecosta:
We also derive the internal state of Vc++ based on log, see https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-squid/src/main/java/org/sonar/cxx/CxxVCppBuildLogParser.java where all the compiler flags, defines, and includes (system includes) are retrieved based on the build log, that has been created based on https://msdn.microsoft.com/en-us/library/vstudio/b0084kay(v=vs.140).aspx 

For your case in theory you can dump all those definitions into the command line in the compilation database, but my feeling is that it would work more like the VC++ compiler. Where retrieve most of those internal states based on the compiler that is used (this was likely the reason @Bertk refered that we could extend the support to other compilers) and a such complex set of defines and includes would not likely be needed.

It all comes down, were do we put complexity. In the plugin, so implement all those internal states as we did for the VC++ or ask each user to create that information. If that is the case, i am ensure how many users will be able to come up with that list, since this is a klockworth output and most users will not be able to generated if usiing simply gcc in verbose mode. 

Regarding this pr itself, my only comment is about the format. Supporting a clang compilation database seems to be a more wider and supported scenario. Where users using cmake can create this information easily and feed it to the cxx plugin. Later improvemnt could include check the compiler used and further implement internal states.  But i dont know about the

vesajaaskelainen:
@jmecosta With gcc/g++ you can get this kind of data with command line like following:

gcc -dM -E - < /dev/null
g++ -dM -E - < /dev/null

Catch here is that you need to also provide arch settings so that it adapts a bit.

In our case this would be like ($CC -dM ...):

arm-custom-linux-gnueabi-gcc -march=armv7-a -mfloat-abi=hard -mfpu=neon -mtune=cortex-a8 --sysroot=/foobar/sysroots/cortexa8hf-vfp-neon-custom-linux-gnueabi -dM -E - < /dev/null

Supporting different compilers and their versions is quite huge effort especially when you include cross compilers in mix which are highly tuned for target architectures (and custom compilations like in when using Yocto Project). I would say from sonar-cxx plugins point of view it would be easier if there would be some process before that to capture actual build, alternatively there could be plugin architecture in sonar-cxx that can be used to piggyback build information.

As you seem to prefer that JSON format -- do you mind if I extend it as I proposed above defines+includes for compiler internal stuff and then re-construct simulated compiler command line? I would like to keep compiler internal stuff away from simulated command line, it will just confuse user as it would not match used command line. Is Jackson OK for parsing JSON files? As that would be new dependency. I can update the PR to match JSON format.

jmecosta:
@vesajaaskelainen the main idea would be to support the main compilers. Your case, as you mention would be a huge effort. I like the idea of having a backdoor that allows the user to define all those, and i supose augmenting the json for me seems ok.  So ok from my side on that. This should cover also the clang compilation database, that some other users requested. Please still wait for feedback from @guwirth and @Bertk regarding the general approach

i dont know Jackson OK, but just be sure that it does not add dependencies to guava and other libs that you dont find already in the plugin. Those have cause issues in past since Sonar Core has their own versions of those libs

Bertk:
@vesajaaskelainen I currently use the VC log file parser and the MSBuild definitions to collect all required definitions and include paths for the SonarQube analysis. The Klocwork's kwinject way is another more flexible solution but I do not need it and it adds some overhead to the build and analysis process. I would appreciate a more general and easier way to manage pre-defined compiler macros and definitions but I am also not sure whether the spec-file-support PR is the solution for this.

vesajaaskelainen:
Sorry that it took a while. But I have now implemented proposed JSON Compilation Database support with extension to support specifying defines and includes as override for analysis.

Tried out with CMake to generate the file and it kinda works but with that method one can only capture command line arguments and not knowledge of compiler's internal defines and include directories.

Command line argument parser is model quick and dirty assuming that -I or -D is used to define includes and defines.

sonar-cxx-plugin/pom.xml
```diff
@@ -100,6 +100,16 @@
       <version>2.6</version>
       <type>jar</type>
     </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
```
guwirth: What's the advantage of using this library?

vesajaaskelainen: It is widely used where ever I have seen JSON being processed with Java. No other than that.

vesajaaskelainen:
Now it is rebased and it has passed your CI tests. Also updated copyright years for licenses as it was checking now for 2017.

I can write something to wiki about it. Do you have proposal for page name / location for it?

jmecosta:
@vesajaaskelainen i think this would be https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Compilers 

a side question if may i ask, you mention klocwork. Are you using it? how have you endup using cxx plugin, since klocwork by itself already provide all static code analysis you need.

vesajaaskelainen:
@jmecosta We use Klocwork to analyze our C/C++ code as it kinda does much deeper analysis of the code. But it only works in AST level. So Klocwork acts like a compiler and linker to do its analysis. Sonar C++ plugin works a bit differently by looking at source files and from that it can also analyze different properties of the code. What we do is we import Klocwork results (as much as it can be imported) to Sonar so that we get better view of the quality of the software (metrics). Then we use Sonar to observe quality of our code (dashboards). (We are also developing some Java code which is the reason we have Sonar initially here)

Problem without the changes here Sonar was way too noisy as it didn't know compiler symbols or project symbols nor include files -- and by improving Sonar C++ plugins symbol knowledge and include knowledge we get more sane results from Sonar C++ plugin too. So we can actually benefit from both tools.

To improve the for Sonar C++ plugin we use Klocwork to capture compiler internals and command lines so that it provides the knowledge for Sonar C++ plugin that it cannot provide just by looking at the code.

We use the same analysis pipe to analyse normal Linux desktop code, Windows desktop code and then cross compiled embedded Linux code and other cross compiled embedded code.

With this combo we can do all this.

guwirth:
@vesajaaskelainen would add description to 'Going deeper / JSON Compilation Database support'

jmecosta:
@guwirth can we clean up the documentation a bit, in the compilers part there is stuff like parsing the buildlog to get compiler information. this is related. So if we create the section in Going Deeper has you suggested but making it like "Going deeper/Reusing build information" inside we can put, the json, the buildlog, tfs builds, and cxx wrapper?

guwirth:
@jmecosta think you are right that the whole wiki must be revised. But I think that's not part of this PR. Would be happy if at least all new features have a docu. Cleaning it up can be done afterwards.

guwirth:
@vesajaaskelainen please rebase that we can merge it.

vesajaaskelainen:
Wiki https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Compilers has been update with JSON Compilation database.

vesajaaskelainen:
This PR has been rebased.

guwirth:
@vesajaaskelainen thanks a lot for providing this extension and the docu!

Issue #1011, opened 2016-12-21
Support CppCheck 1.77


guwirth:
Support CppCheck 1.77, due date is 31.12.2016: http://trac.cppcheck.net/roadmap

PR #1020, opened 2016-12-29
Support CppCheck V1.77 messages

Bertk:
#1011


guwirth:
@Bertk thx and a good start into 2017!

Issue #1009, opened 2016-12-16
#include for files without extension are not handled correctly


sreeramk08:
Sonar-cxx plugin complains that iostream is missing while the problem really is that the developer forgot to add the .h to iostream.  I've already verified that iostream.h exists in one of the include paths.  

Is there a way to tell Sonarqube not to flag that as a missing header file?

guwirth:
@sreeramk08 Which error message do you get? Normally you can ignore missing includes.

sreeramk08:
There are lot of Warnings like this
WARN: [/home/skrishna/jenkins/workspace/Sonarqube/full-xcipio-sonar-analysis/xcipio/LIS/inc/CORE/common/GMOSHelper.H:37]: cannot find the sources for '#include <string>'

WARN: [/home/skrishna/jenkins/workspace/Sonarqube/full-xcipio-sonar-analysis/xcipio/LIS/inc/CORE/common/GMOSHelper.H:39]: cannot find the sources for '#include <cnfg.h>'

As you can see the first warning is something I don't want because string.h is very much available.  The code has been written without a .h. 

But the second warning is a legitimate one which means a .h file is missing and the team is building applications without realizing it.

sreeramk08:
For the moment since I'm not including the rule which says 'flag missing include files' the analysis does not show these missing includes.  But the scan log on jenkins is full of them.

jmecosta:
And you have all include paths defined for the cxx analysis in the sonar.props file?

sreeramk08:
That is correct.  I took one, for example string.h, then ran a find command from the top level of the checkout directory...and it found it in a path.  Then I made sure the path was included in the sonar's include directories list. 

And like I mentioned above, some are legitimate missing header files which sonar is correctly flagging which I could not find as expected.

jmecosta:
i think we are missing this fallback in the search function

https://github.com/SonarOpenCommunity/sonar-cxx/blob/985c0d7849dffe25eff00aa692bcbd4a1d972e63/cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java#L1120

guwirth:
@jmecosta 

> i think we are missing this fallback in the search function

What exactly are we missing?

sreeramk08:
Any updates on this?

guwirth:
@sreeramk08 for me it's not clear what exactly is missing? Answer from @jmecosta is missing.

[...]

PR #1021, opened 2016-12-30
#1009 support finding include without extension

jmecosta:
#1009 fix

### DIFF
--- a/cxx-squid/src/test/java/org/sonar/cxx/preprocessor/SourceCodeProviderTest.java
+++ b/cxx-squid/src/test/java/org/sonar/cxx/preprocessor/SourceCodeProviderTest.java
@@ -25,6 +25,8 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.charset.Charset;
 import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
 
 import org.junit.Test;
 
@@ -121,6 +123,9 @@ public class SourceCodeProviderTest {
   public void getting_file_with_filename_and_cwd_no_extension() {
     String cwd = new File("src/test/resources/codeprovider").getAbsolutePath();
     String path = "source";
+    List headers = new LinkedList<>();
+    headers.add(".hh");
+    codeProvider.setHeadearSuffixes(headers);
     assertEquals(expected, codeProvider.getSourceCodeFile(path, cwd, true));
     assertEquals(null, codeProvider.getSourceCodeFile(path, cwd, false));
   }
--- a/cxx-squid/src/test/java/org/sonar/cxx/preprocessor/SourceCodeProviderTest.java
+++ b/cxx-squid/src/test/java/org/sonar/cxx/preprocessor/SourceCodeProviderTest.java
@@ -118,6 +118,14 @@ public class SourceCodeProviderTest {
   }
 
   @Test
+  public void getting_file_with_filename_and_cwd_no_extension() {
+    String cwd = new File("src/test/resources/codeprovider").getAbsolutePath();
+    String path = "source";
+    assertEquals(expected, codeProvider.getSourceCodeFile(path, cwd, true));
+    assertEquals(null, codeProvider.getSourceCodeFile(path, cwd, false));
+  }
+  
+  @Test
   public void getting_file_with_relpath_and_cwd() {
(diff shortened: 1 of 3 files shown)


jmecosta:
@guwirth this should handle it

cxx-squid/src/test/java/org/sonar/cxx/preprocessor/SourceCodeProviderTest.java
```diff
@@ -118,6 +118,14 @@ public void getting_file_with_filename_and_cwd() {
   }
 
   @Test
+  public void getting_file_with_filename_and_cwd_no_extension() {
+    String cwd = new File("src/test/resources/codeprovider").getAbsolutePath();
+    String path = "source";
```
jmecosta: @guwirth this is the file without extension, correctly lookup up

jmecosta:
yep, i use sonar.cxx.suffixes.headers

guwirth:
Hello @jmecosta,

looking deeper into this my understanding is:
* that we have to add a new test file to https://github.com/SonarOpenCommunity/sonar-cxx/tree/985c0d7849dffe25eff00aa692bcbd4a1d972e63/cxx-squid/src/test/resources/codeprovider ==> source (without extension)
* a new test to https://github.com/SonarOpenCommunity/sonar-cxx/blob/985c0d7849dffe25eff00aa692bcbd4a1d972e63/cxx-squid/src/test/java/org/sonar/cxx/preprocessor/SourceCodeProviderTest.java

```Java
 private final File expected2 = new File(new File("src/test/resources/codeprovider/source").getAbsolutePath());
```

And test like this:

```Java
@Test
  public void getting_file_relpath_without_extension() {
    String baseDir = new File("src/test").getAbsolutePath();
    String dummycwd = "/";
    String path = "source";
    String includeRoot = new File("src/test/resources/codeprovider").getAbsolutePath();

    codeProvider.setIncludeRoots(Arrays.asList(includeRoot), baseDir);
    assertEquals(expected2, codeProvider.getSourceCodeFile(path, dummycwd, true));
    assertEquals(expected2, codeProvider.getSourceCodeFile(path, dummycwd, false));
  }
```
Is my understanding right?

jmecosta:
@guwirth i think the prob is when a file has a extension but not included in the include directive, the current code (without this) handles files without extensions already. the test ive added covers this limitation

cxx-squid/src/test/java/org/sonar/cxx/preprocessor/SourceCodeProviderTest.java
jmecosta: @guwirth this is supposed to mimic a include directive, its not the real source file in filesystem

guwirth:
@jmecosta after playing around with VS: the point is that `#include <iostream>` tries to open a file `ìostream` and not `iostream.h`. So your solution must realy try to open a file without extension and not iterate over `sonar.cxx.suffixes.headers`- sorry for my first wrong hint.

jmecosta:
@guwirth this is how its handled now. it first checks if can find filepath (with or witout extension). if it does not find and it does not contain "." then it fallbacks to other extensions

guwirth:
@jmecosta 

>if it does not find and it does not contain "." then it fallbacks to other extensions

But I think this is wrong. Why should it fallback to other extensions?

jmecosta:
its the case of the #include <string> the user reported. where string.h is available however if we dont fallback it will never be lookup

guwirth:
@jmecosta reading #1009 the issue seems to be that we are not able to handle include files without extension. @sreeramk08 is not asking for a fallback? Please @sreeramk08 correct me if I'm wrong.

jmecosta:
@guwirth the current implementation works fine without extension. file.exists() should handle both files with or without extension. or @wichtounet request is not valid for the current implementation. but this implementation is my understanding of the issue

wichtounet:
@jmecosta I dont' think it was me ;P

jmecosta:
@wichtounet  sorry, i meant @sreeramk08

jmecosta:
in fact current implementation is as per standard. visual studio is also unable to find header if extension is not provided. and string header is found as string in include folders not string.h.

PR #1023, opened 2016-12-30
test source code provider for include files without extension

guwirth:


guwirth:
@jmecosta current solution is working for include files without extension.

Why I'm thinking iterating over diferent extensions is wrong: http://stackoverflow.com/questions/441568/when-can-you-omit-the-file-extension-in-an-include-directive

jmecosta:
@guwirth yes there is no issues with the current implementation, everything works as long as the include file is used as its found in filesystem (with or without extension)

the problem is only if a user uses a #include in source file without extension and its found in filesystem with extension. 

ive just tested this in visual studio, and it does not work. vs is  not able to find for example vadefs when its available as vadefs.h

so i suppose this must have been a badly defined include path for @sreeramk08.  @sreeramk08 please confirm your include paths, or please clarify what is the thing we are not getting from your use case.

guwirth:
I merge this to improve testing.

PR #1032, opened 2017-01-15
fix CPD sensor

guwirth:
- feedback from developer is that default should be sonar.cxx.cpd.ignoreLiterals=False & sonar.cxx.cpd.ignoreIdentifiers=False (not True)
- CPD sensor is looking now only in functionDefinition for duplicates. Gave false positives e.g. with interfaces before


guwirth:
@all ready to review and merge, please give feedback

PR #1068, opened 2017-02-24
support custom rules in cxx-lint

jmecosta:
supports analysis with template rules for local analyses


guwirth:
Did a squash & merge

PR #1088, opened 2017-03-10
fix Clang-Tidy error

jmecosta:


guwirth:
@jmecosta wouldn't it be better
* to use first the part before the first : 
* check if file path, normalize
* and the parse the rest of the line?

jmecosta:
@guwirth nevermind, i can make it all with a single regular expression. That would be:

(.+|[a-zA-Z]:\\.+):([0-9]+):([0-9]+): ([^:]+): ([^]]+)

PR #1096, opened 2017-03-21
Add an XSLT pre-analyse step

### LINKED ISSUE #1080: Add an XSLT pre-analyse step
Hi,

would be useful to have an build-in XSLT transformation converting XML files to the format needed by our plugin.
* the XSLT transformation should be executed prior to any other steps
* should support more then one transformation for the case different inputs should be supported
* should support source and output with placeholder in the name

Sample:
```
sonar.cxx.xslt.1.stylesheet=MyStylesheets1.xslt
sonar.cxx.xslt.1.source=tool1/reports/xxx-*.xml
sonar.cxx.xslt.1.output=reports/other-tool1-*.xml

sonar.cxx.xslt.2.stylesheet=MyStylesheets2.xslt
sonar.cxx.xslt.2.source=tool2/reports/xxx-*.xml
sonar.cxx.xslt.2.output=reports/other-tool2-*.xml
```
What are you thinking?

Regards,

stefanweiser:
Hi folks,

I implemented an XSLT translation. I currently does not support wildcards (don't know how to get the concrete file list out of a wildcard pattern in Java), but multiple input and output files seperated by comma. The translation of files is done before using the report file of the output sensor, so:

`sonar.cxx.xslt.1.stylesheet=MyStylesheets1.xslt
sonar.cxx.xslt.1.source=tool1/reports/xxx-1.xml
sonar.cxx.xslt.1.output=reports/other-tool1-1.xml

sonar.cxx.xslt.2.stylesheet=MyStylesheets2.xslt
sonar.cxx.xslt.2.source=reports/other-tool1-1.xml
sonar.cxx.xslt.2.output=reports/other-tool2-2.xml

sonar.cxx.other.reportPath=reports/other-tool2-2.xml`

will report "reports/other-tool2-2.xml" after translation from tool1/reports/xxx-1.xml to reports/other-tool1-1.xml to reports/other-tool2-2.xml.

I tested it on our own server, but was not able to provide a test case. I think I'm not able to understand the unit test infrastructure...

close #1080

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1096)

### DIFF
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/other/CxxOtherSensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/other/CxxOtherSensor.java
@@ -20,8 +20,16 @@
 package org.sonar.cxx.sensors.other;
 
 import java.io.File;
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.util.Arrays;
+import java.util.List;
 import javax.xml.stream.XMLStreamException;
+import javax.xml.transform.Source;
+import javax.xml.transform.TransformerException;
+import javax.xml.transform.stream.StreamSource;
 
+import org.apache.commons.io.FilenameUtils;
 import org.codehaus.staxmate.in.SMHierarchicCursor;
 import org.codehaus.staxmate.in.SMInputCursor;
 import org.sonar.api.batch.sensor.SensorContext;
@@ -29,24 +37,32 @@ import org.sonar.api.batch.sensor.SensorDescriptor;
 import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
 import org.sonar.cxx.CxxLanguage;
+import org.sonar.cxx.sensors.utils.CxxMetrics;
+import org.sonar.cxx.sensors.utils.CxxUtils;
 import org.sonar.cxx.sensors.utils.CxxReportSensor;
 import org.sonar.cxx.sensors.utils.StaxParser;
 
 /**
  * Custom Rule Import, all static analysis are supported.
  *
- * @author jorge costa
+ * @author jorge costa, stefan weiser
  */
 public class CxxOtherSensor extends CxxReportSensor {
   public static final Logger LOG = Loggers.get(CxxOtherSensor.class);
   public static final String REPORT_PATH_KEY = "other.reportPath";
-  public static final String KEY = "Other";
+  public static final String KEY = "other";
+  public static final String OTHER_XSLT_KEY = KEY + ".xslt.";
+  public static final String STYLESHEET_KEY = ".stylesheet";
+  public static final String INPUT_KEY = ".inputs";
+  public static final String OUTPUT_KEY = ".outputs";
+  private CxxLanguage cxxLanguage;
 
   /**
    * {@inheritDoc}
    */
   public CxxOtherSensor(CxxLanguage language) {
     super(language);
+    this.cxxLanguage = language;
   }
 
   @Override
@@ -58,11 +74,20 @@ public class CxxOtherSensor extends CxxReportSensor {
   public void describe(SensorDescriptor descriptor) {
     descriptor.onlyOnLanguage(this.language.getKey()).name(language.getName() + " ExternalRulesSensor");
   }
-  
+
+  /**
+   * {@inheritDoc}
+   */
   @Override
-  public void processReport(final SensorContext context, File report) throws XMLStreamException {
+  public void execute(SensorContext context) {
+    transformFiles(context.fileSystem().baseDir());
+    super.execute(context);
+  }
+
+  @Override
+  public void processReport(final SensorContext context, File report) throws XMLStreamException, IOException, URISyntaxException, TransformerException {
     LOG.debug("Parsing 'other' format");
-    
+
     StaxParser parser = new StaxParser(new StaxParser.XmlStreamHandler() {
 
       /**
@@ -86,9 +111,67 @@ public class CxxOtherSensor extends CxxReportSensor {
 
     parser.parse(report);
   }
-  
+
   @Override
   protected String getSensorKey() {
     return KEY;
   }  
+
+  public void transformFiles(final File baseDir) {
+    Boolean goOn = true;
+    for (int i = 1; (i < 10) && goOn; i++) {
+      String stylesheetKey = OTHER_XSLT_KEY + i + STYLESHEET_KEY;
+      String inputKey = OTHER_XSLT_KEY + i + INPUT_KEY;
+      String outputKey = OTHER_XSLT_KEY + i + OUTPUT_KEY;
+
+      String stylesheet = resolveFilename(baseDir.getAbsolutePath(), cxxLanguage.getStringOption(stylesheetKey));
+      List<File> inputs = getReports(cxxLanguage, baseDir, inputKey);
+      String outputStrings[] = language.getStringArrayOption(outputKey);
+      List<String> outputs = Arrays.asList((outputStrings != null) ? outputStrings : new String[] {});
+
+      if (inputs.size() != outputs.size()) {
+        LOG.error("Number of source XML files is not equal to the the number of output files.");
+        goOn = false;
+      } else if ((stylesheet != null) ||
+        ((inputs != null) && (inputs.size() > 0)) ||
+        ((outputs != null) && (outputs.size() > 0))) {
+        if (stylesheet == null) {
+          LOG.error(stylesheetKey + " is not defined.");
+          goOn = false;
+        } else if (inputs == null) {
+          LOG.error(inputKey + " file is not defined.");
+          goOn = false;
+        } else if (outputs == null) {
+          LOG.error(outputKey + " is not defined.");
+          goOn = false;
+        } else {
+          LOG.debug("Converting " + stylesheet + " with " + inputs.toString() + " to " + outputs.toString() + ".");
+          File stylesheetFile = new File(stylesheet);
+          if (stylesheetFile.isAbsolute()) {
+            transformFileList(baseDir.getAbsolutePath(), stylesheetFile, inputs, outputs);
+          }
+        }
+      } else {
+          // No keys found.
+          goOn = false;
+      }
+    }
+  }
+
+  private void transformFileList(final String baseDir, File stylesheetFile, List<File> inputs, List<String> outputs) {
+    for (int j = 0; j < inputs.size(); j++) {
+      try {
+        String normalizedOutputFilename = resolveFilename(baseDir, outputs.get(j));
+        CxxUtils.transformFile(new StreamSource(stylesheetFile), inputs.get(j), new File(normalizedOutputFilename));
+      } catch (Exception e) {
+        String msg = new StringBuilder()
+          .append("Cannot transform report files: '")
+          .append(e)
+          .append("'")
+          .toString();
+        LOG.error(msg);
+        CxxUtils.validateRecovery(e, cxxLanguage);
+      }
+    }
+  }
 }
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxUtils.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxUtils.java
@@ -25,6 +25,14 @@ import java.io.StringWriter;
 import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
 import org.sonar.cxx.CxxLanguage;
+import javax.xml.transform.OutputKeys;
+import javax.xml.transform.Result;
+import javax.xml.transform.Source;
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerException;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.stream.StreamResult;
+import javax.xml.transform.stream.StreamSource;
 
 
 /**
@@ -65,13 +73,20 @@ public final class CxxUtils {
     if (targetfile.isAbsolute()) {
       filePath = normalizePath(filename);
     } else {
-      // RATS, CppCheck and Vera++ provide names like './file.cpp' - add source folder for index check
+      // RATS, CppCheck and Vera++ provide names like './file.cpp' - add input folder for index check
       filePath = normalizePath(baseDir + File.separator + filename);
     }
     return filePath;
   }
-  
-  
+
+
+  public static void transformFile(Source stylesheetFile, File input, File output) throws TransformerException {
+    TransformerFactory factory = TransformerFactory.newInstance();
+    Transformer transformer = factory.newTransformer(stylesheetFile);
+    transformer.setOutputProperty(OutputKeys.INDENT, "yes");
+    transformer.transform(new StreamSource(input), new StreamResult(output));
+  }
+
 
   /**
    * <p>Gets the stack trace from a Throwable as a String.</p>
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/tests/xunit/CxxXunitSensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/tests/xunit/CxxXunitSensor.java
@@ -27,14 +27,8 @@ import java.util.List;
 import java.util.regex.Pattern;
 import javax.xml.stream.XMLStreamException;
 
-import javax.xml.transform.OutputKeys;
-import javax.xml.transform.Result;
 import javax.xml.transform.Source;
-import javax.xml.transform.Templates;
-import javax.xml.transform.Transformer;
 import javax.xml.transform.TransformerException;
-import javax.xml.transform.TransformerFactory;
-import javax.xml.transform.stream.StreamResult;
 import javax.xml.transform.stream.StreamSource;
 
 import org.sonar.api.batch.sensor.SensorContext;
@@ -251,15 +245,8 @@ public class CxxXunitSensor extends CxxReportSensor {
       }
 
       Source xsl = new StreamSource(inputStream);
-      TransformerFactory factory = TransformerFactory.newInstance();
-      Templates template = factory.newTemplates(xsl);
-      Transformer xformer = template.newTransformer();
-      xformer.setOutputProperty(OutputKeys.INDENT, "yes");
-
-      Source source = new StreamSource(report);
       transformed = new File(report.getAbsolutePath() + ".after_xslt");
-      Result result = new StreamResult(transformed);
-      xformer.transform(source, result);
+      CxxUtils.transformFile(xsl, report, transformed);
     } else {
       LOG.debug("Transformation skipped: no xslt given");
     }
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportSensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportSensor.java
@@ -124,6 +124,23 @@ public abstract class CxxReportSensor implements Sensor {
     return value;
   }
 
+  public static String resolveFilename(final String baseDir, final String filename) {
+
+    // Normalization can return null if path is null, is invalid, or is a path with back-ticks outside known directory structure
+    String normalizedPath = FilenameUtils.normalize(filename);
+    if ((normalizedPath != null) && (new File(normalizedPath).isAbsolute())) {
+      return normalizedPath;
+    }
+
+    // Prefix with absolute module base dir, attempt normalization again -- can still get null here
+    normalizedPath = FilenameUtils.normalize(baseDir + File.separator + filename);
+    if (normalizedPath != null) {
+      return normalizedPath;
+    }
+
+    return null;
+  }
+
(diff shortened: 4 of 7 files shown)


sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
```diff
@@ -32,22 +38,29 @@
...
 /**
  * Custom Rule Import, all static analysis are supported.
  *
- * @author jorge costa
+ * @author jorge costa, stefan weiser
  */
 public class CxxExternalRulesSensor extends CxxReportSensor {
   public static final Logger LOG = Loggers.get(CxxExternalRulesSensor.class);
   public static final String REPORT_PATH_KEY = "sonar.cxx.other.reportPath";
+  public static final String SONAR_CXX_XSLT_KEY = "sonar.cxx.xslt.";
```
guwirth: Not sure, but I think should be in an extra sensor and not within CxxReportSensor?

guwirth: If yes: must be clarified how to define order of sensor execution

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
```diff
@@ -87,4 +119,48 @@ public void stream(SMHierarchicCursor rootCursor) throws XMLStreamException {
...
   }
+
+  private void transformFiles() throws TransformerException {
+    for(int i = 1; i < 100; i++) {
+      String stylesheetKey = SONAR_CXX_XSLT_KEY + i + STYLESHEET_KEY;
+      String sourceKey = SONAR_CXX_XSLT_KEY + i + SOURCE_KEY;
+      String outputKey = SONAR_CXX_XSLT_KEY + i + OUTPUT_KEY;
+
+      String stylesheet = FilenameUtils.normalize(settings.getString(SONAR_CXX_XSLT_KEY + i + STYLESHEET_KEY));
+      String[] sources = settings.getStringArray(SONAR_CXX_XSLT_KEY + i + SOURCE_KEY);
```
guwirth: Would not support comma separated list. Would use wildcard pattern instead

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
```diff
@@ -87,4 +119,48 @@ public void stream(SMHierarchicCursor rootCursor) throws XMLStreamException {
...
+      String[] outputs = settings.getStringArray(SONAR_CXX_XSLT_KEY + i + OUTPUT_KEY);
+
+      if (sources.length != outputs.length) {
+        LOG.error("Number of source XML files is not equal to the the number of output files.");
+      } else if ((stylesheet != null) ||
+        ((sources != null) && (sources.length > 0)) ||
+        ((outputs != null) && (outputs.length > 0))) {
+        if (stylesheet == null) {
```
guwirth: Think we need always all three. So one message would be sufficient?

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
```diff
@@ -59,11 +72,30 @@ protected String reportPathKey() {
...
+
+  /**
+   * {@inheritDoc}
+   */
   @Override
-  public void processReport(final SensorContext context, File report) throws XMLStreamException {
+  public void execute(SensorContext context) {
+    try {
+      transformFiles();
+    } catch (Exception e) {
```
guwirth: Think try/catch is at the wrong place. If one transformation is failing no other is executed?
Would prefer: if one transformation is failing, should try the next one

guwirth:
* unit test is missing
* UI integration is missing
   * think we need then hard coded properties?
* if we add UI: would start with 5 only (instead 10)

guwirth:
@stefanweiser thanks a lot for providing this solution. I started a review with some general questions.

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
stefanweiser: I was uncertain, too. I guess, it's a question of: Which other sensor could benefit? I don't see any use in transforming a report, that already fits to the sensor (which is the case elsewhere).

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
stefanweiser: Doing so would then result in a lot of errors if one transformation fails, because the source of the next transformation step would be missing. I know, that this would be more tolerant.

Its a design decision, which I would leave up to you. So if you want to go on transforming files, even there are errors, then I would implement it that way.

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
jmecosta: I do agree with Stefan, this kind of transformation seems applicable only to external sensor since it's the only one that forces users to provide a specific format, @guwirth which other sensors do you think would benefit from this?

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
stefanweiser: I thought, that getStringArray would return an array of strings, that are comma-seperated in the settings file and I don't know how to solve wildcard patterns. Is there any reference implementation for this in sonar-cxx?

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
stefanweiser: We do need always all three, but I appreciate a tool, that reports to me exactly what's wrong. Otherwise you have to do some research first to find the bug.

stefanweiser:
I want to write a unit test for this, but I was not able get it to work. Seems, that I have to put more effort in this.

What do you mean with UI-integration? How do I get that?

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
guwirth: There is also the XUnit sensor using a XSLT doing the same: https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/tests/xunit/CxxXunitSensor.java#L233. Question is if we wanna have two?

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
guwirth: Up to now we have two modes: strict and tolerant
* strict: break whole analysis after an error
* tolerant: ignore and continue

This is hidden behind CxxUtils.validateRecovery.

In tolerant mode:
* should continue if one of the XSLT transformation fails and there are more source files in the comma separated list
* should continue with next XSLT transformation if one before was failing

stefanweiser: So you suggest different behavior for both modes? ;-)

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
guwirth: We are doing this in CxxReportSensor.getReports. This is a static function, maybe we can move it https://github.com/SonarOpenCommunity/sonar-cxx/tree/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils to make it more general?

https://github.com/SonarOpenCommunity/sonar-cxx/blob/056fbb3fcad753e658d0cc0df6603edf63b25d21/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxReportSensor.java#L126

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
jmecosta: in this case, i think so. the XunitSensor is feeding test metrics, this is to feed issues so they have different purposes. what we can do is to use the same transformation code, @stefanweiser you could take a look into the xunit sensor how its done move the logic to own class and then reuse in both external sensor and xunit sensor?

guwirth:
@stefanweiser 

unit test: maybe you can use this as an sample: https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-cxx-plugin/src/test/java/org/sonar/plugins/cxx/tests/xunit/CxxXunitSensorTest.java#L79

guwirth:
>> What do you mean with UI-integration? How do I get that?

@stefanweiser on server side you can set the properties normally also in the UI <Administration/C++ (Community)>. These properties are defined here: https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxPlugin.java#L79

Question is if XSLT would be a new tab: (6) XSLT Transformation?

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
stefanweiser: To answer the question of being an extra sensor. This affects sensor execution order: I don't think, that we could enforce a specific order. The documentation do not say something about it. Somebody asked this already, but got not very much information: http://stackoverflow.com/q/42571579/3781684

stefanweiser:
OK, whether it is an own tab or not is a question of "Would it be an own sensor or is it part of a sensor", because in the first case the information is associated to the plugin and is of common nature. In the second case it belongs to the sensor and would be different for every sensor.

Currently I believe, that it should belong as a utility class to each sensor.

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
guwirth: In case we keep it for 'other' name must be  sonar.cxx.other.xslt.xxx. 

Think  it's not hard to change later one. Let's start as part of 'other'. There is also no problem with the sequence then.

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
guwirth: Reasons to catch not only on highest level:
* Could be tool A needs transforming A and tool B needs transformation B. If A is failing B should nevertheless work (tolerant mode)
* Same if more then one file has to be transformed. If transformation A handles more then one file it should continue if one transformation is failing. 

> So you suggest different behavior for both modes? ;-)

Yes behavior of all sensors is different depending on setting of 'sonar.cxx.errorRecoveryEnabled'. But all what you have to do is to call:
https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/utils/CxxUtils.java#L88

guwirth:
@stefanweiser point should be changed before we merge it:
* we keep it as part of other => change keys accordingly 
* support 'tolerant' mode in the right way
* support wildcards
* add an unit test

jmecosta:
@stefanweiser i would also recommend that the xslt in https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/tests/xunit/CxxXunitSensor.java to be moved to a utility library. and be reused in both xunit and external sensor.

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
jmecosta: imo this should be like the https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/tests/xunit/CxxXunitSensor.java  so no big complexity here

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/externalrules/CxxExternalRulesSensor.java
```diff
@@ -87,4 +119,48 @@ public void stream(SMHierarchicCursor rootCursor) throws XMLStreamException {
...
+            for (int j = 0; j < sources.length; j++) {
+              transformFile(stylesheetFile, sources[j], outputs[j]);
+            }
+          }
+        }
+      }
+    }
+  }
+
+  private void transformFile(File stylesheetFile, String source, String output) throws TransformerException {
```
jmecosta: transformation can be done in memory, as the xunit sensor. it should speed up the analysis

stefanweiser:
Sorry, for being quiet for a long time. I was a bit busy. I integrated most of the change requests and got stuck with testing now. How can I debug these tests. I added some test files, but it seems that my test will not generate any output files in the test scenario.

jmecosta:
@stefanweiser try intarget\surefire-reports usually all test data ouput is found under those files

stefanweiser:
Thanks @jmecosta. This worked pretty well.

I've pushed all changes and I hope, that I forgot nothing.

- The keys start now with: sonar.cxx.other.xslt.*
- Added two unit tests (one for invalid files and one for successful transformation)
- I use getReports now, so wildcards are supported on the inputs (on the outputs, there is no way, because it would not be possible to guess the output names)
- Support for the tolerant and strict way

I hope, that you won't find some easter eggs in there.

jmecosta:
I would still try to do the transformation on memory, so the output
properties are really not required.  I think the unit test is done like that

stefanweiser:
But if we would skip the temporaries and output just one XML, you'd have much more to debug if a tranlation script is buggy. How realistic is the case, where multiple transformation steps are necessary?

jmecosta:
ok, but then the default behaviour should be without the outputs. in normal cases a end user would want to set only the xslt?

stefanweiser:
I think the usual way is to set the XSLT, provide one or more input XML(s) and go. The configuration scheme should also change into something like this, if we don't want the intermediate output:

`sonar.cxx.xslt.stylesheets=MyStylesheet1.xslt,MyStylesheet2.xslt,MyStylesheet3.xslt
sonar.cxx.xslt.source=tool1/reports/xxx-1.xml,tool1/reports/xxx-2.xml
sonar.cxx.xslt.output=reports/other-tool1-1.xml,reports/other-tool1-2.xml`

Then there is no need for the different blocks sonar.cxx.xslt.1, 2, 3...

The way it is at the moment, we will have to configure more and get more build artifacts but IMHO it is better for debugging and more straight forward.

jmecosta:
Well it's not a big deal, should be ok then,  try to rebase and squash to
get the build green

stefanweiser:
Now all checks are green.

stefanweiser:
I fixed an issue on key naming (sonar.cxx is getting now prefixed by CxxLanguage), that exists since I merged the upstream branch into my fork.

guwirth:
@stefanweiser thanks for providing this. It's also ok from my side besides minor issues:

* because `source` and `output` are **lists** now should be `sources` and `outputs` (with `s`)
```
sonar.cxx.xslt.1.sources=tool1/reports/xxx-1.xml
sonar.cxx.xslt.1.outputs=reports/other-tool1-1.xml
```
* after seeing the solution wondering if we should not better rename `sources` to `inputs`? Think the pairs are source/destination and input/output?

stefanweiser:
ACK, renamed it in `inputs` and `outputs`.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/other/CxxOtherSensor.java
```diff
@@ -20,33 +20,49 @@
...
+ * @author jorge costa, stefan weiser
  */
 public class CxxOtherSensor extends CxxReportSensor {
   public static final Logger LOG = Loggers.get(CxxOtherSensor.class);
   public static final String REPORT_PATH_KEY = "other.reportPath";
-  public static final String KEY = "Other";
+  public static final String KEY = "other";
+  public static final String OTHER_XSLT_KEY = KEY + ".xslt.";
+  public static final String STYLESHEET_KEY = ".stylesheet";
+  public static final String SOURCE_KEY = ".inputs";
```
guwirth: name should be INPUT_KEY

cxx-sensors/src/main/java/org/sonar/cxx/sensors/other/CxxOtherSensor.java
```diff
@@ -86,9 +111,67 @@ public void stream(SMHierarchicCursor rootCursor) throws XMLStreamException {
...
   @Override
   protected String getSensorKey() {
     return KEY;
   }  
+
+  public void transformFiles(final File baseDir) {
+    Boolean goOn = true;
+    for (int i = 1; (i < 10) && goOn; i++) {
+      String stylesheetKey = OTHER_XSLT_KEY + i + STYLESHEET_KEY;
+      String sourceKey = OTHER_XSLT_KEY + i + SOURCE_KEY;
```
guwirth: name should be inputKey

cxx-sensors/src/main/java/org/sonar/cxx/sensors/other/CxxOtherSensor.java
```diff
@@ -86,9 +111,67 @@ public void stream(SMHierarchicCursor rootCursor) throws XMLStreamException {
...
+      String outputKey = OTHER_XSLT_KEY + i + OUTPUT_KEY;
+
+      String stylesheet = resolveFilename(baseDir.getAbsolutePath(), cxxLanguage.getStringOption(stylesheetKey));
+      List<File> sources = getReports(cxxLanguage, baseDir, sourceKey);
```
guwirth: name should be inputs

cxx-sensors/src/main/java/org/sonar/cxx/sensors/other/CxxOtherSensor.java
```diff
@@ -86,9 +111,67 @@ public void stream(SMHierarchicCursor rootCursor) throws XMLStreamException {
...
+          }
+        }
+      } else {
+          // No keys found.
+          goOn = false;
+      }
+    }
+  }
+
+  private void transformFileList(final String baseDir, File stylesheetFile, List<File> sources, List<String> outputs) {
```
guwirth: sources => inputs

cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxUtils.java
```diff
@@ -70,8 +78,15 @@ public static String normalizePathFull(String filename, String baseDir) {
     }
     return filePath;
   }
-  
-  
+
+
+  public static void transformFile(Source stylesheetFile, File source, File output) throws TransformerException {
```
guwirth: source => input

stefanweiser:
Shame on me. I've just fixed that.

guwirth:
Hi @stefanweiser thanks a lot. Maybe you can do a final rebase & squash before merge. AppVeyor problem seems to be a tool problem.

stefanweiser:
@guwirth: The branch is already rebased. Do you want me to squash all my commit into one?

guwirth:
@stefanweiser yes if possible. This will trigger a new build.

guwirth:
@jmecosta can you restart AppVeyor to see if everything is green please.

jmecosta:
Yep, bit later in the day. Tought

stefanweiser:
Sorry for letting you wait. Squashed the feature into one commit.

guwirth:
@stefanweiser thanks a lot for your contribution and the patience with us 😄.

PR #1099, opened 2017-03-27
move sensors to own library

jmecosta:
separate all sensors into own library, connect the cxx plugin to it.


jmecosta:
@guwirth this is a initial implementation. for now only preparatory work to take the sensors out into a reusable library. once this is ready i will create a dedicated pr for c plugin

jmecosta:
@guwirth this is ready, i think

jmecosta:
massive refactoring, without touching integration tests and everything works :)

guwirth:
@jmecosta Sounds good will have a look...

cxx-sensors/src/main/java/org/sonar/cxx/sensors/squid/CxxSquidSensor.java
```diff
@@ -293,4 +334,11 @@ private int saveViolations(InputFile inputFile, SourceFile squidFile, SensorCont
   public String toString() {
     return getClass().getSimpleName();
   }
+
+  //@Override
+  //public List<Metric> getMetrics() {
```
guwirth: can be removed?

cxx-lint/src/main/java/org/sonar/cxx/cxxlint/CppLanguage.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public class CppLanguage extends CxxLanguage {
+
+  public CppLanguage() {
+    super("c++", new Settings());
+  }  
+  
+  @Override
+  public String[] getFileSuffixes() {
+    return new String [] {"cpp", "hpp", "h", "hxx", "cxx"};
```
guwirth: is getSourceFileSuffixes + getHeaderFileSuffixes?

cxx-sensors/src/test/java/org/sonar/plugins/cxx/squid/CustomCxxRulesDefinitionTest.java
```diff
@@ -48,7 +50,7 @@ public void test() {
     RulesDefinition.Repository repository = context.repository(REPOSITORY_KEY);
 
     assertThat(repository.name()).isEqualTo(REPOSITORY_NAME);
-    assertThat(repository.language()).isEqualTo(CxxLanguage.KEY);
+    assertThat(repository.language()).isEqualTo("c++");
```
guwirth: 'c++' hard coded. Is there no function?

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxPlugin.java
```diff
@@ -167,7 +177,7 @@
   private static List<PropertyDefinition> codeAnalysisProperties() {
     String subcateg = "(2) Code analysis";
     return new ArrayList<>(Arrays.asList(
-      PropertyDefinition.builder(CxxCppCheckSensor.REPORT_PATH_KEY)
+      PropertyDefinition.builder("sonar.cxx." + CxxCppCheckSensor.REPORT_PATH_KEY)
```
guwirth: hard coded string 'sonar.cxx': better use a constant

guwirth:
@jmecosta looking over the code without testing it it looks good.
* Question: Is this backward compatible? Using the new version will users loose their history?
* How will this look like in UI? Two separate sensors C++ plugin and C plugin?

cxx-lint/src/main/java/org/sonar/cxx/cxxlint/CppLanguage.java
jmecosta: yep, but it takes less code to do this. we on the next pr we should clarify a bit better this behaviour

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxPlugin.java
jmecosta: fixed

cxx-sensors/src/main/java/org/sonar/cxx/sensors/squid/CxxSquidSensor.java
jmecosta: fixed

cxx-sensors/src/test/java/org/sonar/plugins/cxx/squid/CustomCxxRulesDefinitionTest.java
jmecosta: fixed

jmecosta:
@guwirth this pr should not introduce any regression. this is meant only as a preparatory work to introduce the c plugin. That will go in once this is merged. 

when we have the c plugin, the only regression ive identified was related with 

"sonar.cxx.suffixes.sources" and headers, basically if we dont change the prop name then all users that use both plugin will need to manually modify their values to remove the c extensions. 

if changing the name of the prop and include only the c++ prefix then you are forcing people to use the 2 plugins. 

For me to reduce regression, i would not change the prop nor the prefix. And mention in documentation that if a c plugin is to be used then you must update the default extensions  in the c++ plugin only to include c++ extensions.

jmecosta:
@guwirth should i merge then?

guwirth:
Yes, would merge. Makes it easier to continue.

PR #1100, opened 2017-04-02
add c plugin

jmecosta:
this plugin introduces a dedicated plugin to analyse c code


jmecosta:
@guwirth this is ready for review, if both plugins are installed then in the admin ui the following must be set

sonar.cxx.suffixes.headers = .hpp,.hh
sonar.cxx.suffixes.sources = .cpp,.cc

this ensures both plugins dont clash like this

> Language of file 'src/component1.c' can not be decided as the file matches patterns of both sonar.lang.patterns.c++ : **/*.cpp,**/*.cc,**/*.c,**/*.hpp,**/*.hh,**/*.h and sonar.lang.patterns.c : **/*.c,**/*.h

jmecosta:
@guwirth is there any reason why appveyour is still using last lts? ive put something in it tests that exists only in 6.3

guwirth:
Decision in the past was that we are always supporting latest LTS. Which will be next LTS?

Von meinem iPhone gesendet

jmecosta:
@guwirth ok, that means i need to test it only on travis. i try to work around it

guwirth:
@jmecosta with Travis I added  the API version to work around this problem with our integration tests:
```
SONARHOME=/tmp/sonarqube-6.3   SONARAPI=SqApi62
```

jmecosta:
@guwirth yes, ive up the tests under that tag. but it fails in something complently unrelated. on the google_test there is exception in analysis related with validaterecovery and im not sure how this only fails in 6.2 and 6.3 and not in all the other versions.  the tests ive added are passing. not sure if i understand those error messages in travis

jmecosta:
@guwirth now its passing, the problem was related with order of the tests. supose since i copy past one project and rename files it seems some db thing happens. ive moved the test to the end and everything is passing now

jmecosta:
@guwirth 

- sonar.cxx.cFilesPatterns i left there so it does not cause any regression to users. Perhaps make it deprecated for next version and in the other one then remove it.
- the rename can be done now, yes. but the all the props and the folder will be different. i dont mind either way

jmecosta:
@guwirth lets merge this?

guwirth:
Hi @jmecosta,

Would still vote to rename cxx plugin to cpp plugin then: folder and jar
Why is c plugin only 7 MB and cxx plugin 15 MB?

Regards

jmecosta:
The linter is not included. If rename is considered then the suffixes can
go also and the extensions prop also change

PR #1111, opened 2017-04-17
add sonar analysis of repository

jmecosta:
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1111)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,9 @@
 version: '{build}'
+environment:
+  token:
+    secure: P6Cj7IcyyNJPTG/5q2W1U4qXhRUQX8BazLcSh9Le8+Czfgmo/F/1VqY1HaMHZddu
+  githubtoken:
+    secure: f72Bkv7Kb0xkwE+CXE9BH9aV8iRIilNM2XANDOffRPazDf0luuu4w2wSQNt+RRZl
 install:
   - ps: |
       Add-Type -AssemblyName System.IO.Compression.FileSystem
@@ -50,6 +55,8 @@ install:
 
 build_script:
   - dir
+  - echo on
+  - echo %APPVEYOR_PULL_REQUEST_NUMBER%
   # SONAR-7154 : workaround
   - mkdir C:\WINDOWS\system32\config\systemprofile\AppData\Local\Temp
   - cd C:\WINDOWS\system32\config\systemprofile\AppData\Local\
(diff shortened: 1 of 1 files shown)


jmecosta:
@guwirth making pr in the main repo also makes both cis to build twice. so 2 much noise, or we need to configure somehting on both envs

guwirth:
> making pr in the main repo also makes both cis to build twice. so 2 much noise, or we need to configure somehting on both envs

@jmecosta 

1. Don't understand `makes both cis to build twice` ? What exactly do you mean?
2. Where can I see the analysis results?

guwirth:
@jmecosta Ok I think I understand now. **branch and pr** are build every time. What I don't understand is how an AppVeyor change could change the Travis behaviour?

jmecosta:
@guwirth because this happens in main sonar-cxx repo. normally we create pr from are own forks, in that case only pr is build. if in the main, then both pr and branch are build. It should be configurable, to ignore all branches except master. In appveyour it can be done, ive tryed it but for some reason it did not work as expected. 

the analysis, if there are issues, will ne inline. In this pr, in the checks you can see: sonarqube — SonarQube reported no issues

So it works, but not for our current workflow, mostly becuase of security reasons in Appveyour. As the security token is available to eveyrone. Someone can just get the token by printing it to console by creating a pull request in fork.  well probably not printing, since the ci mask it, but they can for example post it to a external server and catpure the token. With the token they can do a few things with it, that are consider unsafe. Altough in our case, doubt that it would be the case.

jmecosta:
@guwirth this unfortunately was not meant to be merged.all pull requests that are made from forks will fail since they don't have access to token.

If we want this we need to make pulls request from the repo and nor from forks

guwirth:
@jmecosta found description on Travis: https://docs.travis-ci.com/user/sonarqube/. They are also supporting master only.

jmecosta:
At least it's in the list of upcoming improvements,  so maybe we just wait
until travis support this and the enable the analysis.

Meanwhile could try to enable the analysis only for master. And upload to
SonarQube.com perhaps this time it works

guwirth:
For master is better than nothing.

> Am 29.04.2017 um 11:33 schrieb Jorge Costa <notifications@github.com>:
> 
> At least it's in the list of upcoming improvements, so maybe we just wait
> until travis support this and the enable the analysis.
> 
> Meanwhile could try to enable the analysis only for master. And upload to
> SonarQube.com perhaps this time it works
>

PR #1114, opened 2017-04-18
fix sonarlint issues and reduce false positives

Bertk:
- reduce SonarQube issues (potential bugs)
- mark some issues as false positives in code

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1114)


cxx-sensors/src/main/java/org/sonar/cxx/sensors/drmemory/CxxDrMemorySensor.java
```diff
@@ -85,15 +91,6 @@ protected void processReport(final SensorContext context, File report) {
           }
         }
       }
-    } catch (IOException e) {
```
guwirth: @Bertk why is this removed? Should break or continue (handling inside validateRecovery).

guwirth:
@Bertk can you have a look to the comment above please.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/drmemory/CxxDrMemorySensor.java
Bertk: The only IO operations in this scope are handled with a try..catch block in ```org.sonar.cxx.sensors.utils.normalizePath(String filename)``` and ```getElements(file, charset)```.

I have also some doubts with the validateRecovery in this context. Whenever a report cannot be parsed (IOException) the analysis should not be stopped. This is the current behavior for other tools with reliability issues like Vera++, PC-lint and RATS.

guwirth:
@Bertk yes you are right.

PR #1153, opened 2017-06-05
fix quality flaws

Bertk:
fix bugs, vulnerabilities

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1153)

### DIFF
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportSensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportSensor.java
@@ -269,7 +269,7 @@ public abstract class CxxReportSensor implements Sensor {
             newIssue.at(location);
             newIssue.save();
             violationsCount++;
-          } catch (Exception ex) {
+          } catch (RuntimeException ex) {
             LOG.error("Could not add the issue '{}', skipping issue", ex.getMessage());
             CxxUtils.validateRecovery(ex, this.language);
           }
@@ -280,7 +280,7 @@ public abstract class CxxReportSensor implements Sensor {
       }
     } else {
       // project level
-//      try {
+      try {
         NewIssue newIssue = sensorContext.newIssue().forRule(
                                  RuleKey.of(ruleRepoKey + this.language.getRepositorySuffix(), ruleId));
         NewIssueLocation location = newIssue.newLocation()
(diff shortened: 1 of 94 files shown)


Bertk:
@jmecosta Please check CxxCoverageSensor.java. I removed some catch blocks (still available as comment) because I was not sure whether this is still necessary and I cannot test the coverage on large source code projects.

guwirth:
@Bertk this try/catch block are to continue after coverage errors. @jmecosta added them to switch the behavior on/off. In case the function in the try block can throw an exception we have to keep them (my opinion).

Bertk:
@guwirth That's OK but I could not find a throw in the builder (V6.x API) but I only checked java doc.
http://javadocs.sonarsource.org/latest/apidocs/org/sonar/api/batch/sensor/coverage/NewCoverage.html

SonarQube V6.2 merges the coverage data from multiple sources which is not done with the LTS version and maybe an exception is thrown. Nevertheless we should not use a catches like this 
```
...
catch(Exception ex) {
              ...
} 
```
The java doc for the LTS does only mention exceptions for ```public static CoverageType valueOf(String name)``` but this is not used.
see http://javadocs.sonarsource.org/5.6.6/apidocs/org/sonar/api/batch/sensor/coverage/NewCoverage.html
```
Throws:
IllegalArgumentException - if this enum type has no constant with the specified name
NullPointerException - if the argument is null
```

guwirth:
@Bertk for history see #959 and #842

Bertk:
@guwirth Thanks for the hints. I enabled the exception handling again but used ```RuntimeException```.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportSensor.java
```diff
@@ -247,20 +278,23 @@ private void saveViolation(SensorContext sensorContext, String ruleRepoKey,
           notFoundFiles.add(normalPath);
         }
       }
-    } else { // project level
-      try {
-        NewIssue newIssue = sensorContext.newIssue().forRule(RuleKey.of(ruleRepoKey + this.language.getRepositorySuffix(), ruleId));
+    } else {
+      // project level
+//      try {
+        NewIssue newIssue = sensorContext.newIssue().forRule(
```
guwirth: @Bertk missing try/catch

Bertk: I added the catch for RuntimeException.

guwirth:
@jmecosta do you also have time to review it? Otherwise will merge it latest on Sunday evening.

guwirth:
@Bertk thx for cleaning things up

Issue #1094, opened 2017-03-19
syntax highlighting error:  concatenated strings


guwirth:
Get a syntax highlighting error with the sample below:

```C++
#define T_STRAT_BRAC        "["
#define T_END_BRAC          "]"
#define T_ST_NOT_SUP    T_STRAT_BRAC"START_NOT_SUPPORTED"T_END_BRAC

void test()
{
    fileOutput << T_ST_NOT_SUP << endl;
}
```
The sonar-scanner analysis succeeds but server logs from background tasks mention:
```
[o.s.s.c.s.HighlightingLineReader] Inconsistency detected in Highlighting data. Highlighting will be ignored for file 'com.org.XXX:XXX:OSS/Sources/sample.cpp'
org.sonar.server.computation.source.RangeOffsetConverter$RangeOffsetConverterException: End offset 40 is defined outside the length (39) of the line 1392
        at org.sonar.server.computation.source.RangeOffsetConverter.checkExpression(RangeOffsetConverter.java:65) ~[sonar-server-5.6.4.jar:na]
        at org.sonar.server.computation.source.RangeOffsetConverter.validateEndOffsetNotGreaterThanLineLength(RangeOffsetConverter.java:59) ~[sonar-server-5.6.4.jar:na]
        at org.sonar.server.computation.source.RangeOffsetConverter.offsetToString(RangeOffsetConverter.java:34) ~[sonar-server-5.6.4.jar:na]
        at org.sonar.server.computation.source.HighlightingLineReader.processHighlighting(HighlightingLineReader.java:104) ~[sonar-server-5.6.4.jar:na]
        at org.sonar.server.computation.source.HighlightingLineReader.processHighlightings(HighlightingLineReader.java:91) ~[sonar-server-5.6.4.jar:na]
        at org.sonar.server.computation.source.HighlightingLineReader.read(HighlightingLineReader.java:78) ~[sonar-server-5.6.4.jar:na]
        at org.sonar.server.computation.source.ComputeFileSourceData.read(ComputeFileSourceData.java:61) [sonar-server-5.6.4.jar:na]
```

guwirth:
My first feeling is that `isGeneratedCode` flag is missing?

jmecosta:
what flag is that isGeneratedCode, just curious as i have no idea what going on there.

guwirth:
@jmecosta 

Sample in preprocessor:
[...]

PR #1163, opened 2017-06-15
Preserve generated status in concatenated strings

tomassdepakin:
Concatenated strings doesn't preserve generated status, thus macro concatenations are considered as StringLiteralDuplicates.

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1163)


guwirth:
Should also solve #1094?

guwirth:
@tomassdepakin thanks for you fix.

PR #1166, opened 2017-06-19
Trailing comma is not removed in complex empty variadic macro

tomassdepakin:
Example:
```
#define MACRO(s, ...) do { printf(s, __VA_ARGS__); } while (false)
int main() { MACRO(\"error\"); }
```
Parser doesn't remove comma on empty \_\_VA_ARGS__

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1166)


tomassdepakin:
shouldn't have rushed with pull request, need more testing. created an issue #1167

PR #1182, opened 2017-06-29
Fixes syntax error in list initialization with template

tomassdepakin:
Fixing syntax error on having template inside list initializer
Test example:
```c++
#include <string>


template <typename T>
std::string* string_creator()
{
    return new T();
}

int main()
{
  static const struct
  {
    std::string* (*create)();
  }
  strings[] =
  {
    &string_creator<std::string>
  };
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1182)

### DIFF
--- a/cxx-squid/src/test/resources/parser/own/C++11/list-initialization.cc
+++ b/cxx-squid/src/test/resources/parser/own/C++11/list-initialization.cc
@@ -13,7 +13,7 @@ Object anArray[] = {{13.4f, 3}, {43.28f, 29}, {5.934f, 17}}; // An array of thre
 
 class SequenceClass {
 public:
-    SequenceClass(std::initializer_list<int> list);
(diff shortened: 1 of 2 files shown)


cxx-squid/src/test/resources/parser/own/C++11/list-initialization.cc
```diff
@@ -13,7 +13,7 @@ Object anArray[] = {{13.4f, 3}, {43.28f, 29}, {5.934f, 17}}; // An array of thre
 
 class SequenceClass {
 public:
-    SequenceClass(std::initializer_list<int> list);
+    SequenceClass(std::initializer_list<int> list){};
```
tomassdepakin: Yep, my change is working.
It's just to fix compilation of this file by real compiler.

PR #1188, opened 2017-07-14
Updating GCC rules according to gcc/g++ documentation V6.3

### LINKED ISSUE #1173: Update gcc rules
INFO: Parsing 'GCC' format
ERROR: Could not add the issue 'The rule 'compiler-gcc:-Wliteral-suffix' does not exist.', skipping issue
ERROR: Could not add the issue 'The rule 'compiler-gcc:-Wliteral-suffix' does not exist.', skipping issue

gena01:
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1188)

### DIFF
--- a/cxx-sensors/src/main/resources/compiler-gcc.xml
+++ b/cxx-sensors/src/main/resources/compiler-gcc.xml
@@ -20,7 +20,7 @@ Follow these steps to make your custom Custom rules available in SonarQube:
       ]]>
     </description>
     <severity>MAJOR</severity>
-  </rule>    
+  </rule>
   <rule>
     <key>enabled by default</key>
     <name>Default compiler warnings</name>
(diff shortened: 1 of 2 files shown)


guwirth:
@gena01 thanks for your contribution. Think you have to fix unit test CxxCompilerRuleRepositoryTest.createGccRulesTest to make CI green. 

Other question: which is the gcc version you are using?

gena01:
@guwirth The rules i copied from the official GCC documentation.

As far as the gcc versions that I am using are : 4.8.3 5.x (need to check the exact one) 6.4.0 (mostly testing for now)

guwirth:
@gena01 reason why I'm asking this: what should we write to the documentation: compatible with gcc ???.

gena01:
@guwirth I based it on the current gcc/g++ documentation. I would say it's safe to say 6.3

PR #1194, opened 2017-07-28
test with SQ 6.5 RC2

guwirth:
With SQ6.5 it is no more possible to change default profiles (https://jira.sonarsource.com/browse/SONAR-9303). For the integration tests we therefore create a copy of the default C++ profile and use this as new default for the projects.

https://groups.google.com/forum/#!topic/sonarqube/1ThRpLMAw1s

This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1194)


gena01:
@guwirth Looks like 6.5 final is out.

Issue #1199, opened 2017-08-08
Relative path causes NPE


yachoor:
After change in https://github.com/SonarOpenCommunity/sonar-cxx/commit/ef20a95ff34b665397f7da9135f7c00d1c8db3a9#diff-12a8ec1d2a533c7305380c2611271cc4R254 I get java.lang.NullPointerException in org.sonar.cxx.sensors.utils.CxxUtils.normalizePathFull(CxxUtils.java:74), called from org.sonar.cxx.sensors.utils.CxxReportSensor.saveViolation(CxxReportSensor.java:265) for filename with relative path ("../../src/hardware/ISerialPort.cpp")

Also you might want to improve logging exceptions in CxxReportSensor - "Cannot feed the data into sonar, details: 'java.lang.NullPointerException'" isn't really helpful without stack trace.

yachoor:
Sorry, latest versions: sonar-cxx 0.9.8-SNAPSHOT and SonarQube 6.5

The problem is that PathUtils.sanitize() returns null for relative path filename here: https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportSensor.java#L265

Stack trace:
```
java.lang.NullPointerException: null
        at org.sonar.cxx.sensors.utils.CxxUtils.normalizePathFull(CxxUtils.java:74)
        at org.sonar.cxx.sensors.utils.CxxReportSensor.saveViolation(CxxReportSensor.java:265)
        at org.sonar.cxx.sensors.utils.CxxReportSensor.saveUniqueViolation(CxxReportSensor.java:244)
        at org.sonar.cxx.sensors.compiler.CxxCompilerSensor.processReport(CxxCompilerSensor.java:128)
        at org.sonar.cxx.sensors.utils.CxxReportSensor.executeReport(CxxReportSensor.java:116)
        at org.sonar.cxx.sensors.utils.CxxReportSensor.execute(CxxReportSensor.java:87)
        at org.sonar.scanner.sensor.SensorWrapper.analyse(SensorWrapper.java:53)
        at org.sonar.scanner.phases.SensorsExecutor.executeSensor(SensorsExecutor.java:88)
        at org.sonar.scanner.phases.SensorsExecutor.execute(SensorsExecutor.java:82)
        at org.sonar.scanner.phases.SensorsExecutor.execute(SensorsExecutor.java:68)
[...]

PR #1200, opened 2017-08-08
do not use PathUtils.sanitize(filename)

Bertk:
cannot handle relative path and returns null
#1199

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1200)


guwirth:
@Bertk thx for your fast response.

Issue #1203, opened 2017-08-09
improve EXECUTABLE_LINES_DATA_KEY metrics support


Bertk:
CxxFileLinesVisitor should only consider source code lines from **CxxGrammarImpl.functionBody** for EXECUTABLE_LINES_DATA_KEY (Set<Integer> executableLines).
The current implementation also adds **functionDefinition**, **functionDeclSpecifierSeq**, and **declarator** AST node for executable code lines. 

Therefore the calculation of code coverage (%) is not correct.

guwirth:
Hi @Bertk please have a look to the graphics at https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Get-code-coverage-metrics. Seems that at least Bullseye provide also for functionDefinition, functionDeclSpecifierSeq, and declarator coverage information? One reason why VS coverage does typically not match to the numbers using VisualStudio is the "partial" coverage.

guwirth:
Hi @Bertk see a sample in our code base with 100% coverage. First/last bracket from functionBody are covered. Think there is the code to create/destroy local variables.

![2017-08-10_16h55_47](https://user-images.githubusercontent.com/6077367/29176764-12b04a08-7ded-11e7-87de-57d2f9645439.png)

Bertk:
@guwirth Hi, thank you for the hints. Can you please share with me which coverage tool was used.
Here is a snippet from an analysis with **SQ 6.5** and **bullseye** :confused:
![bullseye](https://user-images.githubusercontent.com/2315215/29219021-1463335e-7eb7-11e7-8140-af6a70d59663.png)
The coverage sensor need an enhancement like it is done in the sonar [java plug-in FileLinesVisitor.java](https://github.com/SonarSource/sonar-java/blob/master/java-frontend/src/main/java/org/sonar/java/ast/visitors/FileLinesVisitor.java) and calculate the correct values for executable lines based on the AST.

guwirth:
@Bertk I agree. We have to handle
* CoreMetrics.NCLOC_DATA_KEY
* CoreMetrics.COMMENT_LINES_DATA_KEY,
* CoreMetrics.EXECUTABLE_LINES_DATA_KEY,

different.

The only thing I'm not really sure is what are "executable lines":
* Are these only lines within a function body?
* Or also the functionDefinition: because at the beginning/end there is code to create/destroy local variables?
* What's with global variables in case they have a Ctor?
* What's with initializer lists?
* ...

[...]

PR #1206, opened 2017-08-13
improve executable lines code metric

Bertk:
- use AST to identify executable lines (SQ 6.2 and newer versions)
- still experimental (some AST nodes of CxxGrammarImpl might be missing)
#1203

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1206)


guwirth:
Hi @Bertk,

I played around with our code base and found the following points:

* **case/default are not marked:**
coverage3

* **if / else: else is not marked**
coverage2

* **try / catch: catch not marked:**
coverage7

* **differences old / new:**
old:
coverage6
new:
coverage7
old:
coverage5
new:
coverage4

Wondering if { } should not be ignored in general. Is not really 'executable code'?

Regards,

guwirth:
Hi @Bertk,

second proposal for an improvement. Like the construct with **forEach** to set the metric like here:
https://github.com/SonarSource/sonar-javascript/blob/7df6fc71b6f481dce2739843db3e2e8e5a908efc/javascript-frontend/src/main/java/org/sonar/javascript/metrics/MetricsVisitor.java#L173

Regards,

guwirth:
@Bertk 
Also some hints I found:
* Things that are executable: https://docs.sonarqube.org/display/DEV/Executable+Lines
* [MMF-739](https://jira.sonarsource.com/browse/MMF-739)
* [EXECUTABLE_LINES_DATA](http://javadocs.sonarsource.org/latest/apidocs/index.html?org/sonar/api/measures/CoreMetrics.html): Information about executable lines of code in file. Key-value pairs, where key - is a number of line, and value - is an indicator of whether line contains executable code (1) or not (0). If a line number is missing in the map it is equivalent to the default value (0).

guwirth:
@Bertk ... and by the way: thanks a lot for spending your time!

Bertk:
@guwirth I am still waiting on an analysis on our code base. Thanks for the hints - the rule only to consider the first line of a multi-line statement is interesting and make the solution more complex. I check this details on the weekend.
The handling of punctuation seems to be different with the Microsoft VS way ("{","}"). I want to have the same UI experience for the developers as they know from their preferred IDE and maybe a configuration item could help.

guwirth:
Hello @Bertk,

would extend the unit test with the file https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/test/resources/org/sonar/cxx/sensors/ncloc.cc with the following cases:

```C++
void func7(bool flag)
{
    int iii = 0;

    if (flag) {
        iii = 1;
    }
    else {
        iii = 0;
    }
}

int func8(int p)
{
    int r;

    switch (p) {
    case 0:
        r = 100;
        break;
    case 1:
        r = 200;
        break;
    default:
        r = 300;
        break;
    }

    return r;
}

void func9()
{
    int *p;

    try {
        p = new int[10];
    }
    catch (...) {
        p = nullptr;
    }

}

void func10()
{
    // comment
    for (int iii = 0;
        iii < NUMBER;
        ++iii)
    {
        h1 += iii; // comment
    }
}
```

Than just another idea: Is it not possible to handle only `CxxGrammarImpl.statement`? And mark only lines containing a statement as 'executable'?

guwirth:
Hi @Bertk,

below the coverage with VS2015.3 with sample file as it is shown inside of VS:

vs1
grafik
grafik
grafik
grafik

**Try / catch:**
grafik
Throwing an exception:
grafik

**if/else:**
if
grafik
else
grafik

Regards

Bertk:
@guwirth I think the coverage tools handle the line, statement, branch coverage differently and a better solution might consider this behavior. Microsoft coverage tool and Bullseye ccover are used by us but until now the result from the tools and SonarQube are different  :confused:
By the way, what about **_[constexpr](http://en.cppreference.com/w/cpp/language/constexpr)_** code coverage? At the moment the block is assigned to executable lines ```7, 8, 9, 17, 22, 23, 24, 25, 32, 33, 36, 42, 46, 47, 48, 50, 51, 53, 54, 55```

constexpr

I think for now the current solution should be OK and refactoring should be postponed. Maybe the [current coverage approach for executable lines](https://docs.sonarqube.org/display/DEV/Executable+Lines) is too simple and will be updated again.

see also [Bullseye: What is Wrong with Statement Coverage](http://bullseye.com/statementCoverage.html#b2)

guwirth:
@Bertk thanks for your update.

Comparing the executable lines in the unit test with the sample code I would still propose two changes:
* Don't  mark  {} as executable.  Think not all coverage tool will mark it as covered. It's then a false positive left. In case of VS coverage report they will anyway be overwritten and green. 
* **else** should also be not marked as executable

guwirth:
@Bertk **constexpr** looks like this in VS:

grafik

So no special handling needed.

guwirth:
Would merge it as is.

PR #1236, opened 2017-09-21
Fixes 'hardcoded ip' rule false positive on macros

tomassdepakin:
Hi, we got 'Hardcode IP' rule triggered on macros.

```
#include <string>

#define VERSION_STR "1.2.3.4"

using namespace std;

class A {
    const string version = VERSION_STR; // Compliant (generated code)
};

int main() {
    return 0;
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1236)

### DIFF
--- a/cxx-checks/src/main/java/org/sonar/cxx/checks/HardcodedIpCheck.java
+++ b/cxx-checks/src/main/java/org/sonar/cxx/checks/HardcodedIpCheck.java
@@ -79,7 +79,7 @@ public class HardcodedIpCheck extends SquidCheck<Grammar> {
 
   @Override
   public void visitNode(AstNode node) {
-    if (node.is(CxxGrammarImpl.LITERAL)) {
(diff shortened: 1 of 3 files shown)


guwirth:
@tomassdepakin thanks for your PR.

Code looks good but in general I'm not sure if hiding an IP address behind a macro is 'compliant'? Was the implementation before a bug or a feature?

tomassdepakin:
> I'm not sure if hiding an IP address behind a macro is 'compliant'?
I'm not sure either. It is less 'hardcoded' than direct ip in code, yet it is definitely worse than moving it to configuration option. Your question made me think that it is better to leave it as is and decline my request.
In our case it is triggered on version string macro (like '12.0.5.2'), which is a false positive. We have only 2 false positives and can mark 'em at SQ.

PR #1245, opened 2017-10-07
Initial implementation of cognitive complexity visitor and function c…

### LINKED ISSUE #1006: Add Cognitive Complexity
**Abstract**
Cyclomatic Complexity was initially formulated as a measurement of the “testability and
maintainability” of the control flow of a module. While it excels at measuring the former, its
underlying mathematical model is unsatisfactory at producing a value that measures the
latter. This white paper describes a new metric that breaks from the use of mathematical
models to evaluate code in order to remedy Cyclomatic Complexity’s shortcomings and
produce a measurement that more accurately reflects the relative difficulty of understanding,
and therefore of maintaining methods, classes, and applications.

source: https://www.sonarsource.com/docs/CognitiveComplexity.pdf

nathanaelg:
…heck based on https://www.sonarsource.com/docs/CognitiveComplexity.pdf

Recursion is currently not given a complexity increment.

close #1006 

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1245)


nathanaelg:
So here's what I've done so far - the main thing that I know of that diverges from the cognitive complexity whitepaper is I was entirely sure what they meant by:

> Recursion
Unlike Cyclomatic Complexity, Cognitive Complexity adds a fundamental increment for each
method in a recursion cycle, whether direct or indirect. There are two motivations for this
decision. First, recursion represents a kind of “meta-loop”, and Cognitive Complexity
increments for loops. Second, Cognitive Complexity is about estimating the relative difficulty
of understanding the control flow of a method, and even some seasoned programmers find
recursion difficult to understand. 

In something like this for example:
```c
int fibonacci(int n)  {
   if (n == 0) {
      return 0;
   }
   else if (n == 1) {
      return 1;
   }
   else {
      return fibonacci(n - 1) + fibonacci(n - 2);
   }
}
```
I'm not sure whether the return fibonacci would gain +1, or perhaps +2 because there are two recursive calls, or something completely different again?

guwirth:
@nathanaelg thanks a lot for providing this PR. Code looks good but I need some more time to validate the samples.

guwirth:
@nathanaelg 

> I'm not sure whether the return fibonacci would gain +1, or perhaps +2 because there are two recursive calls, or something completely different again?

Would count each function call to itself as +1,  like a while/goto.

cxx-checks/src/main/java/org/sonar/cxx/checks/FunctionCognitiveComplexityCheck.java
```diff
@@ -0,0 +1,77 @@
...
+  tags = {Tag.BRAIN_OVERLOAD},
+  priority = Priority.MAJOR)
+@ActivatedByDefault
+@SqaleLinearWithOffsetRemediation(
+  coeff = "1min",
+  offset = "10min",
+  effortToFixDescription = "per complexity point above the threshold")
+public class FunctionCognitiveComplexityCheck extends SquidCheck<Grammar> {
+
+  private static final int DEFAULT_MAX = 10;
```
guwirth: Default in other languages (Java and JS) is 15: https://github.com/SonarSource/SonarJS/blob/master/javascript-checks/src/main/java/org/sonar/javascript/checks/CognitiveComplexityFunctionCheck.java#L46

cxx-checks/src/main/java/org/sonar/cxx/checks/FunctionCognitiveComplexityCheck.java
```diff
@@ -0,0 +1,77 @@
...
+import org.sonar.squidbridge.checks.ChecksHelper;
+import org.sonar.squidbridge.checks.SquidCheck;
+import com.sonar.sslr.api.AstNode;
+import com.sonar.sslr.api.Grammar;
+import org.sonar.squidbridge.annotations.ActivatedByDefault;
+import org.sonar.squidbridge.annotations.SqaleLinearWithOffsetRemediation;
+import org.sonar.cxx.tag.Tag;
+
+@Rule(
+  key = "FunctionCognitiveComplexity",
```
guwirth: You should add an description, like: https://github.com/SonarSource/SonarJS/blob/149e36eb12db8af81cb538e09f555639dcb3a691/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S3776.html

Bertk:
Very nice work 👍 +1
You need to add `FunctionCognitiveComplexityCheck.class` to the list of checks for the CXX language (sonar-cxx-plugin\src\main\java\org\sonar\plugins\cxx\CppLanguage.java) and increase the number of checks in the unit test (sonar-cxx-plugin\src\test\java\org\sonar\plugins\cxx\CxxCheckListTest.java, sonar-cxx-plugin\src\test\java\org\sonar\plugins\cxx\CxxRuleRepositoryTest.java).
After this updates the build will fail:
```
rulesTest(org.sonar.plugins.cxx.CxxRuleRepositoryTest)  Time elapsed: 0 sec  <<< ERROR!
java.lang.IllegalStateException: One of HTML description or Markdown description must be defined for rule [repository=cxx, key=FunctionCognitiveComplexity]
        at org.sonar.api.server.rule.RulesDefinition$NewRule.validate(RulesDefinition.java:891)
        at org.sonar.api.server.rule.RulesDefinition$NewRule.access$800(RulesDefinition.java:659)
        at org.sonar.api.server.rule.RulesDefinition$RepositoryImpl.<init>(RulesDefinition.java:566)
        at org.sonar.api.server.rule.RulesDefinition$RepositoryImpl.<init>(RulesDefinition.java:540)
        at org.sonar.api.server.rule.RulesDefinition$Context.registerRepository(RulesDefinition.java:436)
        at org.sonar.api.server.rule.RulesDefinition$Context.access$500(RulesDefinition.java:380)
        at org.sonar.api.server.rule.RulesDefinition$NewRepositoryImpl.done(RulesDefinition.java:512)
        at org.sonar.plugins.cxx.CxxRuleRepository.define(CxxRuleRepository.java:41)
        at org.sonar.plugins.cxx.CxxRuleRepositoryTest.rulesTest(CxxRuleRepositoryTest.java:34)
```
The description for the check has to added here: **_sonar-cxx-plugin/src/main/resources/org/sonar/l10n/c++/rules/cxx/FunctionCognitiveComplexity.html_**

Bertk:
Sorry - I accidently closed this PR.

cxx-squid/src/test/resources/visitors/binary_logical.cc
```diff
@@ -0,0 +1,7 @@
+int main(int argc, char* argv[])
+{
+   if ((argc == 1) &&         // +1
+       (argv[0] != NULL)) {   // +1
```
guwirth: Not sure if this should not be 1 only? Maybe you can add also the other samples from PDF section "Sequences of logical operators" to verify it.

cxx-checks/src/main/java/org/sonar/cxx/checks/FunctionCognitiveComplexityCheck.java
nathanaelg: Fixed.

cxx-checks/src/main/java/org/sonar/cxx/checks/FunctionCognitiveComplexityCheck.java
nathanaelg: Added sonar-cxx-plugin/src/main/resources/org/sonar/l10n/c++/rules/cxx/FunctionCognitiveComplexity.html and sonar-c-plugin/src/main/resources/org/sonar/l10n/c/rules/c/FunctionCognitiveComplexity.html (not sure whether we wanted both but I presume this will work with C as well as C++)

cxx-squid/src/test/resources/visitors/binary_logical.cc
nathanaelg: Good catch, this example is actually correct (though not so helpfully annotated, should be +1 for if, and +1 for &&), but I hadn't read the 'sequence of logical operators' section so the other samples were incorrect.

nathanaelg:
I've been struggling with the recursion - my thinking was to track the name of the current function and then increment whenever that was found within the function but I haven't been able to work out exactly the right grammar to use? `CxxGrammarImpl.declarator` possibly seemed to be the right thing for the name of the function (though it seemed the name of parameters passed into the function were also declarators?) and I couldn't work out what the function call (to itself) would be.

guwirth:
@nathanaelg recursion:  think easiest is to copy sample code into sslr-cxx-toolkit to identify right grammar part.

guwirth:
@nathanaelg thanks for the updates. Will have on Saturday a closer look to it.

guwirth:
@Bertk also ok for you to merge?

cxx-squid/src/test/java/org/sonar/cxx/visitors/CxxCognitiveComplexityVisitorTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+  private void testFile(String fileName, int expectedCognitiveComplexity) throws UnsupportedEncodingException, IOException {
+    CxxCognitiveComplexityVisitor<Grammar> visitor = CxxCognitiveComplexityVisitor.<Grammar>builder()
+      .setMetricDef(CxxMetric.COGNITIVE_COMPLEXITY)
+      .build();
+
+    CxxFileTester tester = CxxFileTesterHelper.CreateCxxFileTester(fileName, ".");
+    SourceFile sourceFile = CxxAstScanner.scanSingleFile(tester.cxxFile, tester.sensorContext, CxxFileTesterHelper.mockCxxLanguage(), visitor);
+
+    assertThat(sourceFile.getInt(CxxMetric.COGNITIVE_COMPLEXITY)).isEqualTo(expectedCognitiveComplexity);
```
Bertk: Assert should be implemented in tests and not in helper method. This is a minor issue and could be improved in another PR e.g. private int testFile(String fileName) returns CxxMetric.COGNITIVE_COMPLEXITY

nathanaelg: I was just about to change this to go with another PR, but you've already fixed it, thanks!

PR #1262, opened 2017-10-21
Clang Static Analyzer rules xml generator script

gyorb:
I've created a script wich can be used to generate the rules xml file for the Clang Static Analyzer.

Only a small manual change is needed at the header after the xml is generated (adding a comment and encoding).

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1262)


gyorb:
@guwirth I will update it.

gyorb:
I've made some updates to the script. I've added description and help to the script.
The clang version which is used to get the checkers list is also automatically added to the rules xml comment section.
There is a small change in the generated xml file but with this change no manual editing is required.

guwirth:
Hi @gyorb thanks a lot for your improvements. Should be easy to update the rules in future now.

Issue #1277, opened 2017-11-03
ERROR: sonar.cxx.other.xslt.1.inputs file is not defined.


guwirth:
Hi,

Get the following message testing 0.9.8 with SQ 5.6.4:

```
INFO: Sensor c++ ExternalRulesSensor
INFO: Undefined report path value for key 'sonar.cxx.other.xslt.1.inputs'
ERROR: sonar.cxx.other.xslt.1.inputs file is not defined.
```

XSLT is optional, there should be no message?

Regards,

PR #1283, opened 2017-11-04
suppress error message (input file is not available)

Bertk:
#1277

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1283)


Bertk:
@guwirth OK - this solution is better than a simple bug fix. Now this is an improvement ant needs additional unit tests.

PR #1348, opened 2017-11-30
Fix Issue1277: ERROR: sonar.cxx.other.xslt.1.inputs file is not defined.

rufinio:
XSLT is optional, there should be no message.

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1348)

### DIFF
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/other/CxxOtherSensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/other/CxxOtherSensor.java
@@ -138,7 +138,7 @@ public class CxxOtherSensor extends CxxReportSensor {
       }
       List<String> outputs = Arrays.asList((outputStrings != null) ? outputStrings : new String[] {});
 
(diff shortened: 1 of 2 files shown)


guwirth:
Hello @rufinio, can you please explain the use case behind this. In general it would be good to have a description what this xslt feature is doing.

rufinio:
Hi @guwirth,
I am also curious what's the use case behind this xslt feature.
I have updated sonar-cxx plugin to 0.9.8 and got "ERROR: sonar.cxx.other.xslt.1.inputs file is not defined." without using this feature.
I found the closed issue #1277 that already dealt with this but obviously it was not solved.

guwirth:
@rufinio :-)  the original idea was #1096. Idea is to have a built-in preceding XSLT transformation for the `sonar.cxx.other.` sensor. Therefore I was wondering what you are doing with it without assigning an XSLT? Setting `inputs` and `outputs` without an `xslt` makes no sense to me.

But you are right that `sonar.cxx.other.` should work without  `inputs`, `outputs`  and `xslt`.

PR #1398, opened 2018-01-19
Add function complexity and size metrics

ericlemes:
Contains "Complex Functions" metric, which counts number of function that has cyclomatic complexity greater than sonar.cxx.funccomplexity.threshold (default 10). "% of Complex Functions" is the same metric expressed as a percent ((complex functions * 100)/total functions).

"Big functions", counts number of functions with more than sonar.cxx.funcsize.threshold (default 20). "% of Big Functions" is the same expressed as a percent.

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1398)


Bertk:
@ericlemes Hi, please just create another commit and sync it or close obsolete PRs.

SonarQube server fails to start:
```
2018.01.19 14:46:03 INFO  web[][o.s.s.s.GeneratePluginIndex] Generate scanner plugin index
2018.01.19 14:46:04 INFO  web[][o.s.s.s.RegisterPlugins] Register plugins
2018.01.19 14:46:04 ERROR web[][o.s.s.p.Platform] Background initialization failed. Stopping SonarQube
java.lang.IllegalStateException: Metric [complex_functions] is already defined by the repository [org.sonar.cxx.sensors.functioncomplexity.FunctionComplexityMetrics@1713eca4]
	at org.sonar.server.startup.RegisterMetrics.checkMetrics(RegisterMetrics.java:127)
	at org.sonar.server.startup.RegisterMetrics.getPluginMetrics(RegisterMetrics.java:112)
	at org.sonar.server.startup.RegisterMetrics.start(RegisterMetrics.java:63)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.picocontainer.lifecycle.ReflectionLifecycleStrategy.invokeMethod(ReflectionLifecycleStrategy.java:110)
	at org.picocontainer.lifecycle.ReflectionLifecycleStrategy.start(ReflectionLifecycleStrategy.java:89)
	at org.picocontainer.injectors.AbstractInjectionFactory$LifecycleAdapter.start(AbstractInjectionFactory.java:84)
	at org.picocontainer.behaviors.AbstractBehavior.start(AbstractBehavior.java:169)
	at org.picocontainer.behaviors.Stored$RealComponentLifecycle.start(Stored.java:132)
	at org.picocontainer.behaviors.Stored.start(Stored.java:110)
	at org.picocontainer.DefaultPicoContainer.potentiallyStartAdapter(DefaultPicoContainer.java:1016)
	at org.picocontainer.DefaultPicoContainer.startAdapters(DefaultPicoContainer.java:1009)
	at org.picocontainer.DefaultPicoContainer.start(DefaultPicoContainer.java:767)
	at org.sonar.core.platform.ComponentContainer.startComponents(ComponentContainer.java:134)
	at org.sonar.server.platform.platformlevel.PlatformLevel.start(PlatformLevel.java:90)
	at org.sonar.server.platform.platformlevel.PlatformLevelStartup.access$001(PlatformLevelStartup.java:45)
	at org.sonar.server.platform.platformlevel.PlatformLevelStartup$1.doPrivileged(PlatformLevelStartup.java:83)
	at org.sonar.server.user.DoPrivileged.execute(DoPrivileged.java:45)
	at org.sonar.server.platform.platformlevel.PlatformLevelStartup.start(PlatformLevelStartup.java:80)
	at org.sonar.server.platform.Platform.executeStartupTasks(Platform.java:196)
	at org.sonar.server.platform.Platform.access$400(Platform.java:46)
	at org.sonar.server.platform.Platform$1.lambda$doRun$1(Platform.java:121)
	at org.sonar.server.platform.Platform$AutoStarterRunnable.runIfNotAborted(Platform.java:371)
	at org.sonar.server.platform.Platform$1.doRun(Platform.java:121)
	at org.sonar.server.platform.Platform$AutoStarterRunnable.run(Platform.java:355)
	at java.lang.Thread.run(Thread.java:748)
2018.01.19 14:46:04 INFO  web[][o.s.p.StopWatcher] Stopping process
```

ericlemes:
Hi @Bertk, they are 3 independent pull requests. None are obsolete. I can merge them in one, if this facilitates the review process.

Bertk:
@ericlemes Hi,
the last PR also contains the other commits. Merging this PR is sufficient.
Cheers Bert
dumping list of ranked functions

ericlemes:
@Bertk, you are right. It contains the other commits. I'll close the other PR's then. I still didn't understand why it is not starting sonar server. My local version works. I'll investigate further.

ericlemes:
The reason why it is failing is because I'm registering the metrics on both plugins (C and C++) and they have the same ID. I'm not sure which is the right way of doing this and will investigate further. If you load only one once, works fine.

I'm not sure if the right way here is having one different metric ID per language (does not make much sense for me), or have one separated plugin just to register the metrics.

guwirth:
@ericlemes before going into details:
* Could you explain a little bit what the intention of the metrics is and how they are to use?
* What can someone derive for his code base using these metrics?

ericlemes:
Hi @guwirth 

The goals here are:
- Have a percentage of code that is hard to maintain, according to lines of code per function criteria (as suggested by Clean Code)
- Have a percentage of code that is hard to maintain, according to cyclomatic  complexity criteria (as suggested by McCabe)
- Have a metric that can be easily compared different code bases.

What someone can derive is:
- How much of my code base needs refactoring?
- How well crafted is my source code compared to other code bases?

Regards,

Eric

guwirth:
@ericlemes 

> Have a percentage of code that is hard to maintain, according to cyclomatic complexity criteria (as suggested by McCabe)

Here the question is if cyclomatic complexity or cognitive complexity is the right choice to us. 
* cyclomatic complexity: is testability
* cognitive complexity: is readability

Somewhere I saw / read that you like to create files on server side. Think that's not a good idea because it's not ensured to have write access. Right way would be to create a REST API to ask for it?

ericlemes:
Hi @guwirth,

I'd keep both complexities. The way the metric is expressed as %, can be done for cognitive complexity as well.

In terms of the file dump. What I'm trying to do is just the "drill down" per function. Since sonar doesn't have a mechanism which it can be sent to the server (measure per function instead of per file), my implementation just dumps to a file (scanner side) if the file name is provided as a property. I'm investigating options of how to upload this.

I've tested these metrics in my sonar instance, and I'm quite happy with the results. I'm thinking about implementing another one which is the same idea of expressing as a %, but expressing % of Lines of Code in Functions with more than 20 lines of code and % of lines of code in functions with more than cyclomatic complexity 10. The idea is just to observe what is the distortion in the %. Probably one of both would be enough, but I personally don't see harm in keeping both views.

I really enjoyed how the % of functions with cyclomatic complexity is more representative than the avg cc/function. In some repos I've analysed, the avg cc/function has a variation of .5 CC/function between 2 repos. Using the % of complex functions, the same 2 repos has difference of 1.5% of code in complex functions.

ericlemes:
So guys, after a while trying to figure out what is going on here, I finally got to the point. A lot of learning about sonar plugin building going on.

The simple explanation is: I've defined new metrics in both plugins. You can't do that. Metrics should have a unique key and be shared across plugins.

There are 2 fixes for this, one that is hack, the other one which is sounds right, but requires more work. I'm happy to do the work, but I'd like to have an idea about how likely is for this PR being merged before I do a lot of work then throw everything on the bin. To cut to the chase, I feel that you are not comfortable with the new metrics at all.

Fix 1 (the hack): Define 2 different metrics, with different keys. One for C the other one for C++. The result won't be consolidated. I don't like this. Because I believe these metrics can be consolidated between different languages.
Fix 2: Define the metrics on another plugin, that defines the metrics and compute the consolidated numbers (using a MeasureComputer). This plugin with these new metrics can be shared with other languages that want to implement the same metrics (for example, C#, Java or TypeScript), and ideally should be in a new project at the same level of sonar-cxx (I'm really accepting name suggestions). Then make the C and C++ plugin implementing these extra metrics.

I'd like to hear your thoughts before moving forward, so I can manage my expectations.


Regards

Eric

guwirth:
@ericlemes like the ideas of the new metrics. I'm not sure if the c plugin has a future. So maybe we combine it anyway to one plugin again. Think what should not be done is to write local files during scanner run. So most likely fix 1 is ok.

ericlemes:
@guwirth nice to hear that.

I think the discussion about the new metrics in a separated, non cxx project would make sense anyway. I'm not sure if the future of the c plugin would be blocking this discussion.

Regarding the files, I just got some hope about that. I was chatting in the sonar list, and apparently it is possible to push this data and display in a custom JS page. I'll put some effort on that, but I'm happy to take this off the plugin to simplify the merge.

So, can you help me giving some sort of "steps" we can follow together to help this code and this PR be accepted? In my point of view they are:

1) Create a separated project only defined in the metrics. The reason is that other languages could push the same metrics.
2) Remove the "file dump" code from existing PR
3) Adapt existing PR to cope with separated metrics project and make sure C and C++ plugin works.
4) Merge the PR.

Does that make sense?

Cheers

Eric

ericlemes:
Hey @guwirth,

I did some changes, removing the file dumps and code from C Plugin. Now apparently some integration tests are breaking and I have no idea why, since I'm not dumping any new warnings.

Can you help me?

Bertk:
Hi,

2Tests fail.

```
Failing scenarios:
  integration-tests/features/multimodule_analysis.feature:5  cpp-multimodule-project
  integration-tests/features/x_importing_cppcheck_reports_c_cpp.feature:19  Importing cppcheck issues when c language issues are in report. -- @1.1 

```
The log file shows the exception

https://appveyorcidata.blob.core.windows.net/sonaropencommunity-29197/sonar-cxx/1180/i80otcvrnp5j2xqj/_cppcheck_project_c_cpp_.log?sv=2015-12-11&sr=c&sig=f3rcEy%2BEKuFEgu80YrO8LwxO6HjtDSdMfNzcx8GV7l8%3D&st=2018-01-31T18%3A32%3A52Z&se=2018-01-31T18%3A38%3A52Z&sp=r

Best regards
Bert

ericlemes:
Thanks @Bertk . After I learnt how to read the logs, I finally managed to fix all the problems.

guwirth:
@ericlemes is it ready to review?

ericlemes:
@guwirth, yes it is.

cxx-sensors/src/test/java/org/sonar/cxx/sensors/functioncomplexity/CxxFunctionComplexitySquidSensorTest.java
```diff
@@ -0,0 +1,169 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.cxx.sensors.functioncomplexity;
+
+import java.io.ByteArrayOutputStream;
```
Bertk: Please remove unused imports.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/functioncomplexity/CxxFunctionComplexitySquidSensor.java
```diff
@@ -0,0 +1,217 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.cxx.sensors.functioncomplexity;
+
+import com.sonar.sslr.api.AstNode;
```
Bertk: Please remove unused imports.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/functioncomplexity/CxxFunctionComplexitySquidSensor.java
```diff
@@ -0,0 +1,217 @@
...
+  
+  private Hashtable<SourceFile, FunctionCount> complexFunctionsPerFile = new Hashtable<>();      
+  
+  private Hashtable<SourceFile, FunctionCount> locInComplexFunctionsPerFile = new Hashtable<>();      
+  
+  private String fileName; 
+  
+  public CxxFunctionComplexitySquidSensor(CxxLanguage language){    
+    this.cyclomaticComplexityThreshold = language.getIntegerOption(FUNCTION_COMPLEXITY_THRESHOLD_KEY).orElse(10);
+    LOG.debug("Cyclomatic complexity threshold: " + this.cyclomaticComplexityThreshold);
```
Bertk: Please add log guard
rule pmd:GuardLogStatement

cxx-sensors/src/main/java/org/sonar/cxx/sensors/functionsize/FunctionSizeMetrics.java
```diff
@@ -0,0 +1,71 @@
...
+import org.sonar.api.measures.CoreMetrics;
+import org.sonar.api.measures.Metric;
+import org.sonar.api.measures.Metrics;
+
+public class FunctionSizeMetrics implements Metrics {
+  
+  public static final Metric<Integer> BIG_FUNCTIONS = new Metric.Builder("big_functions", "Big Functions", Metric.ValueType.INT)
+    .setDescription("Number of functions with too many lines")
+    .setDirection(Metric.DIRECTION_WORST)
+    .setQualitative(false)
```
Bertk: Please use `Boolean.FALSE`
rule fb-contrib:NAB_NEEDLESS_BOOLEAN_CONSTANT_CONVERSION

cxx-sensors/src/main/java/org/sonar/cxx/sensors/functioncomplexity/FunctionComplexityMetrics.java
```diff
@@ -0,0 +1,63 @@
...
+import org.sonar.api.measures.CoreMetrics;
+import org.sonar.api.measures.Metric;
+import org.sonar.api.measures.Metrics;
+
+public class FunctionComplexityMetrics implements Metrics {
+  
+  public static final Metric<Integer> COMPLEX_FUNCTIONS = new Metric.Builder("complex_functions", "Complex Functions", Metric.ValueType.INT)
+    .setDescription("Number of functions with high cyclomatic complexity")
+    .setDirection(Metric.DIRECTION_WORST)
+    .setQualitative(false)
```
Bertk: Please use` Boolean.FALSE`
rule fb-contrib:NAB_NEEDLESS_BOOLEAN_CONSTANT_CONVERSION

cxx-sensors/src/main/java/org/sonar/cxx/sensors/functioncomplexity/CxxFunctionComplexitySquidSensor.java
guwirth: add if (LOG.isDebugEnabled())

cxx-sensors/src/main/java/org/sonar/cxx/sensors/functionsize/CxxFunctionSizeSquidSensor.java
```diff
@@ -0,0 +1,225 @@
...
+  
+  private int locOverThreshold = 0;
+  
+  private Hashtable<SourceFile, FunctionCount> bigFunctionsPerFile = new Hashtable<>();
+  
+  private Hashtable<SourceFile, FunctionCount> locInBigFunctionsPerFile = new Hashtable<>();  
+  
+  public CxxFunctionSizeSquidSensor(CxxLanguage language){
+    this.sizeThreshold = language.getIntegerOption(FUNCTION_SIZE_THRESHOLD_KEY).orElse(20);
+    LOG.debug("Function size threshold: " + this.sizeThreshold);
```
guwirth: add if (LOG.isDebugEnabled())

cxx-sensors/src/main/java/org/sonar/cxx/sensors/squid/CxxSquidSensor.java
```diff
@@ -114,6 +122,15 @@ public CxxSquidSensor(CxxLanguage language,
     if (this.language.getMetricsCache().isEmpty()) {
       new CxxMetrics(this.language);
     }
+    
+    registerSquidSensors();
+  }
+  
+  protected void registerSquidSensors(){
+    if (this.language.getKey() == "c++"){
```
guwirth: do not use "c++" use constant for plugin key instead

cxx-squid/src/main/java/org/sonar/cxx/CxxLanguage.java
```diff
@@ -40,7 +40,7 @@
   public CxxLanguage(String key, Configuration settings) {
     super(key);
     this.settings = settings;
-    this.MetricsCache = new HashMap<>();
+    this.MetricsCache = new HashMap<>();
```
guwirth: remove blanks at eol

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxPlugin.java
```diff
@@ -89,7 +93,7 @@
   public static final String JSON_COMPILATION_DATABASE_KEY = LANG_PROP_PREFIX + "jsonCompilationDatabase";
   public static final String SCAN_ONLY_SPECIFIED_SOURCES_KEY = LANG_PROP_PREFIX + "scanOnlySpecifiedSources";
   public static final String CPD_IGNORE_LITERALS_KEY = LANG_PROP_PREFIX + "cpd.ignoreLiterals";
-  public static final String CPD_IGNORE_IDENTIFIERS_KEY = LANG_PROP_PREFIX + "cpd.ignoreIdentifiers";
+  public static final String CPD_IGNORE_IDENTIFIERS_KEY = LANG_PROP_PREFIX + "cpd.ignoreIdentifiers";
```
guwirth: remove blanks at eol

guwirth:
keys in settings file:
```
sonar.cxx.funccomplexity.threshold
sonar.cxx.funcsize.threshold
```
Wouldn't it be better something like?
```
sonar.cxx.measure.func.complexity.threshold
sonar.cxx.measure.func.size.threshold
```

guwirth:
@ericlemes how does the results look like in the UI? Can you add a screenshot please?

guwirth:
@ericlemes question: think the metrics are very generic and could be used for all languages. Wouldn't it be better to add it to an own plugin? Like to avoid that we have it in one version and one version later it's removed again.

guwirth:
@ericlemes solution I expected to solve this was:
* add new rule/check to create messages if over threshold
* define threshold in rule/check
* count rule violations and add result to a measure

Why did you select this approach?

guwirth:
@ericlemes code looks good 👍

ericlemes:
Thanks @Bertk and @guwirth  for the review. There's a lot going on here, but I'm glad that apart from all the requests you like the code ;-)

First question to @Bertk: Where did you get all this feedback about the rules? Is there any instance of Sonar that the CI is pushing information? I was struggling to get this, run the integration tests locally and the code coverage locally, and that's one of the reasons that I got loads of violations. If you can help me setting up a good environment, I can definitely be more productive. And Java is probably my 3rd language, so I know very little about the environment.

Segund to @guwirth. I totally agree with you. I think the metrics are very generic and should be implemented in a different plugin. But, I'd really like your thoughts and guidance on that. First, if they are generic and could be implemented to other languages (which I'm intending to do), I guess they should be outside sonar-cxx project. I was really thinking about something like sonar-opencommunity-metrics in a new github project. I'm definitely help to implement like that but I'd like to agree on general design before spending a lot of time refactoring everything. Also, I was really struggling with maven/sonar stuff to create a new project. If you could create a skeleton on Git Hub which I could fork, that would definitely not only save me time, but help us to be more assertive.

Generally speaking, I'm really happy that you guys are keen to accept this code. I personally doing a lot of experimentation with the metrics to understand how the numbers can be generic and meaningful at the same time.

I'm planning to do all the reviews, but I'd really would like a definition about the new plugin structure before moving forward. Can you help me with this?

ericlemes:
Here is the screenshot. The numbers are from WebKit project.

screenshot

guwirth:
> I'm planning to do all the reviews, but I'd really would like a definition about the new plugin structure before moving forward. Can you help me with this?

@ericlemes Think no one of us has really experience with writing a new plugin from scratch. But there is a description and a template. Maybe that helps:

* Developing a Plugin: https://docs.sonarqube.org/display/DEV/Developing+a+Plugin
* The recommended way is to duplicate the example located in the GitHub repository: https://github.com/SonarSource/sonar-custom-plugin-example.

guwirth:
> Here is the screenshot....

@ericlemes screen-shot and metrics look good. Only thing I would change is the  naming. Looking to other metric names:
* % is always at the end as (%)
* start always with same name => easier to read

e.g.:
* Big Functions
* Big Functions (%)
* Big Functions Lines of Code
* Big Functions Lines of Code (%)
...

ericlemes:
Hi @guwirth  and @Bertk,

Took a while, but I think I got there now.. I had to write from scratch the new project that defines the metric, so I did it, and it is in https://github.com/ericlemes/sonar-opencommunity-metrics

I need to transfer to you, so we can merge this pull request, since sonar-cxx will depend on that project.

So, the project contains 100% code coverage, and it is on a CI in the cloud. Please, let me know how we can work together to get this merged.

ericlemes:
The CI is obviously failing, because it couldn't find the dependency.

guwirth:
@ericlemes what are exactly your plans?
* Code in your repository or in our repository?
* At the beginning your idea was to provide generic metrics for all languages. Is this still the case?

ericlemes:
Hi @guwirth ,

1) Your repository. I just would like how to discuss this. I think you can fork or I can just transfer the whole repo, But I guess you would like to review first, right?
2) Generic metrics. Hence the reason I had a lot of work to move the metric definition and computation to a main project. The way it works in Sonar is that you have to create "Measure Computers" to agregate values from components, that can be generated from different languages and then aggregated in a single value. The way it works is, one or more languages provide the "low level metrics" (Ex.: Lines of Code in Functions, Big Functions LoC) than the measure computers takes those values and computes the high level metrics, like Big Functions LoC (%).

Please, I need your advice about how we can organise this code together and integrate properly.

ericlemes:
Hi @guwirth, 

Do you have any updates on this? Are we still moving forward?

guwirth:
Hi @ericlemes sorry for the late answer - I had always the hope the someone else is also adding his opinion.
For my understanding the code should be inside of our plug-in. Don't see that someone else will use it.

ericlemes:
Hi @guwirth,

I'm very confused now.

In your comment in 25/Jan you suggested you would be Ok keeping everything on the C++ plugin, because the C plugin didn't have a future.

Then in your other comment, in 3/Feb you said they should be moved to their own plugin. Took me a lot of work to get that done, hence the reason why this PR was untouched for a while.

Then you came back on 21/May and said they should be in the same plugin again.

So, can you please make a final decision about this? 

The reason why this is in another plugin is because if someone wants to the new metrics (for example Complex Functions LoC (%)) consolidated between different languages, for example C# and C++, they can implement in the same metric if it is defined in its own plugin. If it is defined on C++ plugin, this can't be done. Ever. Because you can't implement 2 metrics in different plugins with the same key.


Regards,

Eric

guwirth:
Hi @ericlemes, sorry for the confusion.

* My first statement was: in case these are language independent metrics it would make sense to put it into an own plug-in which works with all languages. So users can install your plug-in as addition if they want the functionality. I understood you that your idea with the own plug-in was to to provide an own independent plug-in?

* My understanding (without doing a review) is that your code has hard dependencies to the cxx plug-in. That's the reason why I said: if only working with cxx plug-in would be better to have it completely inside of the cxx plug-in.

Are there hard dependencies?
Is it working with other languages?

Regards,

ericlemes:
Hi @guwirth,

There's no "my plugin". This code that is in my GitHub, the sonar-opencommunity-metrics, the intention is to transfer this whole repo to the SonarOpenCommunity org. I don't want to own and maintain any of those alone. I just didn't do this yet because I don't know if you want to keep the code. If we coordinate how to do this, that is my intention.

So, there is a hard dependency from sonar.cxx to sonar-opencommunity-metrics, because sonar.cxx must know the definition of the metric to provide the numbers.

The metric itself is language-independent and if I implement the metric definition inside sonar.cxx it won't be. If someone else wants to implement the same metric in a different language it won't be able to aggregate the numbers.

It is not implemented for other languages, hence the reason why it is not working for other languages. If in the future other language wants to implement these metrics, with the metrics inside sonar.cxx, this won't be possible.

Regards,

Eric

guwirth:
Hi @ericlemes,

I take a first look into the code:

* Code is an own SQ plug-in, you implement: https://github.com/ericlemes/sonar-opencommunity-metrics/blob/master/src/main/java/org/sonar/opencommunity/metrics/MetricsPlugin.java

> Please, let me know how we can work together to get this merged.

So my current understanding is:
* Keep the code (plug-in) independent but move it to https://github.com/SonarOpenCommunity/. But not inside sonar-cxx, one level higher e.g. on MetricsPlugin?
* We build and distribute plug-in on https://github.com/SonarOpenCommunity/
* An user has to install cxx-plugin and MetricsPlugin?
   * What is if only one plugin is installed?
* On the other hand I saw that you are calling the new sensors in the cxx-plugin. Means hard dependency. But then your MetricsPlugin makes no sense? The code is then more a package others can use?

Maybe I'm slow on the uptake, but I still don't understand it exactly? Can you explain it please?

Regards,

ericlemes:
@guwirth  Thank you for you time.

> Code is an own SQ plug-in, you implement: https://github.com/ericlemes/sonar-opencommunity-metrics/blob/master/src/main/java/org/sonar/opencommunity/metrics/MetricsPlugin.java

Correct!

> Keep the code (plug-in) independent but move it to https://github.com/SonarOpenCommunity/. But not inside sonar-cxx, one level higher e.g. on MetricsPlugin?

Correct! I suggested the name "OpenCommunityMetrics", which is the metrics defined with SonarOpenCommunity. But you can change if you are not happy.

> We build and distribute plug-in on https://github.com/SonarOpenCommunity/

Correct!

> An user has to install cxx-plugin and MetricsPlugin?

Correct!

> What is if only one plugin is installed?

The sonar.cxx plugin will collect the metrics, the scanner will give an error because the definition of the metric will not be found.

> On the other hand I saw that you are calling the new sensors in the cxx-plugin. Means hard dependency. But then your MetricsPlugin makes no sense? The code is then more a package others can use?
> Maybe I'm slow on the uptake, but I still don't understand it exactly? Can you explain it please?

It is not simple. Your understanding is correct. It is a hard dependency. On the other hand if it is not designed like this, other languages can't aggregate their number on the same metric if they are implemented in different plugins. We don't have this problem with Sonar Metrics, because they are implemented in the class CoreMetrics, which is part of the framework.

For example, if we implement Complex Functions LoC (%) for let's say, TypeScript, with this design we can aggregate for the same project C++ and TypeScript. If we implement inside sonar.cxx, there will be 2 different metrics with the same (or similar name) in the same project. This will show within other aggregated measures, like Cyclomatic Complexity.

If this helps, we can schedule a call or an IM session to clarify all the points.


Regards,

ericlemes:
@guwirth 

> If one plug-in is not working without the other it makes no sense to split it into two plug-ins?

We could let the sonar.cxx not pushing the custom metrics by default and enable it on a property. With this you won't break people that don't want the new metrics and break the dependency between the plugins.

> Other languages are all implemented by SonarSources. My experience: chance is very low that they add it.

Or community. I'm interested in implementing this for other languages.

> I'm still wondering if there is no way to decouple your metric plugin from the language plugins? If it is ensured that your sensors are running after the language plugins they should be able to read out the metrics from other languages and create own metrics on top of it?

I've tried this and failed. In fact I've run in the bug whilst doing this between C and C++. If both languages pushes the aggregated metrics, you have an exception. If both languages adds the Metric class to the Plugin Context, it is an exception in the initalisation.

guwirth:
@ericlemes 

> break the dependency between the plugins.

This is a precondition to do it. 

> I've tried this and failed. In fact I've run in the bug whilst doing this between C and C++. If both languages pushes the aggregated metrics, you have an exception. If both languages adds the Metric class to the Plugin Context, it is an exception in the initalisation.

That’s a thing I still don’t understand. Looking to other metrics they are defined in the core or a plugin and you save the value for the context and a file https://github.com/SonarOpenCommunity/sonar-cxx/blob/8edbb8cbe0d3d270a7b783603d82d5841dc5083a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/squid/CxxSquidSensor.java#L219. The metric itself is always language independent? 

I'm still wondering if there is no way to decouple your metric plugin from the language plugins? If it is ensured that your sensors are running after the language plugins they should be able to read out the metrics from other languages and create own metrics on top of it?

ericlemes:
@guwirth 

The only way of breaking the dependency is through a property. Even though the binary dependency would still be there. sonar.cxx will have the maven dependency to sonar.opencommunity.metrics.
 
The reason why it works in sonar core metrics is because the metric class is defined by Sonar core, not by the plugin. You can't define a metric in more than one plugin with the same ID. You can't publish measures (which will be needed by the plugin) if you don't have the metric. That's how it works by design. I tested, tried and spent a lot of time to get to this conclusion and working.

So, I really appreciate if you decide if you want the contribution or not. I did the first pull request, and first you asked me to split the plugin, now you are not happy with it. I spent a lot of time and effort to get this done and I'm honestly spending more time with this discussion then actually if I'd go down the route of writing a new one. Soon this PR will make 6 months being active. So, I don't think this is reasonable. Is totally fair for you to not accept the contribution, but is not fair to not decide what kind of contribution you would like or not and pushing this back forever.

If you don't want contributions to the project or if you want very specific ones, I really suggest you to update your CONTRIBUTING guidelines so you don't waste anyone time.

guwirth:
Hi @ericlemes, 

I still see the value of your metrics. There are two possibilities:

1. Make your plugin independent from others. If you want we can host it in our repository. The hard dependency to the cxx plugin is a no-go and result in a maintenance hell. 
2. Move the code inside of the cxx plugin. With the restrictions of the review comments: e.g. writing local files is not supported. 

You say 1 is not possible so there is only option 2. 

Regards,

guwirth:
@ericlemes please tell us when we can do final review.

ericlemes:
@ivangalkin, I've fixed your style comments.

Related to the visualization of the metric, I think your comment does not apply. This metrics does not raise any issues in the code. Thy only present high level metrics, based on all the code base.

ericlemes:
@guwirth Ready for final review.

guwirth:
Hi @ericlemes,
Thanks a lot for your contribution!
Regards,

ericlemes:
Hey @guwirth. You are welcome. 
Do you have any idea when this will be released?

PR #1419, opened 2018-02-03
RATS 2.4 support

### LINKED ISSUE #1395: RATS: Errors when processing reports
When processing RATS reports (version 2.3) we get the following errors:
```
ERROR: Could not add the issue 'The rule 'rats:strcpy' does not exist.', skipping issue
ERROR: Could not add the issue 'The rule 'rats:strcat' does not exist.', skipping issue
ERROR: Could not add the issue 'The rule 'rats:fread' does not exist.', skipping issue
ERROR: Could not add the issue 'The rule 'rats:program_name' does not exist.', skipping issue
```

I have noted that `strcpy` and `strcat` are available here but with a different case:
https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/resources/rats.xml

### LINKED ISSUE #1396: RATS: Strange entries in SQ after adding RATS reports
As seen int the provided screenshot, importing RATS reports adds issues to SQ with the wrong subject (**SQ 6.7** with **C++ (Community) 0.9.8**).

In the displayed case the true issue is `rats:printf` but the subject in the issues list is "A function call is not being made here..." instead of "Check to be sure that the non-constant format string...".

image

guwirth:
- https://github.com/andrew-d/rough-auditing-tool-for-security/blob/master/rats-c.xml
- close #1395
- close #1396

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1419)

### DIFF
--- a/cxx-sensors/src/main/resources/rats.xml
+++ b/cxx-sensors/src/main/resources/rats.xml
@@ -30,7 +30,7 @@ Follow these steps to make your custom Custom rules available in SonarQube:
       sufficient entropy should be used.
     </description>
     <tag>security</tag>
-    <internalKey>config/drand48</internalKey>
+    <internalKey>drand48</internalKey>
     <severity>MAJOR</severity>
     <type>CODE_SMELL</type>
     <remediationFunction>LINEAR</remediationFunction>
@@ -47,7 +47,7 @@ Follow these steps to make your custom Custom rules available in SonarQube:
       sufficient entropy should be used.
     </description>
     <tag>security</tag>
-    <internalKey>config/erand48</internalKey>
+    <internalKey>erand48</internalKey>
     <severity>MAJOR</severity>
     <type>CODE_SMELL</type>
     <remediationFunction>LINEAR</remediationFunction>
@@ -64,7 +64,7 @@ Follow these steps to make your custom Custom rules available in SonarQube:
       sufficient entropy should be used.
     </description>
     <tag>security</tag>
-    <internalKey>config/initstate</internalKey>
+    <internalKey>initstate</internalKey>
     <severity>MAJOR</severity>
     <type>CODE_SMELL</type>
     <remediationFunction>LINEAR</remediationFunction>
@@ -81,7 +81,7 @@ Follow these steps to make your custom Custom rules available in SonarQube:
       sufficient entropy should be used.
     </description>
     <tag>security</tag>
-    <internalKey>config/jrand48</internalKey>
+    <internalKey>jrand48</internalKey>
     <severity>MAJOR</severity>
     <type>CODE_SMELL</type>
     <remediationFunction>LINEAR</remediationFunction>
@@ -98,7 +98,7 @@ Follow these steps to make your custom Custom rules available in SonarQube:
       sufficient entropy should be used.
     </description>
     <tag>security</tag>
-    <internalKey>config/lcong48</internalKey>
+    <internalKey>lcong48</internalKey>
     <severity>MAJOR</severity>
     <type>CODE_SMELL</type>
     <remediationFunction>LINEAR</remediationFunction>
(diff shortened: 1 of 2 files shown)


guwirth:
@Bertk do yo remember why internalKey has this config/...? Normally key is equal to internalKey?

```XML
<rule>
   <key>strcpy</key>
   <internalKey>config/strcpy</internalKey>
</rule>
```

Bertk:
@guwirth Sorry, but I do not know the reason why this was implemented in this way. This was introduced with the initial version of the CXX plug-in.

```
commit 20b6de69d1896b8c8b8bc23713b30ba7304bcf1d
Author: Francois Dorin <f.dorin@free.fr> 2010-11-29 21:16:38
Committer: Francois Dorin <f.dorin@free.fr> 2010-11-29 21:16:38
```

jmecosta:
@Bertk the idea to remove those was in case if we would implement those rules into the CxxChecks. Have i got the impression that @guwirth had a different view for those CxxChecks? 

Anyway, for Vera because its mostly a linter. I supose replacing it with something more mainstream like cpplint with be more beneficial for end users. 

for Rats, for security  analysis i dont think there is no replacement for that. 

CppLint could be a interesting replacement for linter, where the sensor actually runs the linter since it does require little configuration. the only dependcy is python, and that should be available in most boxes

guwirth:
@Bertk and @jmecosta my idea is still:
* don't reinvent the wheel
* the plugin is only to forward issues from other tool to SQ

Think we have no chance to do the same work as other are already did: VS, Core Guideline Checker, Clang, Cppcheck, ... are too good to reimplement everything again...

jmecosta:
@guwirth i dont know, if we have a grammar, it seems logic that we also have checks. otherwise just run a external tool collect those metrics and simplify the plugin a lot.

alternatively if really willing to just consume external stuff, integrate a clang static analyser by default that gives all metrics and provides basic checks. If those are handle with clang, there is much more possibility of community particiating on developing those. That means dropping totally the sslr stuff from the plugin.

By the way, thats what c# plugin does, so they use the roslyn compiler so they dont need to worry about supporting the grammar. 

as far as i see you spend loads of time supporting the grammar only for the marginal gain of having some code metrics

Bertk:
@guwirth , @jmecosta 
We should definitely ask the users of the C++ Community plug-in which sensors are used and decide/vote on deprecating some outdated analyzer tools. My favorite candidates are [RATS](https://code.google.com/archive/p/rough-auditing-tool-for-security/) and [Vera++](https://bitbucket.org/verateam/vera/wiki/Home) because the last updates of this tools were December 2013 for RATS and January 2015 for Vera++. The last updates of Vera++ added support for Python and Lua with the goal to replace Tcl. I think using RATS does not improve the security and I disabled the tool already 3 years ago in our build pipelines. There are better tools available and other ways to handle security requirements e.g. [banned.h](https://cloudblogs.microsoft.com/microsoftsecure/2012/08/30/microsofts-free-security-tools-banned-h/) ...

I think the current implementation of AST based checkers and CppCheck are better alternatives. I used AST based checkers to implement custom rules and also extended the grammar for CUDA and C++/CLI.
I disagree with the last comment about the C# plug-in because there is only one compiler and we use many C++ compilers (VC++, gcc, clang). This will not change at least for the next decade. 

We should continue source code analysis with AST but create checks and rules which are not available in other tools. This can be revised in 2020 with the next C++ standard.

jmecosta:
I'm sure clang or derivatives can be used as frontend for many c flavours
https://clang.llvm.org.

But that's not the point, I'm only saying if everything is from external
tools, the usage of cxx grammar brings only  a marginal benifit because
it's only use for metrics.

Now if you start developing new checks then effectively keeping tools like
Vera or rats makes no sense since what that provides can be implemented on
top of sslr.

Imo opinion it's a difficult road, because we are always affected by the
sslr versions and it's updates. So far it has been challenging to keep up
with the new versions.

But I'm not against or in favour of either alternative. And I would
probably vote on dropping those tools since they can anyway be provided as
external rules.

Issue #1425, opened 2018-02-23
Wrong code coverage for empty line, constexpr, method declaration


Bertk:
The code coverage for my [demo project](https://github.com/Bertk/cpptesting/blob/master/Source/MotorController/MotorController.cpp) is not correct:
<img width="652" alt="wrong-coverage" src="https://user-images.githubusercontent.com/2315215/36585990-e1ec8938-1877-11e8-89f9-32fcf028ebd6.PNG">

### Steps to reproduce the problem
1. Execute TFS vNext build with SonarQube TFS extension
2. Check Coverage for MotorController.cpp
### Expected behavior
line 8, 12, 17 should not be reported as "not covered by test"
<img width="675" alt="vs-coverage" src="https://user-images.githubusercontent.com/2315215/36605237-1423258e-18b8-11e8-9909-6d7ff131b49e.PNG">

### Actual behavior
see screen shot above
### LOG file
see attachment without debug log setting - no warning related to wrong coverage
```
2018-02-23T08:31:03.1371871Z INFO: Sensor c++ CoverageSensor [cxx]
2018-02-23T08:31:03.1373372Z INFO: Searching coverage reports by path with basedir 'C:\agent\_work\1\s\Source\MotorController' and search prop 'sonar.cxx.coverage.reportPath'
2018-02-23T08:31:03.1373372Z INFO: Searching coverage reports by path with basedir 'C:\agent\_work\1\s\Source\MotorController' and search prop 'sonar.cxx.coverage.reportPath'
2018-02-23T08:31:03.1374868Z INFO: Searching for coverage reports '[C:/agent/_work/1/s/TestResults/**/*.coveragexml]'
2018-02-23T08:31:03.1374868Z INFO: Searching for coverage reports '[C:/agent/_work/1/s/TestResults/**/*.coveragexml]'
2018-02-23T08:31:03.1376110Z INFO: Coverage BaseDir 'C:\agent\_work\1\s\Source\MotorController' 
2018-02-23T08:31:03.1381299Z WARNING: WARN: Access to the multi-values/property set property 'sonar.cxx.coverage.reportPath' should be made using 'getStringArray' method. The SonarQube plugin using this property should be updated.
2018-02-23T08:31:03.1594529Z INFO: Saved '5' coverage measures for file 'C:\agent\_work\1\s\Source\MotorController\dllmain.cpp'
[...]

PR #1426, opened 2018-02-24
refactor CxxFileLinesVisitorTest

Bertk:
- only one assert per test
- add constexpr in test file "ncloc.cc"

### DIFF
--- a/cxx-sensors/src/test/java/org/sonar/cxx/sensors/visitors/CxxFileLinesVisitorTest.java
+++ b/cxx-sensors/src/test/java/org/sonar/cxx/sensors/visitors/CxxFileLinesVisitorTest.java
@@ -25,12 +25,17 @@ import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import static org.assertj.core.api.Assertions.assertThat;
+import org.assertj.core.api.SoftAssertions;
 import org.junit.Before;
 import org.junit.Test;
 import static org.mockito.Mockito.mock;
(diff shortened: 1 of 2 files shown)


Bertk:
Solves issue #1425

cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/VisualStudioParser.java
```diff
@@ -122,7 +119,6 @@ private static void collectRangeMeasures(SMInputCursor function, Map<String, Cov
           coverageData.put(sourceId, builder);
         }
 
-        builder.setConditions(startLine - 1, conditions, coveredConditions);
```
guwirth: @Bertk don't understand why this is no more needed?

Bertk: There are 2 issues with this line:
- The calculation ```startLine - 1``` creates measures for wrong code lines
- Microsoft coverage tool only provides block coverage but not branch coverage (_selection statements_). The indicator for **partial covered condition** cannot be set for the relevant code line but only for the code block and the code metric shows more conditions for a source file than conditions exists in the file itself.

A compromise is to remove the calculation and set condition for every block but this is a different way Visual Studio does it and should be avoided.

motorcontroller-coverage

Bertk: The solution with the compromise produces this output:
partial-covered-block

motorcontroller-measuredetails

guwirth:
@Bertk does this still fit to description in rdwiki https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Get-code-coverage-metrics?

Bertk:
@guwirth I updated the page and removed the invalid parts.

Issue #1440, opened 2018-03-23
Handle multiple "stack" nodes for errors in Valgrind XML report files


jhoussay:
Hi,

For some Valgrind errors (such as "InvalidWrite"), the corresponding XML node contains two distinct "stack" nodes, one corresponding to the allocation stack and the other one corresponding to the memory misuse stack.

Unfortunately, sonar cxx only presents the content of one "stack'" node found in the "error" node. In my case, this is pretty annoying as the stack available in Sonar is the allocation stack and does not allow easy identication of the memory misuse.

It seems to come from the `parseErrorTag `method in [ValgrindReportParser.java](https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/valgrind/ValgrindReportParser.java). When multiple "stack" nodes are contained in the "error" node, only the last one is kept:
```java
String kind = null;
String text = null;
ValgrindStack stack = null;
while (child.getNext() != null) {
   String tagName = child.getLocalName();
    if ("kind".equalsIgnoreCase(tagName)) {
      kind = child.getElemStringValue();
    } else if ("xwhat".equalsIgnoreCase(tagName)) {
      text = child.childElementCursor("text").advance().getElemStringValue();
    } else if ("what".equalsIgnoreCase(tagName)) {
      text = child.getElemStringValue();
    } else if ("stack".equalsIgnoreCase(tagName)) {
      stack = parseStackTag(child);
    }
}

[...]

return new ValgrindError(kind, text, stack);
```

Handling multliple stacks at this point would require to do so in [ValgrindError.java](https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/valgrind/ValgrindError.java) as well and modify the `getLastOwnFrame` method in order to provide the "best" stack origin...

Julien

guwirth:
@jhoussay could you
* provide an XML report sample please. 
* Is this one message with two locations or two messages? Please have also a look to https://github.com/SonarOpenCommunity/sonar-cxx/pull/1436. Similar to this?

jhoussay:
Hi @guwirth ,

Here is an example of report : [valgrind_output.xml.txt](https://github.com/SonarOpenCommunity/sonar-cxx/files/1846742/valgrind_output.xml.txt).

[...]

PR #1436, opened 2018-03-22
fix CppcheckParserV2; allow multiple NewIssueLocations

ivangalkin:
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1436)

PR #1447, opened 2018-04-02
Implement support for multiple locations in CxxReportSensor

ivangalkin:
enable multiple locations for
* CxxDrMemorySensor
* CppcheckParserV2
* CxxValgrindSensor
enable the new interface for all remaining sensors

this pull request is replacement for #1436
see also issue  #1440

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1447)


cxx-sensors/src/main/java/org/sonar/cxx/sensors/drmemory/CxxDrMemorySensor.java
```diff
@@ -82,22 +87,39 @@ protected void processReport(final SensorContext context, File report) {
...
-          break;
-        }
+      if (!locations.isEmpty()) {
+        CxxReportIssue issue = new CxxReportIssue(CxxDrMemoryRuleRepository.KEY, error.getType().getId(), locations);
+        saveUniqueViolation(context, issue);
       }
     }
   }
 
+  private void addFileLocations(final SensorContext context, DrMemoryError error, List<CxxReportLocation> locations) {
```
ivangalkin: same logic is implemented in CxxValgrindSensor

For each CxxReportLocation CxxReportSensor will check if path belongs to the project or not. If it doesn't - this [secondary] location will be ignored. So finally we will have 

* a primary location which is guaranteed to belong to the project (same as before)
* ```>= 0``` additional locations, which reference files from the analyzed project;

cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportIssue.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Issue with one or multiple locations
+ */
+public class CxxReportIssue {
```
ivangalkin: no logic here - CxxReportIssue and CxxReportLocation (1:n) were introduced more or less in order to implement the proper semantics of ::equals() and ::hashCode() methods

ivangalkin:
@guwirth Thank you for your proposal. I started to implement the `CxxReportIssue` in the way you suggested. However the existing code is very careful w.r.t. the XML (or other source) validation, so we save an issue only if at least 1 valid location exists. This makes the generic logic work like that:

1. inspect and collect all possible locations
2. if there is at least 1 valid one -> save the issue

Creation of an `CxxReportIssue` in advance is often impossible or makes the logic more complex.

Also there is a (minor) performance aspect (or rather memory footprint). Since the most of sensors support only primary location we can benefit from the usage of `Collections.singletonList()`. The suggested approach would always require a full-blown container.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/cppcheck/CppcheckParserV2.java
```diff
@@ -70,18 +76,18 @@ public void stream(SMHierarchicCursor rootCursor) throws XMLStreamException {
...
-                String id = errorCursor.getAttrValue("id");
-                String msg = createMsg(
-                  errorCursor.getAttrValue("inconclusive"),
-                  errorCursor.getAttrValue("msg")
-                );
-                String file = null;
-                String line = null;
+                String id = Objects.requireNonNull(errorCursor.getAttrValue("id"),
+                    "Missing mandatory attribute /results/errors/error[@id]");
+                String msg = Objects.requireNonNull(errorCursor.getAttrValue("msg"),
```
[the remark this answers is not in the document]
ivangalkin: thx! done

ivangalkin:
@Bertk please review manually, GitHub doesn't notice, that I've changed the SonarQube finding. I've fixed an issue, but in an other line as reviewed.

guwirth:
@ivangalkin 

> if there is at least 1 valid one -> save the issue

No location means issue on module/project level.

> Creation of an CxxReportIssue in advance is often impossible or makes the logic more complex.

Maybe ctor without location and only this `addLocation `method? If you need an extra check why not add an `isValid` method to `CxxReportIssue`?

ivangalkin:
@guwirth I'll try to apply the comments in the next few days.

PR #1445, opened 2018-03-30
cppcheck: support external library issues

ivangalkin:
Many cppcheck rules were missing, since they are are not listed by cppcheck --errorlist by default.
RuleIDs become visible if library descriptions are loaded (e.g. cppcheck --library=posix.cfg)

FYI: library description (configuration) is a metadata about some C/C++ API (STL, posix, QT, WinAPI etc).
Some function might be marked as undesired (obsolete, not thread-safe etc.). Their usage in code
causes warnings. That's the reason to add missing ruleIDs.

In order to find the missing ruleIDs I've fixed the old bash/python scripts.

```bash
cppcheck --errorlist --library=avr.cfg --library=bsd.cfg --library=cppcheck-lib.cfg --library=embedded_sql.cfg --library=gnu.cfg --library=gtk.cfg --library=microsoft_sal.cfg --library=motif.cfg --library=posix.cfg --library=qt.cfg --library=sdl.cfg --library=sfml.cfg --library=std.cfg --library=windows.cfg --library=wxwidgets.cfg
```

The following rules were added:

**v1.82**

* asctime_rCalled
* bcmpCalled
* bcopyCalled
* bsd_signalCalled
* bzeroCalled
* cryptCalled
* ctermidCalled
* ctime_rCalled
* ecvtCalled
* fcvtCalled
* fgetgrentCalled
* fgetpwentCalled
* fgetspentCalled
* gcvtCalled
* getcontextCalled
* getgrentCalled
* getgrgidCalled
* getgrnamCalled
* gethostbyaddrCalled
* gethostbyname2Called
* gethostentCalled
* getloginCalled
* getnetbyaddrCalled
* getnetbynameCalled
* getnetentCalled
* getnetgrentCalled
* getprotobynameCalled
* getprotoentCalled
* getpwentCalled
* getpwnamCalled
* getpwuidCalled
* getrpcbynameCalled
* getrpcbynumberCalled
* getrpcentCalled
* getsCalled
* getservbynameCalled
* getservbyportCalled
* getserventCalled
* getspentCalled
* getspnamCalled
* getwdCalled
* indexCalled
* lstrcatCalled
* makecontextCalled
* mktempCalled
* pthread_attr_getstackaddrCalled
* pthread_attr_setstackaddrCalled
* rand_rCalled
* readdirCalled
* rindexCalled
* scalbCalled
* sgetspentCalled
* std::asctimeCalled
* std::strtokCalled
* swapcontextCalled
* tempnamCalled
[...]

PR #1448, opened 2018-04-03
remove duplicates in sonar-cxx-plugin/src/tools/

ivangalkin:
```bash
diff  sonar-cxx-plugin/src/tools/ cxx-sensors/src/tools
Only in cxx-sensors/src/tools: clangsa_createrules.py.
```

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1448)


ivangalkin:
I've noticed there is an 1:1 duplication while working on #1445

Issue #1442, opened 2018-03-27
clangsa report is not uploaded into sonar because of NullPointerException


fukanchik:
### Description
clangsa report is not uploaded into sonar. Enabling debug does not produce full stack trace in sonar-runner.

### Steps to reproduce the problem

1. Step A
Add `sonar.cxx.clangsa.reportPath=cpp/src/analyzer_reports/*/*.plist`

2. Step B
clang version 3.5.1
Run a build with `scan-build -plist --intercept-first --analyze-headers -o analyzer_reports make`
Run `sonar-runner -X -Dsonar.verbose=true -Dsonar.log.level=TRACE`
I also tried latest `sonar-scanner` with the same result: NPE+no stack trace even with DEBUG enabled.

### Expected behavior
clangsa report is uploaded to sonar

### Actual behavior

NullPointerException happens and report is not uploaded

### Known workarounds

I do not know workarounds

### LOG file

sonar-runner log:
```
11:02:51.394 INFO  - Processing report '/tmp/cpp/src/analyzer_reports/scan-build-2018-03-27-11-59-21-275517-dlfe8oko/report-8kay5uvd.plist'
11:02:51.394 DEBUG - Processing clangsa report 'report-8kay5uvd.plist''
11:02:51.424 ERROR - Cannot feed the data into sonar, details: 'java.lang.NullPointerException'
11:02:51.424 INFO  - Sensor c++ ClangSASensor [cxx] (done) | time=31ms
```
sonar log:
no NPE in sonar logs

### Related information

* cxx plugin version: 0.9.8
* SonarQube version: 6.5

guwirth:
@fukanchik first question is always if clangsa rules are activated?

guwirth:
@fukanchik second question is why not using 6.7 LTS?

guwirth:
@fukanchik is there no exception stacktrace in the log file to get more details?
https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Get-debug-information

Maybe you can provide the stacktrace?

fukanchik:
> @fukanchik first question is always if clangsa rules are activated?

I followed instructions from wiki pages `Installation` and `Running-the-analysis`. Do I need to do something else?

How do I distinguish `Clang Static Analyzer` rules from the list of 3712 rules in the sonar server? This page does not say anything about rules: https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Code-checkers#clang-static-analyzer

[...]

PR #1449, opened 2018-04-03
handling clangsa report errors

fukanchik:
Exception stack trace is not printed on this line.

Increasing verbosity and adding other debug options won't help producing a stack trace because this catch block by itself is not expected to produce stack trace.

Next `CxxUtils.validateRecovery()` could print the stack trace,  however on [this line](https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxUtils.java#L122) 
```
throw new IllegalStateException(ex.getMessage(), ex.getCause());
```
`ex.getCase()` tries to get cause of NPE instead of NPE itself, which of course is missing since NPE in this case if the original exception. Therefore even with `sonar.cxx.errorRecoveryEnabled=false` NPE stack trace is not printed

Error message before:
```
14:16:53.983 ERROR - Cannot feed the data into sonar, details: 'java.lang.NullPointerException'
```

Error message after this change applied:
```
14:18:11.336 ERROR - Cannot feed the data into sonar, details: 'java.lang.NullPointerException'
java.lang.NullPointerException: null
        at org.sonar.cxx.sensors.clangsa.CxxClangSASensor.processReport(CxxClangSASensor.java:91)
        at org.sonar.cxx.sensors.utils.CxxReportSensor.executeReport(CxxReportSensor.java:140)
        at org.sonar.cxx.sensors.utils.CxxReportSensor.execute(CxxReportSensor.java:99)
        at org.sonar.scanner.sensor.SensorWrapper.analyse(SensorWrapper.java:53)
        at org.sonar.scanner.phases.SensorsExecutor.executeSensor(SensorsExecutor.java:88)
        at org.sonar.scanner.phases.SensorsExecutor.execute(SensorsExecutor.java:82)
        at org.sonar.scanner.phases.SensorsExecutor.execute(SensorsExecutor.java:68)
        at org.sonar.scanner.phases.AbstractPhaseExecutor.execute(AbstractPhaseExecutor.java:88)
        at org.sonar.scanner.scan.ModuleScanContainer.doAfterStart(ModuleScanContainer.java:180)
        at org.sonar.core.platform.ComponentContainer.startComponents(ComponentContainer.java:135)
        at org.sonar.core.platform.ComponentContainer.execute(ComponentContainer.java:121)
        at org.sonar.scanner.scan.ProjectScanContainer.scan(ProjectScanContainer.java:288)
        at org.sonar.scanner.scan.ProjectScanContainer.scanRecursively(ProjectScanContainer.java:283)
        at org.sonar.scanner.scan.ProjectScanContainer.doAfterStart(ProjectScanContainer.java:261)
        at org.sonar.core.platform.ComponentContainer.startComponents(ComponentContainer.java:135)
        at org.sonar.core.platform.ComponentContainer.execute(ComponentContainer.java:121)
        at org.sonar.scanner.task.ScanTask.execute(ScanTask.java:48)
        at org.sonar.scanner.task.TaskContainer.doAfterStart(TaskContainer.java:84)
        at org.sonar.core.platform.ComponentContainer.startComponents(ComponentContainer.java:135)
        at org.sonar.core.platform.ComponentContainer.execute(ComponentContainer.java:121)
        at org.sonar.scanner.bootstrap.GlobalContainer.executeTask(GlobalContainer.java:121)
        at org.sonar.batch.bootstrapper.Batch.doExecuteTask(Batch.java:116)
        at org.sonar.batch.bootstrapper.Batch.execute(Batch.java:71)
        at org.sonar.runner.batch.IsolatedLauncher.execute(IsolatedLauncher.java:48)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.sonar.runner.impl.BatchLauncher$1.delegateExecution(BatchLauncher.java:87)
        at org.sonar.runner.impl.BatchLauncher$1.run(BatchLauncher.java:75)
        at java.security.AccessController.doPrivileged(Native Method)
        at org.sonar.runner.impl.BatchLauncher.doExecute(BatchLauncher.java:69)
        at org.sonar.runner.impl.BatchLauncher.execute(BatchLauncher.java:50)
        at org.sonar.runner.api.EmbeddedRunner.doExecute(EmbeddedRunner.java:102)
        at org.sonar.runner.api.Runner.execute(Runner.java:100)
        at org.sonar.runner.Main.executeTask(Main.java:70)
        at org.sonar.runner.Main.execute(Main.java:59)
        at org.sonar.runner.Main.main(Main.java:53)
```

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1449)

### DIFF
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensor.java
@@ -77,6 +77,8 @@ public class CxxClangSASensor extends CxxReportSensor {
 
       NSDictionary rootDict = (NSDictionary) PropertyListParser.parse(f);
 
+      final String clangVersion = ((NSString) rootDict.get("clang_version")).getContent();
+
       // Array of file paths where an issue was detected.
       NSObject[] sourceFiles = ((NSArray) rootDict.objectForKey("files")).getArray();
 
@@ -88,7 +90,12 @@ public class CxxClangSASensor extends CxxReportSensor {
         NSString desc = (NSString) diag.get("description");
         String description = desc.getContent();
 
-        String checkerName = ((NSString) diag.get("check_name")).getContent();
+        final NSObject checkNameValue = diag.get("check_name");
+        if(checkNameValue == null) {
+            LOG.error("clang version > 3.7.0 is required. Your version is: " + clangVersion);
+            return;
+        }
+        String checkerName = ((NSString) checkNameValue).getContent();
 
         NSDictionary location = (NSDictionary) diag.get("location");
 
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportSensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportSensor.java
@@ -124,7 +124,7 @@ public abstract class CxxReportSensor implements Sensor {
         .append(e)
         .append("'")
         .toString();
-      LOG.error(msg);
+      LOG.error(msg, e);
       CxxUtils.validateRecovery(e, this.language);
     }
   }
--- a/cxx-sensors/src/test/resources/org/sonar/cxx/sensors/reports-project/clangsa-reports/clangsa-wrong-version.plist
+++ b/cxx-sensors/src/test/resources/org/sonar/cxx/sensors/reports-project/clangsa-reports/clangsa-wrong-version.plist
@@ -1,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+ <key>clang_version</key>
+<string>clang version 3.5.1 (tags/RELEASE_351/final)</string>
+ <key>files</key>
+ <array>
+  <string>src/cxx-library-sample/foo.cpp</string>
+ </array>
+ <key>diagnostics</key>
+ <array>
+  <dict>
+   <key>path</key>
+   <array>
+    <dict>
+     <key>kind</key><string>event</string>
+     <key>location</key>
+     <dict>
+      <key>line</key><integer>17</integer>
+      <key>col</key><integer>11</integer>
+      <key>file</key><integer>0</integer>
+     </dict>
+     <key>ranges</key>
+     <array>
+       <array>
+        <dict>
+         <key>line</key><integer>17</integer>
+         <key>col</key><integer>11</integer>
+         <key>file</key><integer>0</integer>
+        </dict>
+        <dict>
+         <key>line</key><integer>17</integer>
+         <key>col</key><integer>11</integer>
+         <key>file</key><integer>0</integer>
+        </dict>
+       </array>
+       <array>
+        <dict>
+         <key>line</key><integer>17</integer>
+         <key>col</key><integer>15</integer>
+         <key>file</key><integer>0</integer>
+        </dict>
+        <dict>
+         <key>line</key><integer>17</integer>
+         <key>col</key><integer>28</integer>
+         <key>file</key><integer>0</integer>
+        </dict>
+       </array>
+     </array>
+     <key>depth</key><integer>0</integer>
+     <key>extended_message</key>
+     <string>Value stored to &apos;x&apos; during its initialization is never read</string>
+     <key>message</key>
+     <string>Value stored to &apos;x&apos; during its initialization is never read</string>
+    </dict>
+   </array>
+   <key>description</key><string>Value stored to &apos;x&apos; during its initialization is never read</string>
+   <key>category</key><string>Dead store</string>
+   <key>type</key><string>Dead initialization</string>
+  <key>issue_hash</key><string>1</string>
+  <key>location</key>
+  <dict>
+   <key>line</key><integer>17</integer>
+   <key>col</key><integer>11</integer>
+   <key>file</key><integer>0</integer>
+  </dict>
+  </dict>
+ </array>
+</dict>
+</plist>
--- a/cxx-sensors/src/test/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensorTest.java
+++ b/cxx-sensors/src/test/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensorTest.java
@@ -92,6 +92,21 @@ public class CxxClangSASensorTest {
   }
 
   @Test
+  public void clangVersionIsTooOld() {
+    SensorContextTester context = SensorContextTester.create(fs.baseDir());
+
+    settings.setProperty(language.getPluginProperty(CxxClangSASensor.REPORT_PATH_KEY), "clangsa-reports/clangsa-wrong-version.plist");
+    context.setSettings(settings);
+
+    context.fileSystem().add(TestInputFileBuilder.create("ProjectKey", "src/cxx-library-sample/foo.cpp")
+      .setLanguage("cpp").initMetadata(new String("asd\nasdas\nasda\n")).build());
+
+    CxxClangSASensor sensor = new CxxClangSASensor(language);
+    sensor.execute(context);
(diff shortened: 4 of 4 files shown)


guwirth:
@fukanchik thanks a lot for investigation.

Looking to the stack trace 
```
 at org.sonar.cxx.sensors.clangsa.CxxClangSASensor.processReport(CxxClangSASensor.java:91)
```
The problem is here:
https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensor.java#L91
```Java
 String checkerName = ((NSString) diag.get("check_name")).getContent();
```
* The method `NSDictionary ge`t can return also `null`. Seems that `check_name` is missing for one of the issues?
* There is also no catch for such kind of error: https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensor.java#L109

In case of a problem `validateRecovery `https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxUtils.java#L116 descides if analysis should continue or break.

Think https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensor.java#L91 have to be fixed to handle the `null ` `get`.

Maybe https://github.com/SonarOpenCommunity/sonar-cxx/pull/1447/files#diff-e2fa7d302186280fa2ad45873a2294baR79 `Objects.requireNonNull` is a solution?

fukanchik:
As I described in [this comment](https://github.com/SonarOpenCommunity/sonar-cxx/issues/1442#issuecomment-378389788) `check_name` was introduced in `LLVM 3.7`. Therefore running clangsa sensor is useless with earlier versions of LLVM.

In my opinion the reasonable behaviour of clangsa sensor in case of older LLVM versions is to log explanation message to error log and skip processing. This is what was done in 778a5b4d8c4dba7977a977014d9f12f7b45abffa.

Indeed `Objects.requireNonNull` might be a better choice. This would require setting at least `<java.version>1.6</java.version>` in `maven-compiler-plugin` though.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportSensor.java
```diff
@@ -124,7 +124,7 @@ public void execute(SensorContext context) {
         .append(e)
         .append("'")
         .toString();
-      LOG.error(msg);
+      LOG.error(msg, e);
```
guwirth: No need for this. There is the `.append(e)` why again in `.error`?

cxx-sensors/src/test/resources/org/sonar/cxx/sensors/reports-project/clangsa-reports/clangsa-wrong-version.plist
```diff
@@ -0,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
guwirth: Could you shorten this sample to the minimum please

cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensor.java
```diff
@@ -77,6 +77,8 @@ protected void processReport(final SensorContext context, File report)
 
       NSDictionary rootDict = (NSDictionary) PropertyListParser.parse(f);
 
+      final String clangVersion = ((NSString) rootDict.get("clang_version")).getContent();
```
guwirth: Why not checking for the clang version directly here?

cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensor.java
```diff
@@ -88,7 +90,12 @@ protected void processReport(final SensorContext context, File report)
         NSString desc = (NSString) diag.get("description");
         String description = desc.getContent();
 
-        String checkerName = ((NSString) diag.get("check_name")).getContent();
```
guwirth: Would add for all `.get` a `Objects.requireNonNull` check (see https://github.com/SonarOpenCommunity/sonar-cxx/pull/1447/files#diff-e2fa7d302186280fa2ad45873a2294baR79)

cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensor.java
```diff
@@ -88,7 +90,12 @@ protected void processReport(final SensorContext context, File report)
         NSString desc = (NSString) diag.get("description");
         String description = desc.getContent();
 
-        String checkerName = ((NSString) diag.get("check_name")).getContent();
+        final NSObject checkNameValue = diag.get("check_name");
+        if(checkNameValue == null) {
+            LOG.error("clang version > 3.7.0 is required. Your version is: " + clangVersion);
```
guwirth: Would remove this error, see comment above

cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportSensor.java
guwirth: add also a check to  `((NSString) filePath).getContent()`

guwirth: Catch requireNonNull exception

guwirth:
Hi @fukanchik I added some comments. Would be perfect if you could rework it. Thx

fukanchik:
Thanks, I will have a look

cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensor.java
fukanchik: I thought about this. Checking version would only work for stock builds of clang. Imagine someone builds custom version of clang with the version string like "rh-3.7.1-internal". For such compiler checking version would not work.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensor.java
fukanchik: Checking for presence of `check_name` allows to parse output of custom-named internal builds of clang like `prefix-version-suffix`. I'm ok to check the version, this makes situation clearer.

However you should make this decision: support only names with numbers in front or check for existence of `check_name` instead.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportSensor.java
fukanchik: I commented about this earlier [here](https://github.com/SonarOpenCommunity/sonar-cxx/issues/1442#issuecomment-378314000). In situation when any _[unexpected](https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensor.java#L109)_ exception happens stack trace is never printed. Because `.append(e)` only appends error message without stack trace. Same happens here in `LOG.error(msg)` which again only prints error message without stack trace. Note that `LOG.error(msg)` was there before me. Increasing verbosity and enabling debug does not help here, I had to modify this line in my custom plugin to see what is going on.

cxx-sensors/src/test/resources/org/sonar/cxx/sensors/reports-project/clangsa-reports/clangsa-wrong-version.plist
fukanchik: ok

PR #1456, opened 2018-04-11
introduce HTML tidy checks for HTML descriptions

ivangalkin:
* create check_rules.sh and extend cppcheck_createrules.py

```python
for each rule.xml in cxx-sensors/main/resources:
    check rule.xml with xmllint
    parse rule.xml
        for each rule in rule.xml:
            dump rule.attr("description") to tmp.ruledump
            check tmp.ruledump with tiny-html5
```

* fix issues w.r.t. HTML formatting and escaping
* many important HTML issues were fixed, but also many [escaping] issues which SonarQube seems to tolerate

@guwirth is there a chance to activate check_rules.sh in continuous integration? FYI it uses [linux] tools (xmllint and tidy-html5), which probably have to installed additionally.

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1456)

### DIFF
--- a/cxx-sensors/src/tools/cppcheck_createrules.py
+++ b/cxx-sensors/src/tools/cppcheck_createrules.py
@@ -25,6 +25,7 @@
 
 import sys
 import os
+import subprocess
 import xml.etree.ElementTree as et
 import textwrap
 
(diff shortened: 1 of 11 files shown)


guwirth:
> is there a chance to activate check_rules.sh in continuous integration? FYI it uses [linux] tools (xmllint and tidy-html5), which probably have to installed additionally.

@ivangalkin we are using Travis for testing with Linux. The entry point is the file https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/.travis.yml. You can try to add your script in the `script:` section. In https://travis-ci.org/SonarOpenCommunity/sonar-cxx you see the LOG files.

Not sure if `xmllint `and `tidy-html5` is pre-installed (think not). In this case you have to install it in one of the `before_script`, `install `or `before_script` steps.

cxx-sensors/src/tools/cppcheck_createrules.py
```diff
@@ -203,8 +204,77 @@ def makeDescriptionToCDATA(rule_t):
...
+        print err
+        print "### OUT"
+        print out
+        print "\n"
+        return True
+    return False
+
+
+def call_tidy(file_path):
+    command = ["/home/igalkin/workspace_myGITHUB/tidy-html5/build/cmake/tidy", file_path]
```
guwirth: script is using a local path from your PC

ivangalkin:
@guwirth thank you for the comments and the information about travis
since the change is already pretty large I would limit its scope to the XML adjustment
I will address the CI integration in the future pull request

guwirth:
@Bertk ready to merge?

guwirth:
@Bertk any comments. Will merge it this evening...

PR #1464, opened 2018-04-24
CoberturaParser: extract loop invariant

ivangalkin:
before:

```python
  for each package:
     for each file in package:
         for each class in file:
             for each branch in class:
                 Pattern.compile("\\((.*?)\\)")
```

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1464)


ivangalkin:
@guwirth The behavior was not changed: Pattern class is immutable (see [javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html)). There is no sense to compile the same pattern in loop, compiled `Pattern` can be used for multiple matches

PR #1481, opened 2018-05-23
Prepare for next development iteration

guwirth:
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1481)

### DIFF
--- a/cxx-checks/pom.xml
+++ b/cxx-checks/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.sonarsource.sonarqube-plugins.cxx</groupId>
     <artifactId>cxx</artifactId>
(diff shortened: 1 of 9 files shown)


ivangalkin:
0.9.8 -> 0.9.9 -> 1.0.0 -> **1.1.0**? It doesn't look correct to me.

guwirth:
@ivangalkin in the past we were very careful increasing the version number. There was always the question: when is it finished - when should we use 1.0?

**major.minor[.maintenance[.build]]**

Because we have no roadmap it's always hard to say if there are only bugfixes (**maintenace**) or also features in a version (**major.minor)**. During the next time I like to increase the **minor** version with every release.

Issue #1465, opened 2018-04-24
coverage cache vs multi-module projects


ivangalkin:
Hi All

I believe there is a severe design problem in the coverage sensors. It is observable if 
a)  multi-module project is analyzed plus
b)  coverage report (e.g. the cobertura one) contains relative paths (see below, `<source>` is `.`)

If path is relative, `SensorContext::fileSystem().baseDir().getAbsolutePath()` is used as a base directory while parsing. After report is parsed  the corresponding `CoverageMeasures` structure is cached. For the next sub-project the base directory should be actually recalculated, but it isn't.

I believe the cache can be used only if all paths are absolute. This design idea applicable to all sensors. If we don't check/force this fact, it's not safe to cache any reports (and we had a similar discussion in  #1423).

The correct solution in my opinion should be

a. Terminate if coverage report of a if multi-module project contains relative paths and/or
b. Provide a property in order disable the cache

I know, that the coverage output can be customized, but the current implementation is a pitfall and must be fixed or at least documented IMHO.

```XML
<coverage ... version="gcovr 3.4">
  <sources><source>.</source></sources>
  <packages>
     <package>
        <classes><class filename="relativePath"> ... <class/>
        </classes>
    </package>
 </packages>
</coverage>
```

Best regards,

guwirth:
Hi @ivangalkin, instead of fixing the sensor I think it's the better way to use the generic test and coverage sensors in future: https://docs.sonarqube.org/display/SONAR/Generic+Test+Data. In our plugin we can provide a transformation mechanism and the XLST.

PR #1486, opened 2018-05-30
 Add coverage sensor 'Testwell CTC++' for textural reports

rufinio:
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1486)

### DIFF
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/CxxCoverageSensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/CxxCoverageSensor.java
@@ -67,6 +67,7 @@ public class CxxCoverageSensor extends CxxReportSensor {
     parsers.add(new CoberturaParser());
     parsers.add(new BullseyeParser());
     parsers.add(new VisualStudioParser());
+    parsers.add(new TestwellCtcParser());
   }
 
   @Override
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/CxxCoverageSensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/CxxCoverageSensor.java
@@ -67,7 +67,7 @@ public class CxxCoverageSensor extends CxxReportSensor {
     parsers.add(new CoberturaParser());
     parsers.add(new BullseyeParser());
     parsers.add(new VisualStudioParser());
-    parsers.add(new TestwellCtcParser());
+    parsers.add(new TestwellCtcTxtParser());
   }
 
   @Override
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/TestwellCtcParser.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/TestwellCtcParser.java
@@ -39,7 +39,6 @@ import static org.sonar.cxx.sensors.coverage.TestwellCtcResult.FILE_HEADER;
 import static org.sonar.cxx.sensors.coverage.TestwellCtcResult.FILE_RESULT;
 import static org.sonar.cxx.sensors.coverage.TestwellCtcResult.LINE_RESULT;
 import static org.sonar.cxx.sensors.coverage.TestwellCtcResult.SECTION_SEP;
-import static org.sonar.cxx.sensors.coverage.TestwellCtcResult.REPORT_FOOTER;
 
 
 
@@ -92,8 +91,8 @@ public class TestwellCtcParser extends CxxCoverageParser {
     } catch (FileNotFoundException e) {
       LOG.warn("TestwellCtcParser file not found '{}'", e.getMessage());
     }
-
   }
+
   private void parseReportHead() {
     try {
       if (!matcher.reset(scanner.next()).find()) {
@@ -105,7 +104,7 @@ public class TestwellCtcParser extends CxxCoverageParser {
       LOG.debug("'Testwell CTC++' file section not found!");
     }
   }
-  
+
   private void parseUnit(final Map<String, CoverageMeasures> coverageData) {
     LOG.debug(matcher.toString());
     
@@ -116,7 +115,7 @@ public class TestwellCtcParser extends CxxCoverageParser {
       scanner.close();
     }
   }
-  
+
   private void parseFileUnit(final Map<String, CoverageMeasures> coverageData) {
     LOG.debug("Parsing file section...");
     
@@ -134,7 +133,7 @@ public class TestwellCtcParser extends CxxCoverageParser {
     addLines(file, coverageData);
     matcher.reset(scanner.next());
   }
-  
+
   private void addLines(File file, final Map<String, CoverageMeasures> coverageData) {
     LOG.debug("Parsing function sections...");
 
@@ -144,7 +143,7 @@ public class TestwellCtcParser extends CxxCoverageParser {
     }
     coverageData.put(file.getPath(), coverageMeasures);
   }
-  
+
   private void parseLineSection(CoverageMeasures coverageMeasures) {
     LOG.debug("Found line section...");
     
@@ -154,12 +153,11 @@ public class TestwellCtcParser extends CxxCoverageParser {
       LOG.warn("Neither File Result nor Line Result after FileHeader!");
     }
   }
-  
+
   private void addEachLine(CoverageMeasures coverageMeasures) {
 
     int lineIdCur;
     int lineIdPrev;
-    int lineIdNext;
     int lineHits;
     int conditions;
     int coveredConditions;
@@ -167,29 +165,29 @@ public class TestwellCtcParser extends CxxCoverageParser {
     int lineHitsTrue;
     int lineHitsFalse;
     boolean conditionDetected;
-    
+
     lineIdPrev = 0;
     conditions = 0;
     coveredConditions = 0;
     lineIdCond = 0;
     conditionDetected = false;
     lineHits = 0;
-    
+
     do {
       lineIdCur = Integer.parseInt(matcher.group(LINE_NR_GROUP));
-     
+
       String condsTrue = matcher.group(CONDS_TRUE);
       String condsFalse = matcher.group(CONDS_FALSE);
-      
+
       if ((condsTrue != null) || (condsFalse != null)) {
-        
+
         lineHitsTrue = (condsTrue != null ? new BigDecimal(condsTrue).intValue() : 0);
         lineHitsFalse = (condsFalse != null ? new BigDecimal(condsFalse).intValue() : 0);
         lineHits = lineHitsTrue + lineHitsFalse;
-        
+
         if (lineIdPrev != lineIdCur) {
           coverageMeasures.setHits(lineIdCur, lineHits);
-          
+
           if (lineIdCond > 0) {
             coverageMeasures.setConditions(lineIdCond, conditions, coveredConditions);
             lineIdCond = 0;
@@ -197,7 +195,7 @@ public class TestwellCtcParser extends CxxCoverageParser {
             coveredConditions = 0;
             conditionDetected = false;
           }
-          
+
           if ((condsTrue != null) && (condsFalse != null)) {
             // suppose single condition
             lineIdCond = lineIdCur;
@@ -207,7 +205,7 @@ public class TestwellCtcParser extends CxxCoverageParser {
           }
         } else {
           // multicondition
-          
+
           if (conditionDetected == true) {
             // reset supposed single condition
             conditions = 0;
@@ -222,28 +220,40 @@ public class TestwellCtcParser extends CxxCoverageParser {
         }
       } else {
         // Parse information for statement coverage needed in decising the line coverage
-        String blockEnd = matcher.group(4);
-        if (blockEnd != null) {
-          if (blockEnd.endsWith("-")) {
-            lineHits = 0;
-          } else if (blockEnd.endsWith("+")) {
-            lineHits = 1;
-          } else {
-            lineHits = 0;
-            LOG.warn("Undefined information for statement coverage!");
-          }
-          coverageMeasures.setHits(lineIdCur, lineHits);
-        }
+        setLinehitsByBlockend(coverageMeasures, lineIdCur);
       }
 
-      if (lineIdPrev > 0) {
-        lineIdNext = lineIdPrev + 1;
-        while (lineIdNext < lineIdCur) {
-          coverageMeasures.setHits(lineIdNext, lineHits);
-          lineIdNext++;
-        }
-      }
+      setLinehits(coverageMeasures, lineIdPrev, lineIdCur, lineHits);
       lineIdPrev = lineIdCur;
     } while (matcher.find());
+  }
+
+  private void setLinehitsByBlockend(CoverageMeasures coverageMeasures, int lineIdCur) {
+    int lineHits;
+    String blockEnd = matcher.group(4);
+
+    if (blockEnd != null) {
+      if (blockEnd.endsWith("-")) {
+        lineHits = 0;
+      } else if (blockEnd.endsWith("+")) {
+        lineHits = 1;
+      } else {
+        lineHits = 0;
+        LOG.warn("Undefined information for statement coverage!");
+      }
+      coverageMeasures.setHits(lineIdCur, lineHits);
+    }
+  }
+  
+  private void setLinehits(CoverageMeasures coverageMeasures, int lineIdPrev, int lineIdCur, int lineHits) {
+    int lineIdNext;
+
+    if (lineIdPrev > 0) {
+      lineIdNext = lineIdPrev + 1;
+      while (lineIdNext < lineIdCur) {
+        coverageMeasures.setHits(lineIdNext, lineHits);
+        lineIdNext++;
+      }
+    }
   }
 }
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/TestwellCtcParser.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/TestwellCtcParser.java
@@ -111,7 +111,7 @@ public class TestwellCtcParser extends CxxCoverageParser {
     
     if (matcher.usePattern(FILE_HEADER).find(FROM_START)) {
       parseFileUnit(coverageData);
-    } else if (matcher.usePattern(REPORT_FOOTER).find(FROM_START)) {
+    } else {
       state = State.END;
       scanner.close();
     }
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/TestwellCtcParser.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/TestwellCtcParser.java
@@ -174,6 +174,7 @@ public class TestwellCtcParser extends CxxCoverageParser {
     coveredConditions = 0;
     lineIdCond = 0;
     conditionDetected = false;
+    lineHits = 0;
     
     do {
       lineIdCur = Integer.parseInt(matcher.group(LINE_NR_GROUP));
@@ -237,9 +238,6 @@ public class TestwellCtcParser extends CxxCoverageParser {
             LOG.warn("Undefined information for statement coverage!");
           }
           coverageMeasures.setHits(lineIdCur, lineHits);
-        }
-        else {
-          lineHits = 0;
         }
       }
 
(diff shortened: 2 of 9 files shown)


guwirth:
@rufinio thx for providing this extension. 

Could you provide some information please. 

1. A link to the tool page.  
2. How to call / use the tool to generate a report for the cxx plugin?
3. Wondering why you are parsing an ASCII report? Does the tool not provide XML?

Did not have the time to do a code review yet. Points should work:

- Handle empty reports. 
- Support of https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Supported-configuration-properties#sonarcxxerrorrecoveryenabled

cxx-sensors/src/test/java/org/sonar/cxx/sensors/coverage/CxxTestwellCtcParserTest.java
```diff
@@ -0,0 +1,96 @@
...
+  public void shouldReportCorrectCoverage() {
+    context = SensorContextTester.create(fs.baseDir());
+    settings.setProperty(language.getPluginProperty(CxxCoverageSensor.REPORT_PATH_KEY), "coverage-reports/TestwellCTC/report_small_v8.txt");
+    context.setSettings(settings);
+    context.fileSystem().add(TestInputFileBuilder.create("ProjectKey", "HGBuildNumberLookup.cpp")
+      .setLanguage("cpp").initMetadata("\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n").build());
+    sensor = new CxxCoverageSensor(new CxxCoverageCache(), language, context);
+    sensor.execute(context);
+    
+    assertThat(context.lineHits("ProjectKey:HGBuildNumberLookup.cpp", 42)).isEqualTo(10);
```
Bertk: Please use one assertThat statement per test.  You can also use SoftAssertions (see http://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/SoftAssertions.html)

cxx-sensors/src/test/java/org/sonar/cxx/sensors/coverage/CxxTestwellCtcParserTest.java
```diff
@@ -0,0 +1,96 @@
...
+    settings.setProperty(language.getPluginProperty(CxxCoverageSensor.REPORT_PATH_KEY), "coverage-reports/TestwellCTC/report_big.txt");
+    context.setSettings(settings);
+    context.fileSystem().add(TestInputFileBuilder.create("ProjectKey", "test-wildmatch.c")
+      .setLanguage("cpp").initMetadata("\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n").build());
+    context.fileSystem().add(TestInputFileBuilder.create("ProjectKey", "credential-store.c")
+      .setLanguage("cpp").initMetadata("\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n").build());
+    sensor = new CxxCoverageSensor(new CxxCoverageCache(), language, context);
+    sensor.execute(context);
+    
+    assertThat(context.lineHits("ProjectKey:test-wildmatch.c", 3)).isEqualTo(209);
```
Bertk: Please use one assertThat statement per test.  You can also use SoftAssertions (see http://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/SoftAssertions.html)

cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/TestwellCtcParser.java
```diff
@@ -0,0 +1,256 @@
...
+    LOG.debug("Found line section...");
+    
+    if (matcher.usePattern(LINE_RESULT).find(FROM_START)) {
+      addEachLine(coverageMeasures);
+    } else  {
+      LOG.warn("Neither File Result nor Line Result after FileHeader!");
+    }
+  }
+  
+  private void addEachLine(CoverageMeasures coverageMeasures) {
```
Bertk: Please reduce complexity of this method (e.g. extract methods). The Cognitive Complexity of this method is 57.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/TestwellCtcParser.java
```diff
@@ -0,0 +1,256 @@
...
+    
+    String normalFilename;
+    String filename;
+    File file;
+    
+    filename = matcher.group(1);
+    if((new File(filename)).isAbsolute() == false) {
+      normalFilename = FilenameUtils.normalize("./" + filename);
+    }
+    else {
```
Bertk: Style: Move this "else" on the same line that the previous closing curly brace (4 times in this file)

Please align with the current style.

rufinio:
### A link to the tool page.
https://www.verifysoft.com/en_ctcpp.html


### How to call / use the tool to generate a report for the cxx plugin?
Adapted from Testwell CTC++ User's Guid (Basic Use of CTC++)
Suppose we have the following five source files that form a complete program: prime.c io.c calc.c. This program can be compiled and linked in many ways, one way being the following 
`cl –Feprime.exe prime.c io.c calc.c` 
which results in the prime.exe program.
Now, we wish to apply CTC++ on our program, that is, we want to measure the files prime.c, io.c and calc.c and find out how thoroughly they were exercised in our test runs. First we need to instrument the files we wish to measure. Assume we wish to measure multicondition coverage. This can be done as follows: 
`ctc -i m cl –Feprime.exe prime.c io.c calc.`c 
As a result we get the instrumented prime.exe program. Here 'ctc' is the CTC++ Preprocessor utility, which makes the instrumentation on the given C and C++ source files and drives compiling/linking of the new instrumented target. The '-i m' command-line options to ctc mean "instrument for multicondition". 
when ctc instruments source files, it maintains descriptions what the files contain (what interesting code there is to ctc, on what lines, etc.). This file is called symbolfile, and when it is not specified (like here) it will be MON.sym in current directory.
All right, now prime.exe is the instrumented version of the program. Let's do one test run.
After this test run we notice that the file MON.dat has born in the current directory (same directory as the symbolfile MON.sym was created to). It is a datafile, containing the collected execution counters when the code in the instrumented files was executed.
Now we wish to see the results of our test, i.e. what parts of the program the above run has executed. We use the CTC++ Postprocessor utility as follows: 
`ctcpost MON.sym MON.dat -p profile.txt `
What we are asking here is that ctcpost takes the symbolfile MON.sym and the datafile MON.dat as input and produces an Execution Profile Listing to the file profile.txt.
To get the results of our test to sonarqube we add the follwoing line to configuration file _sonar-project.properties_:
`sonar.cxx.coverage.reportPath=profile.txt`

### Wondering why you are parsing an ASCII report? Does the tool not provide XML?
Testwell CTC++ also provides a XML output as well as JSON output. I mainly use TXT and therefore I know better about it than XML.


### Handle empty reports.
Tested by Test shouldConsumeEmptyReport

### Support of sonar.cxx.errorRecoveryEnabled
I wonder if that is applicable for this extension: In my optinion invalid file references or line numbers has to be handled by calling layer (CxxCoverageSensor).
In case of errors in a report file: This extension can't deside between tolerant and strict mode. No data and corrupt data have the same effect that regex doesn't match and no coverage data is collected.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/TestwellCtcParser.java
rufinio: Can you recommend a tool for Cognitive Complexity?
Which value range is good?

cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/TestwellCtcParser.java
```diff
@@ -0,0 +1,259 @@
...
+  public TestwellCtcParser() {
+    // no operation but necessary for list of coverage parsers 
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void processReport(final SensorContext context, File report, final Map<String, CoverageMeasures> coverageData) {
+    LOG.debug("Parsing 'Testwell CTC++' format");
```
guwirth: To make clear that ‚only‘ textual format is supported:
Parsing 'Testwell CTC++' **textual** format

guwirth:
@rufinio thx for your answer. Which version of the tool are you using / testing? What should we write in the docu? Supporting version x and later...

guwirth:
@rufinio found also this page http://www.testwell.fi/ctcdesc.html. Is the tool from Testwell or Verisoft?

cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/CxxCoverageSensor.java
```diff
@@ -67,6 +67,7 @@ public CxxCoverageSensor(CxxCoverageCache cache, CxxLanguage language, SensorCon
     parsers.add(new CoberturaParser());
     parsers.add(new BullseyeParser());
     parsers.add(new VisualStudioParser());
+    parsers.add(new TestwellCtcParser());
```
guwirth: Maybe better TestwellCtcTxtParser?

ivangalkin:
@rufinio Technical feedback: I am not sure, if your parser fits into the existing framework. Please see the `CxxCoverageSensor::processReports()`: it iterates over all registered `CoverageParser`s and gives each of them a chance to process the coverage report. If parser doesn't recognize the format it must either throw an exception `XMLStreamException`/`EmptyReportException` or produce an empty map `Map<String, CoverageMeasures>` (please see  `CxxCoverageSensor::parseCoverageReport()`). Not sure about the map emptiness, but your parser never throws as far as I see. That's suspicious. Please consider using `EmptyReportException` as a signal, that the file format was not recognized.

@guwirth  w.r.t. #1465 and [the generic coverage format](https://docs.sonarqube.org/display/SONAR/Generic+Test+Data), wouldn't it be more foresighted to create an external (?) converter CTC++ -> XML and import the coverage by means of `sonar.coverageReportPaths`?

ivangalkin:
one more side remark: TestwellCtcParser.java has common parts with https://github.com/Londran/sonar-ctc/blob/master/src/main/java/org/sonar/plugins/ctc/api/parser/CtcTextParser.java (among others some logging messages are 100% identical).

The original file is copyrighted

```
/*
 * Testwell CTC++ Plugin
 * Copyright (C) 2014 Verifysoft Technology GmbH
...
*/
```

Not sure if you can just skip the original copyright of LGPL3 code. Please double-check.

guwirth:
> Not sure if you can just skip the original copyright of LGPL3 code. Please double-check.

Source is LGPL and as long derivative work is again under license LGPL this should be no problem. But for later documentation it makes sense to add a comment below our header with a link to the original source.

rufinio:
@guwirth:

> Which version of the tool are you using / testing? What should we write in the docu? Supporting version x and later...

# Testwell CTC++

## Prerequisites
Testwell CTC++ Version 7.2 and later.

## 1. Generate instrumented executable
Suppose we have the following source files that form a complete program: prime.c io.c calc.c. This program can be compiled and linked in many ways, one way being the following
`cl –Feprime.exe prime.c io.c calc.c`
or
`gcc -o prime prime.c io.c calc.c`
which results in the prime.exe / prime program.
Now, we wish to apply CTC++ on our program, that is, we want to measure the files prime.c, io.c and calc.c and find out how thoroughly they were exercised in our test runs. First we need to instrument the files we wish to measure. Assume we wish to measure multicondition coverage. This can be done as follows:
`ctc -i m cl –Feprime.exe prime.c io.c calc.c`
or
`ctc -i m gcc -o prime prime.c io.c calc.c`
As a result we get the instrumented prime.exe program. Here 'ctc' is the CTC++ Preprocessor utility, which makes the instrumentation on the given C and C++ source files and drives compiling/linking of the new instrumented target. The '-i m' command-line options to ctc mean "instrument for multicondition".
when ctc instruments source files, it maintains descriptions what the files contain (what interesting code there is to ctc, on what lines, etc.). This file is called symbolfile, and when it is not specified (like here) it will be MON.sym in current directory.
All right, now prime.exe is the instrumented version of the program. Let's do one test run.

## 2. Generate coverage report
After this test run we notice that the file MON.dat has born in the current directory (same directory as the symbolfile MON.sym was created to). It is a datafile, containing the collected execution counters when the code in the instrumented files was executed.
Now we wish to see the results of our test, i.e. what parts of the program the above run has executed. We use the CTC++ Postprocessor utility as follows:
`ctcpost MON.sym MON.dat -p profile.txt`
What we are asking here is that ctcpost takes the symbolfile MON.sym and the datafile MON.dat as input and produces an Execution Profile Listing to the file profile.txt.

## 3. Read coverage file with SonarQube
To get the results of our test to sonarqube we add the following line to configuration file sonar-project.properties:
`sonar.cxx.coverage.reportPath=profile.txt`




> 
> 
> Testwell or Verifysoft

Verifysoft Technology GmbH (Offenburg/Germany) has acquired all intellectual property rights for the software test and analysis tools of Testwell (Finland).
The aquisition includes Testwell CTC++, CMT++/CMTJava und CTA++ as well as tools for the ADA programming language.
Source: https://www.verifysoft.com/en_verifysoft_acquires_testwell_ipr.html


@ivangalkin:
If scanner can't recognize file format, it's signaled by an empty map Map<String, CoverageMeasures>. You can check this behavior by running test `shouldConsumeEmptyReport`.

cxx-sensors/src/test/java/org/sonar/cxx/sensors/coverage/CxxTestwellCtcTxtParserTest.java
```diff
@@ -0,0 +1,166 @@
...
+public class CxxTestwellCtcTxtParserTest {
+
+  private CxxCoverageSensor sensor;
+  private DefaultFileSystem fs;
+  private SensorContextTester context;
+  private CxxLanguage language;
+  private final MapSettings settings = new MapSettings();
+
+  @Before
+  public void setUp() {
```
ivangalkin: It is very error-prone to share mutable state between test cases
maybe it's convenient (?) to declare ...

+  private CxxCoverageSensor sensor;
+  private DefaultFileSystem fs;
+  private SensorContextTester context;

... and permanently reset them in `@Before`...

but sharing of...

+ private final MapSettings settings = new MapSettings();

... is just wrong IMHO

```
private static MapSettings singletonMapSettings( String key, String value )
{
     MapSettings settings = new MapSettings();
     settings.set( key, value );
     return settings;
}

@Test
public void myWhateverTest() {
  MapSettings settings = singletonMapSettings( "some key", "some value );
}
```

P. S.  I grepped in other tests for `settings` and these pattern seems to be common (which still doesn't make it correct). Feel free to ignore my comment.

guwirth:
> P. S. I grepped in other tests for settings and these pattern seems to be common (which still doesn't make it correct). Feel free to ignore my comment.

@ivangalkin maybe we should fix this in an extra refactoring task for all unit tests?

guwirth:
@rufinio do you have time to fix the review comments from @ivangalkin?

cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/TestwellCtcTxtParser.java
```diff
@@ -0,0 +1,262 @@
...
+
+/**
+ * {@inheritDoc}
+ */
+public class TestwellCtcTxtParser extends CxxCoverageParser {
+
+  private static final Logger LOG = Loggers.get(TestwellCtcTxtParser.class);
+  
+  private Scanner scanner;
+  private Matcher matcher;
```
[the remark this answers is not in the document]
rufinio: Using matcher.reset() is better for performance at the cost of being less readable. In the case of a loop it is one less object to create in each loop iteration, and correspondingly one less object to be garbage collected per loop iteration.
After I moved matcher object in scope of each method I saw in my tests with report_big.txt about 5% decrease in performance. Hope that is compensated by the better readability.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/TestwellCtcTxtParser.java
```diff
@@ -79,113 +71,88 @@ public TestwellCtcTxtParser() {
   @Override
   public void processReport(final SensorContext context, File report, final Map<String, CoverageMeasures> coverageData) {
     LOG.debug("Parsing 'Testwell CTC++' textual format");
-    
+
     try {
       this.scanner = new Scanner(report).useDelimiter(SECTION_SEP);
```
ivangalkin: I missed this during the previous comment: in order to avoid this.scanner.close() calls (e. g. in case of exception), one could use so called "try with resources".
```JAVA
try ( this.scanner = new Scanner(report).useDelimiter(SECTION_SEP) )
{
    ...
}
catch ( ... )
{
}
```

Since Scanner is [AutoCloseable](https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html) it will be automatically closed when control flow lefts the try scope (see [documentation](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html))

cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/TestwellCtcTxtParser.java
```diff
@@ -79,113 +71,88 @@ public TestwellCtcTxtParser() {
...
-    if (matcher.usePattern(FILE_HEADER).find(FROM_START)) {
-      parseFileUnit(coverageData);
+  private boolean parseUnit(final Map<String, CoverageMeasures> coverageData, Matcher headerMatcher) {
+    LOG.debug(headerMatcher.toString());
+
+    if (headerMatcher.find(FROM_START)) {
+      parseFileUnit(coverageData, headerMatcher);
     } else {
-      state = State.END;
       scanner.close();
```
ivangalkin: `scanner.close()` will be not necessary if try-with-resources will be used

cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/TestwellCtcTxtParser.java
```diff
@@ -79,113 +71,88 @@ public TestwellCtcTxtParser() {
...
   }
 
   private void addLines(File file, final Map<String, CoverageMeasures> coverageData) {
     LOG.debug("Parsing function sections...");
 
     CoverageMeasures coverageMeasures = CoverageMeasures.create();
-    while (!matcher.reset(scanner.next()).usePattern(FILE_RESULT).find()) {
-      parseLineSection(coverageMeasures);
+    String nextLine = scanner.next();
+    while (!FILE_RESULT.matcher(nextLine).find()) {
```
ivangalkin: looks like a classic use-case for a `for loop` to me:
```
for (String nextLine = scanner.next(); FILE_RESULT.matcher(nextLine).find(); nextLine = scanner.next())
{
   parseLineSection(coverageMeasures, nextLine);
}
```
... but it's a matter of taste

cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/TestwellCtcTxtParser.java
```diff
@@ -79,113 +71,88 @@ public TestwellCtcTxtParser() {
...
-    
+
     String normalFilename;
-    String filename;
-    File file;
-    
-    filename = matcher.group(1);
-    if((new File(filename)).isAbsolute() == false) {
+    String filename = headerMatcher.group(1);
+    if (!(new File(filename)).isAbsolute()) {
```
ivangalkin: parenthesis are not necessary `if (!new File(filename).isAbsolute()) {`

guwirth:
@rufinio for me the code is ok. If you like to work-in the last points from @ivangalkin tell me.
Otherwise I would merge it now.

rufinio:
Yes, I'd like to apply these last points.

guwirth:
@rufinio thanks a lot for your contribution!

Issue #1494, opened 2018-06-11
US: Add explanation for cognitive/cyclomatic complexity checks


ivangalkin:
Hi All,

while reviewing of the #1398, I've noticed that the complexity checks create an issues without any explanation.
E.g. `ClassComplexityCheck`, `FileComplexityCheck`, `FunctionCognitiveComplexityCheck` or `FunctionComplexityCheck` create an issue on the class/file/function with a message like

```
"The complexity of this function is X which is greater than Xmax authorized."
```

The Java plugin breaks X into several summands, each of them is a secondary location, pointing to the complexity source (nesting, switch/case, loops, try/catch, inner declarations of classes or lambdas etc).

The resulting message looks like:

```
"The complexity of this function is X which is greater than Xmax authorized."
Line L0: nesting +1
Line L1: try statement +2
Line L2: continue statement +1 etc
```

Implementation in Java plugin:

https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/main/java/org/sonar/java/checks/CognitiveComplexityMethodCheck.java
https://github.com/SonarSource/sonar-java/blob/master/java-frontend/src/main/java/org/sonar/java/ast/visitors/CognitiveComplexityVisitor.java

Example of working with multiple/secondary issue locations:  #1447

guwirth:
@ivangalkin maybe interesting to align it with Java, but I don't need it.

ivangalkin:
Hi @guwirth,

I used Java as example, but I believe that other plugins (incl. `SonarCFamily`) have this feature, too. Explanation about the source of [cognitive] complexity is optional, of course. However, if such additional information exists, it improves the sensitivity: now developer know why there is a code smell and which particular parts cause the "brain overload".

By the way, while writing this comment I found the duplicate US #1297. It contains a good illustration, how such issue should look like. It looks, like I am not the only person, who misses this feature.

PR #1499, opened 2018-06-19
CxxCognitiveComplexityVisitor: clean-up

ivangalkin:
minor performance findings while looking at #1494

1. utilize return value of `Set<T>::add(T)`
2. minimize number of `Arrays.asList()` calls
3. remove `checkedNodes.addAll(watchedDescendants);` since all watched descendants are added while `visitChildren()` call

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1499)


guwirth:
unit tests are unchanged and still green

Issue #1509, opened 2018-06-28
sonar-cxx custom metrics are not dispayed/saved


ivangalkin:
1. go to `https://<server>/project/activity?id=<project_id>`
2. see metrics diagram -> choose `Custom` in the drop-down menu
3. click `Add metric` button
4. select one of the custom sonar-cxx metrics (see [CxxMetrics.java](https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxMetrics.java)) e.g. `CppCheck issues`
5. the diagram will show nothing

Plugin version: HEAD
SonarQube version: 7.1.0.11001

I guess, that the `CxxLanguage::MetricsCache` is not functioning properly. Is it really filled with data? Is it really persisted at the end?

ivangalkin:
There were multiple causes for the wrong behavior:

1. the CxxLanguage::MetricsCache didn't function at all: basically it was never filled, the stores and lookups were made with inconsistent keys and results were ignored without any warnings
2. the *PUBLIC*API* metrics were not auto-aggregated, which means that there have to be an additional step in order to accumulate per-file measurement into the per-module metric

The change https://github.com/ivangalkin/sonar-cxx/commit/49fc87623b2a6dce1dc46b7db6311479fa955dcd is pretty large. I will refactor and submit it asap.

Issue #1062, opened 2017-02-21
Doxygen - log file 


tomikais:
Hello,

this is a suggestion for a new function. It would be helpful to have the Doxygen warnings and issues in SonarQube. 
This could be implemented similar to the Visual Studio log file parser.

Regards, Thomas

JStb:
I have a an incomplete set of sonar.cxx.other.rules for doxygen logfiles plus a python converter from doxygen.log to a cxx-reports.xml file. 
Unfortunately I was not able to find a complete set of warnings in the doxygen src.

jmecosta:
@JStb you are more than welcome to add your definitions (even if incomplete) to https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Extending-the-code-analysis#resources  and make the convertes availalbe here: https://github.com/SonarOpenCommunity/sonar-cxx/tree/master/sonar-cxx-plugin/src/tools 

the idea was that anyone can add more tools there

guwirth:
@JStb agree with @jmecosta you are welcome to add your scripts.

On the long run it would be better to write an own sensor for this. Agree that it is more or less a copy of the compiler sensor (https://github.com/SonarOpenCommunity/sonar-cxx/tree/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/compiler)
```
sonar.cxx.doxygen.rules=...
sonar.cxx.doxygen.reportPath=...
```

JStb:
I just got the go to contribute from my employer. I'll refactor the converters a bit and them put them online.

an own sensor would be nice, but as doxygen is not bound to any language... does it make sense to put it in sonar.cxx.doxygen?

jmecosta:
@JStb what about creating the sensor in the doxygen plugin? https://github.com/SonarQubeCommunity/sonar-doxygen that way it will not bee bound too C++ only

guwirth:
@JStb we have a lot of sensor in this plugin couls also be isolated in an extra plugin. Question is always how much effort it is to maintain an own plugin (and who is willing to do it). Adding it here is one XML file describing the rules and two small Java files.

jmecosta:
@guwirth you are forgetting the support for troubles that might or not show up in future. Once this grows to have n+1 sensores becomes difficult to provide support.

[...]

PR #1510, opened 2018-06-28
deprecate *PUBLIC*API* metrics

ivangalkin:
* `PUBLIC_API`, `PUBLIC_UNDOCUMENTED_API` and `PUBLIC_DOCUMENTED_API_DENSITY`
  were depricated since SQ 6.2 (https://jira.sonarsource.com/browse/SONAR-8328)
  but existed as custom sonar-cxx metrics

* the minimal supported SQ version now is 6.7

* moreover, it looks like there is a general problem in displaying
  custom metrics (see #1509)

* that means that nobody a) expects that the SQ plugin implements the
  deprectad metrics and b) nobody misses them (because they are just not
  visualized)

* BTW public API measurements belong to the obligatory AST Visitors
  and introduce the time overhead of ~6% (measured with #1507,
  6% means, that if there is no sensors activated at all
  the importing of C++ project will still cause a calculation
  of a bunch of metrics; summary overhead of this calculation
  considered as 100%)

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1510)


ivangalkin:
`AbstractCxxPublicApiVisitor` and the corresponding tests `CxxPublicApiVisitorTest` (must be renamed) can be moved from `cxx-squid` to `cxx-checks` too. If there will be general agreement w.r.t. this PR, I will update the change. Please review.

guwirth:
Hi @ivangalkin, there was by intention a custom metric introduced. Think there is much value to detect undocumented public API. This is for sure technical debt. Never understood why SonarSource removed it? What’s correct: if someone don’t use it he should not pay for it. Maybe we can check if rule i s active. Regards,

guwirth:
Alternativ would be #1062

ivangalkin:
will be abandoned as soon as #1509 is finished

ivangalkin:
abandon in favor of #1514

PR #1522, opened 2018-07-24
Fix #1521: rules repository ignored the language specific repository suffix

### LINKED ISSUE #1521: C-Plugin: sensors use wrong rules repository keys
### Description

For a C test project I used the python version of scan-build command to generate clang analyzer reports with the command as described in the Wiki:

scan-build.py -plist --intercept-first --analyze-headers -o source-check/analyzer_reports make

The reports are generated as expected.

Using sonar-scanner with the property sonar.c.clangsa.reportPath=source-check/analyzer_reports/*/*.plist set in sonar-project.properties the reports are completely ignored, 
even with -X no clue is shown.

I can provide the report files if needed.

### Related information

* cxx plugin version 1.1.0
* SonarQube version 7.2.1

ivangalkin:
close #1521

BUGFIX

* in order to allow the co-existance of C and C++ plugins, rules
  repositories were doubled

* the effective repository key was built as
  `SENSOR.KEY + LANGUAGE.REPOSITORY_SUFFIX`

* the usage of this format was inconsistent; in the majority of
  cases just `SENSOR.KEY` was used

* The problem affected only subclasses of `CxxReportSensor`. The class `CxxSquidSensor` solved the key ambiguity for the rules repository by means of `CxxLanguage.getRepositoryKey()`. This is not consistent IMHO.

REFACTORING

* simplify CxxReportSensor: use only one rules repository per sensor
    * split CxxReportSensor into CxxReportSensor and CxxIssuesReportSensor
    * split CxxCompilerSensor into CxxCompilerGccSensor and CxxCompilerVcSensor
    * remove repoKey from CxxReportIssue and move it to CxxIssuesReportSensor
    * misc refactoring and simplification

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1522)

### DIFF
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSARuleRepository.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSARuleRepository.java
@@ -29,7 +29,7 @@ import org.sonar.cxx.sensors.utils.CxxAbstractRuleRepository;
  */
 public class CxxClangSARuleRepository extends CxxAbstractRuleRepository {
 
-  public static final String KEY = "ClangSA";
+  private static final String KEY = "ClangSA";
   public static final String CUSTOM_RULES_KEY = "clangsa.customRules";
   private static final String NAME = "Clang-SA";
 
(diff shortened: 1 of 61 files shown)


guwirth:
@ivangalkin most interesting question in such a case is always:
* is this change backward compatible?
* or do we destroy the issue history?

ivangalkin:
Hello @guwirth 

The error was not that dramatic: the warnings pointed to the correct rules repository, BUT the initial sensor activation checks "if corresponding rules are activated" were wrong. The repository suffix was ignored by the check. For C++ plugin the suffix is empty, so the bug was not visible. The C sensors however checked de-facto the activation of C++ rules (because the suffix "-c" was missing).

Additionally I unified and simplified the usage/generation of the rules repository key. Issue history is not destroyed. The change is backwards compatible.

Regards

guwirth:
@ivangalkin in case of a bugfix please add always `close #xxxx` to the description of a PR. This will automatically link and close it in case of merging a pull request.

ivangalkin:
@guwirth thank you for the hint!

Issue #1400, opened 2018-01-21
side by side usage of C++ commercial and community plugin


guwirth:
As an user I like to use the C++ commercial and community plugin in parallel on the same server.

* should be possible to install both plugins on the same server
* settings should not conflict in configuration file
* should be possible to mix sensors of both plugins

guwirth:
Commercial plugin settings to keep in mind making the plugins working side-by-side.

# C / C++
## C file suffixes
Comma-separated list of suffixes of C files to analyze.
Key: `sonar.c.file.suffixes`
## C++ file suffixes
Comma-separated list of suffixes of C++ files to analyze.
Key: `sonar.cpp.file.suffixes`
## Objective-C file suffixes
Comma-separated list of suffixes of Objective-C files to analyze.
Key: `sonar.objc.file.suffixes`
# Coverage
## Bullseye XML report
Path to the Bullseye XML Coverage Report. The path may be either absolute or relative to the project base directory.
Key: `sonar.cfamily.bullseye.reportPath`
## Gcov reports
Path to the directory containing the *.gcov Gcov report files. The path may be either absolute or relative to the project base directory.
Key: `sonar.cfamily.gcov.reportsPath`
## llvm-cov report
Path to the Coverage Report generated by "llvm-cov show". The path may be either absolute or relative to the project base directory.
Key: `sonar.cfamily.llvm-cov.reportPath`
## Visual Studio XML reports
Pattern for search for Visual Studio Coverage XML reports. The pattern may be either absolute or relative to the project base directory. For example: "**/*.coveragexml" will find all "*.coveragexml" files in all sub-directories of current project.
Key: `sonar.cfamily.vscoveragexml.reportsPath`
# Tests
## CppUnit reports
Path to the directory containing the *.xml CppUnit report files. The path may be either absolute or relative to the project base directory.
Key: `sonar.cfamily.cppunit.reportsPath`
# Miscellaneous
## Ignore header comments
If set to "true", the file headers (that are usually the same on each file: licensing information for example) are not considered as comments. Thus metrics such as "Comment lines" do not get incremented. If set to "false", those file headers are considered as comments and metrics such as "Comment lines" get incremented.
Key: `sonar.cfamily.ignoreHeaderComments`
# Deprecated
## C language standard
This property is deprecated, because the build-wrapper gathers this information automatically.
Key: `sonar.c.std`
## C++ language standard
This property is deprecated, because the build-wrapper gathers this information automatically.
[...]

Issue #1345, opened 2017-11-26
Can't use commercial and community C plug-in in parallel on one server


guwirth:
Not possible to use commercial and community C plug-in in parallel, they are using the same plugon key.

**C** property | community | commercial 
 --- | --- | ---
plugin key | c | c
plugin name | c | C
config property key | sonar.c... | ???
repository key | c | c

* First of all I think **plugin key** must be unique. Think here we have already a problem with the C plugin? To try that out you can try to install C pugin together with commercial one on one server. Fear you get an error during server start.
* Second the **property name** is just a label.
* Think makes sense that **config properties in sonar-project.properties** have a different key. To have both in parallel in one file?
* **repository key** is the label you see in the upper right corner of a rule

guwirth:
Duplicate, see #1400

Issue #1328, opened 2017-11-20
SonarSource Plugin and Sonar-CXX Plugin in parallel


niecore:
Hello dear Community,

Is there a official guideline on how to use the SonarSource Plugin and the Sonar-CXX Plugin in parallel?

Currently we archive the result by removing all file extensions from the Sonar-CXX Plugin responsibility. When a user wants to use the  Sonar-CXX Plugin he has to force the language to `c++` and set his file extensions via the option `sonar.lang.patterns.c++`.

This works fine so far. Unfortunately it seems a little bit confusing to have now Rules for `c++` and `C++`. 

 Is there an option to change the language key of the  Sonar-CXX Plugin from 'c++' to i.e. 'C++ (Community Plugin)'?

guwirth:
> Is there a official guideline on how to use the SonarSource Plugin and the Sonar-CXX Plugin in parallel?

Hi @niecore,

As far as I know there is no way to use them in parallel. This is because file extensions (e.g. .cpp) has to be unique assigned to **one** plugin.

> When a user wants to use the Sonar-CXX Plugin he has to force the language to c++ and set his file extensions via the option sonar.lang.patterns.c++.

Don't understand this. For my understanding you have to set properties below.  How exactly you are doing this?

```INI
sonar.language=c++
sonar.cxx.suffixes.sources=...
```

> Unfortunately it seems a little bit confusing to have now Rules for c++ and C++.

Each plugin has a unique name and key. We are using `c++` for name and key and `cxx` as property key.
What is the commercial plugin using?

> Is there an option to change the language key of the Sonar-CXX Plugin from 'c++' to i.e. 'C++ (Community Plugin)'?

Changing the language key would mean that all current users are loosing there history. So has to be discussed...

A general question: Why do you need both plugins in parallel. What's missing in the one or in the other?

Regards,

niecore:
Thanks for you answer @guwirth .

First of all I must state, that a user has to decide which of the two plugins he wants to use. By using them in parallel I mean that both plugins are available on the SonarQube instance and the user can somehow determine which plugin he wants to run.

> As far as I know there is no way to use them in parallel. This is because file extensions (e.g. .cpp) has to be unique assigned to one plugin.

To bypass that restriction I remove all file extensions from the Sonar-CXX Plugin responsibility in the global configuration. 

When a user wants to use the Sonar-CXX Plugin he has to do two things:

1. Force the language to c++
`sonar.language=c++`
[...]

Issue #1251, opened 2017-10-14
C plugin usage


guwirth:
Hi, 

There is a new **C plugin** as part of our repository available for some time.
* Who has already some experience with it? Can you share you experience with us please?
* Who has already both cxx and c plugin in parallel in use? Is this working as expected?

Latest snapshot versions are available here:
https://ci.appveyor.com/project/SonarOpenCommunity/sonar-cxx/branch/master/artifacts

Regards,

guwirth:
@jmecosta is the plugin on your side in productive use?

Wondering how this is working in
* mixed projects with c / cpp files?
* same header files in use by c and cpp file?

jmecosta:
@guwirth when implementing the feature i started using it but after upgrading to last version of SQ i had to disable it since the server was not starting. So i will need to look at this sometime soon. 

the idea is that you need to adjust the extensions in a way they dont clash. 

i will try to get this working before the next release

guwirth:
@jmecosta thanks for your answer. I'm wondering how header files (.h, .hpp) are handled if they are assigned to c and cxx plugin. because they are assigned to both plugins then.

* Which rules are used?
* Are issues assigned twice?

Any experience?

SonarOpenCommunityAdmin:
basically you will have 2 different profiles that you need to maintain. the key here is to decide what is c and what is c++ since they cannot be defined in both plugins. 

i was doing 

c -> .c, .h
c++ -> .cpp .hpp

after that it should work as normally with no duplicated issues i think. the rules we just copied the profiled from c++ but thats is really not good and needs to be worked on

jmecosta:
basically you will have 2 different profiles that you need to maintain. the key here is to decide what is c and what is c++ since they cannot be defined in both plugins.

i was doing

c -> .c, .h
c++ -> .cpp .hpp

after that it should work as normally with no duplicated issues i think. the rules we just copied the profiled from c++ but thats is really not good and needs to be worked on

jmecosta:
looks good in a clean installation

![image](https://user-images.githubusercontent.com/1041352/31779538-04ba0628-b4fc-11e7-9ea2-1f7f832a97fd.png)

the only thing i need to do is to put in sonar properties
sonar.cxx.suffixes.headers=.hxx,.hpp,.hh
sonar.cxx.suffixes.sources=.cxx,.cpp,.cc

guwirth:
[...]

PR #1525, opened 2018-07-25
start implementation

jmecosta:
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1525)


jmecosta:
@guwirth before proceeding,  a couple of questions.

1. because there is no way of not installing the c plugin, the suffix should now be set in a way that both c and c++ are supported out of the box. this means also that the C_FILES_PATTERNS_KEY should be deprecated? 

2. its possible to define rule profiles for both languages, however there will be hardcoded language keys in the sensors module because we can anymore pass those independently from the plugin

3. notice that the rules registering was still trying to load rules from server side, i think this is not possible anymore and not documented, if remove it then we dont need to pass the language to those at all

4. imo we dont need different sensors for each language, supose you agree with this?

ivangalkin:
@jmecosta I would suggest you to start you work on top of the (already merged) #1522. I've done some refactoring there, which, I believe, will simplify/isolate the usage of rules repositories + more sensor refactoring.

guwirth:
> One last thing from me. What is exactly the problem we are solving by merging them? Im a bit confused since there are quite a few issues with c, c++, paid c++, why both c and cxx, etc and dont see how the merge would fix them

@jmecosta this are indeed different problems

1. side by side of C++ plugins
2. side by side of C plugins
3. mixed usage of community and commercial plugin

The problems I see with the current c plugin are:
* People are thinking we have a c plugin, but it isn't really one. It maps only the .c file extension to other repositories.
* c plugin is not working side by side with comercial c plugin because of overlapping language key
* a lot of issues with maintenance effort

jmecosta:
On Fri, 27 Jul 2018 at 09:25 Günter Wirth <notifications@github.com> wrote:

> @jmecosta <https://github.com/jmecosta>
>
> because there is no way of not installing the c plugin, the suffix should
> now be set in a way that both c and c++ are supported out of the box. this
> means also that the C_FILES_PATTERNS_KEY should be deprecated?
>
> Think it helps to read #1400
> <https://github.com/SonarOpenCommunity/sonar-cxx/issues/1400>, #1345
> <https://github.com/SonarOpenCommunity/sonar-cxx/issues/1345>, #1328
> <https://github.com/SonarOpenCommunity/sonar-cxx/issues/1328>, #1251
> <https://github.com/SonarOpenCommunity/sonar-cxx/issues/1251>, #1105
> <https://github.com/SonarOpenCommunity/sonar-cxx/issues/1105> first.
>
from the reading, and from past discussions we have 3 main issues.
1. coexistence of community and commercial plugins
2. coexistence of c and cxx
3. why we need a c plugin, if cxx handles everything.

First what is the current status, after @ivangalkin merged #1522 (that
should handle the issues not reported because of incorrect repository keys)
is that we should have a fully functioning side by side c and c++ plugin.

Regarding 3, imo be able to have dedicated rule sets for the two languages
(as i mentioned earlier) is the main advantage of having the 2 plugins and
from the discussions this has not been stressed enough. I would like to
hear your opinion about this @ivangalkin, @Bertk, @guwirth. But once you
start using these plugins in legacy codebase you will see that be able to
disable many of the rules for c++ brings a huge advantage in terms of
maintenance.

Regarding 2, and since we already discussed there is no technical issue
regarding the merge, the benefit of merging those 2 is that in practice a
end user will need to handle the suffixes immediately after installation
unless we define a default behaviour that will work. however .h files might
and will likely conflict regarding what is the style that should be used
and users will need to configure their best way of handling. As advantages,
is that efficiently its less confusing to have both plugins. Also from past
discussions, its the way, the commercial plugins handles those so it would
be good to follow the same trend.

Regarding 1, from the discussions, most users are most interested to extend
what they get with the commercial with the external rules coming from
cppcheck, etc. First SQ 7.2 provides already a way to import external
issues from external tools without a need for a dedicated plugin, however
this suffers from the fact that you cant define a ruleset for those
external issues (im not sure how that will develop in feature, but if you
can disable those from profile imo that feature does not look very
promising to me).
Regarding 1 also, from discussions, having all plugins together is a bunch
of work to make sure all the suffixes are properly setup. And this will not
be solved by merging the c++ and c plugin.

Regarding 1 again, one possibility to solve all this, would be to create a
third plugin that effectively does not create a new language, instead only
contributes to existing language and creates new rule repositories. You can
find examples of this setup in FindBugs plugin that just contributes to the
main java plugin. Creating such plugin should be straightforward since the
all modules are created to be reused in such away.

So, basically based on all this, im still not so convinced merging is the
way to go.

its possible to define rule profiles for both languages, however there will
> be hardcoded language keys in the sensors module because we can anymore
> pass those independently from the plugin
>
> In the past I saw a possibility to create a list of Languages instead of
> Language. So it should be possible to assign Languages to sensors.
>
> https://github.com/SonarSource/sonarqube/blob/master/sonar-plugin-api/src/main/java/org/sonar/api/resources/Languages.java
>
Suppose it will then be hardcoded languages in the sensors, rather then
specialisation in plugins as it is now. But im not against it. One thing
tough, a issue can spawn over multiple locations so effectively can cross
multiple languages and from what i see we cant create a issue over multiple
languages, but this seems a api issue imo, however will need be handled
somehow in the sensor.

jmecosta:
On Fri, 27 Jul 2018 at 08:44 Günter Wirth <notifications@github.com> wrote:

> One last thing from me. What is exactly the problem we are solving by
> merging them? Im a bit confused since there are quite a few issues with c,
> c++, paid c++, why both c and cxx, etc and dont see how the merge would fix
> them
>
> @jmecosta <https://github.com/jmecosta> this are indeed different problems
>
>    1. side by side of C++ plugins
>    2. side by side of C plugins
>    3. mixed usage of community and commercial plugin
>
> The problems I see with the current c plugin are:
>
>    - People are thinking we have a c plugin, but it isn't really one. It
>    maps only the .c file extension to other repositories.
>
> after  @ivangalkin merged #1522 this should be working now. it gives
metrics you can run sensors. Is there something else to be considered a
language plugin?

>
>    - c plugin is not working side by side with comercial c plugin because
>    of overlapping language key
>
> suppose we should be able to fix this in same way we handle c++ plugin?
suppose instead of calling c, we  can call c(something). although i prefer
not to mix paid and commercial plugins in same instance and produce a third
plugin that handled only the import of external issues.

>
>    - a lot of issues with maintenance effort
>
> partly my fault :) and because we dont have yet a clear direction on how
to handle this. hopefully we can solve this either by eradicating it or by
creating a new plugin for usage with commercial plugin

jmecosta:
On Sun, 29 Jul 2018 at 19:26 Günter Wirth <notifications@github.com> wrote:

> @jmecosta <https://github.com/jmecosta>
>
>
>    1. coexistence of community and commercial plugins
>
> This is very useful if you have one central server and the teams should be
> able to decide if they like to use the community or commercial plugin.
>
but why even pick, whats the benefit of choosing the community to create
metrics when the commercial is far better?

>
>    1. coexistence of community c and cxx
>
> Is the minimum and the idea of multi-language support
>
>
>    1. why we need a c plugin, if cxx handles everything.
>
> It's more like: if c plugin it must be a complete c plugin.
>
lets then make it complete, im pretty sure its not very far. probably some
integration tests are needed to do this i can try to add them also

> ... is that we should have a fully functioning side by side c and c++
> plugin.
>
> Maybe? Hearing less or nothing means everything ok or no one is using it.
>
i am using at least at work. indeed no issues are still reported in c files
but that should be fixed already in latest master i will update and confirm

> But once you start using these plugins in legacy codebase you will see
> that be able to disable many of the rules for c++ brings a huge advantage
> in terms of maintenance.
>
> Currently we don't need it for two reasons:
>
>    - We use C++ compilers also to compile C code.
>
> yes but this is totally different than static code analysis issues. We
also compile c with c++ but they dont care about those. for example we have
plenty of c style casts that we dont allow in modern c++ and this rule
needs to be disable in the c quality profile.

>
>    - We have different quality profiles for C and C++ and assign it on
>    project level (not on extension level)
>
> Yes but this means simply you cant mix both styles in one project, which
is what we need since most of our projects contain both languages and with
your approach you either pick one or the other. Either way loads of false
positives to surpress

>
>
> But I understand the need to map extension (language) to repository for C.
> Only question is if we should do this as part of the cxx plugin (and how
> much maintainer we will have).
>
supose once it works properly, because we are reusing the all cxx checks
lib we it should be sharing the same issues as the c++.

> however .h files might and will likely conflict regarding what is the
> style that should be used and users will need to configure their best way
> of handling.
>
> That's one of the reasons we do it per project.
>
in principle there should not be much more stuff then function/macros
definitions in the .h. So for us, setting only .h to might be giving only a
few false positives to handle. But this would need to see in real life, and
i dont have substantial data to be sure about it. Anyway, per project
approach would not work for us.

> Also from past discussions, its the way, the commercial plugins handles
> those so it would be good to follow the same trend.
>
> Yes, that's a strong argument.
>
> Regarding 1, from the discussions, most users are most interested to
> extend what they get with the commercial with the external rules coming
> from cppcheck, etc.
>
> In our case new users which has the choice start mostly with the
> commercial plugin. This provides everything out of the box. Afterwards they
> like to use the additional sensors of the cxx plugin.
>
one good reason to create a new plugin that contains no language defintions

> Regarding 1 also, from discussions, having all plugins together is a bunch
> of work to make sure all the suffixes are properly setup. And this will not
> be solved by merging the c++ and c plugin.
>
> It's not really complicated. Only thing what you have to do is to assign
> file extensions to plugins (languages/sensors).
>
each project has to take care of the extensions, so its a substencial work
to handle imo

> Regarding 1 again, one possibility to solve all this, would be to create a
> third plugin
>
> Not really a fan of this. Than we have to handle additional compatibility
> issues.
>
what kind of compatibiltiy issues are we talking about?

> You can find examples of this setup in FindBugs plugin that just
> contributes to the main java plugin.
>
> Not sure if this solves the problem. This provides a sensor for one fixed
> language (Java) only.
>
it would be one extension per language i think, but this is a good point.
would probably mean 2 other plugins then if following the same approach.
Would make sense if we have sepearte plugins also per language.

> metrics you can run sensors. Is there something else to be considered a
> language plugin?
>
> If you wanna have a complete plugin you should support:
>
>    - parsing of the language, create AST
>    - syntax highlighting
>    - metrics (need parsing)
>    - rules / rule repositories (external)
>    - checks (base partly on AST)
>
> although i prefer not to mix paid and commercial plugins in same instance
> and produce a third plugin that handled only the import of external issues.
>
> it would be and partial plugin, not a language plugin in this case. So
only rules and checks would be needed i would say

> Use case is: you have one central company wide maintained SQ server.
> You want that users can decide which plugin they like to use for C and C++
> in there local configuration file.
>
in our case we are more interested in importing all available files and let
project admins configure their quality profiles. So we work more on that
basis in our company since its much easier to onboard people into sonar
since we can just run the analysis and show them the results. Going to them
and explain that they needed to do this and that takes much longer time and
its not as effective in our case.

guwirth:
@jmecosta 

> but why even pick, whats the benefit of choosing the community to create metrics when the commercial is far better?

Mostly they use the commercial to read the source files (indexing) and do a basic static code analysis. Afterwards they are missing some sensors we have and the community doesn't (e.g. read warnings fro compilers, read unit test results, read coverage, ...). 

> also compile c with c++ but they dont care about those. for example we have plenty of c style casts that we dont allow in modern c++ and this rule needs to be disable in the c quality profile.

Typically you are using C in C++ code and have to include the C headers. In this case you get a lot of false positives anyway.

> supose once it works properly, because we are reusing the all cxx checks lib we it should be sharing the same issues as the c++.

My point is still that normally the C and the C++ plugin should have two different grammars. Or other possibility: we remove AST based checks and allow only external tools.

> one good reason to create a new plugin that contains no language defintions
> what kind of compatibiltiy issues are we talking about?

I still fear a maintenance hell. You have to ask people which SQ version, which c plugin is installed, which c++ plugin is installed, which sensor plugin is installed, ... There are thousand of combinations possible. What do we guarantee?

> mean 2 other plugins 

c (community) plugin
cxx (community) plugin
c sensor (community) plugin
cxx sensor (community) plugin

> only rules and checks would be needed i would say

Checks are on top of the AST, (grammar). Means you have to support a C grammar.

> you have one central company wide maintained SQ server.

Point here is: if you are using a commercial edition you are paying per instance. So there is a strong interest of being able to install all plugins on one instance.

guwirth:
@ivangalkin, @Bertk , @all any other ideas / comments to the idea of @jmecosta to split the plugin into
* indexing, parsing, AST, checks, basic metrics
* and additional sensors

See also discussion above.

PR #1528, opened 2018-07-27
simplify preprocessor if squid checks are disabled

ivangalkin:
* the role of preprocessor (CxxPreprocessor) is negligible
* the evaluation of preprocessor directives plays some
  minimal role if AST is used for internal checks
  (SquidSensor)

* however if sonar-cxx imports external reports only, the AST
  is used just for highlighting and metrics calcualation
* whether preprocessor directives are evaluated or not
  has no effect on highlighting and almost no effect
  on the metrics

* sonar-cxx is positioned as a plugin for integration
  of external analyzers, yet users are facing
  many proprocessor warnings about missing includes

SUGGESTION:

If no internal checks are activated, let proprocessor run
in simplified mode. No preprocessor directives will be evalated

PROS:

* speed up
* no misleading warnings

CONS:

* metrics like "number of functions", "number of classes" etc
  might change if C/C++ code extensively uses preprocessor magic.
  E.g. if macros are used in order to generate functions or classes.

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1528)

### DIFF
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/squid/CxxSquidSensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/squid/CxxSquidSensor.java
@@ -30,6 +30,7 @@ import javax.annotation.Nullable;
 
 import org.sonar.api.batch.fs.FileSystem;
 import org.sonar.api.batch.fs.InputFile;
+import org.sonar.api.batch.rule.ActiveRules;
 import org.sonar.api.batch.rule.CheckFactory;
 import org.sonar.api.batch.sensor.Sensor;
 import org.sonar.api.batch.sensor.SensorContext;
@@ -117,6 +118,10 @@ public class CxxSquidSensor implements Sensor {
       .onlyOnFileType(InputFile.Type.MAIN);
   }
 
+  private boolean areSquidChecksEnabled(ActiveRules activeRules) {
+    return !activeRules.findByRepository(language.getRepositoryKey()).isEmpty();
+  }
+
   /**
    * {@inheritDoc}
    */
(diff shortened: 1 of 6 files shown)


ivangalkin:
@guwirth thank you for your comments
meanwhile I believe, that it's not enough to categorize the metrics in

* requires preprocessor
* doesn't require preprocessor

Even more: we should run some visitors on original AST (before preprocessor). That means, we have to categorize visitors between

* runs on original code
* runs on preprocessed code

Highlighting visitor is a nice example for the visitor, which is spoiled by preprocessor. Also the visitor for complexity metric and the corresponding squid check are totally falsified by such thing as macro expansion. E.g. I'm working on #1494 and that's how a single `TEST_F` with 3 `EXPECT_EQ`s look like:

complexity

The complexity check recognizes the following AST nodes:
* 2 function definitions
* 3 if statements
* 6 switch labels

This leads to the total complexity of 11, whereby the correct cyclomatic complexity is just 1 (function definition).

I interpret your comment...

> Agree except the metrics.

... as the sign, that collection of metrics is very important functionality for sonar-cxx. So in my opinion our complexity metrics are already broken because of preprocessor.

guwirth:
@ivangalkin 

> Even more: we should run some visitors on original AST (before preprocessor). That means, we have to categorize visitors between
    - runs on original code
    - runs on preprocessed code

Not sure if this is possible to create two AST? Think current implementation is a queue with input => preprocessor => output1 => parser => output2 => AST?

> ... as the sign, that collection of metrics is very important functionality for sonar-cxx. 

Yes I think so. Issues / Technical Debt is one thing. The other is to identify hotspots in the code. That's what metrics are good for. In the past it was possible to create heatmaps with SQ for every metric. It's a shame that this is no more possible.

guwirth:
@ivangalkin  Maybe we should it make more concrete and list all checks with their requirements here in this thread to discuss it?

ivangalkin:
I believe #1552 and #1567 eliminated my concerns about the preprocessor. I'll abandon this PR.

PR #1529, opened 2018-07-29
fix quality flaws

ivangalkin:
* (minor) findings from LGTM
* https://lgtm.com/projects/g/SonarOpenCommunity/sonar-cxx/alerts/?mode=list

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1529)


guwirth:
@ivangalkin 

SQ Quality Gate failing with: https://sonarcloud.io/project/issues?id=org.sonarsource.sonarqube-plugins.cxx%3Acxx&resolved=false&sinceLeakPeriod=true&types=BUG

Maybe you can fix this in another PR.

ivangalkin:
@guwirth thank you for the hint

I'll rewrite the code, although the both findings are false-positives:

```JAVA
if (context.getMeasure(densityKey) != null) {
  LOG.error("{}", context.getMeasure(densityKey).getDoubleValue();
}
```

The warning *"A "NullPointerException" could be thrown; "getMeasure()" can return null"* is not relevant: The fact, that `context.getMeasure(densityKey)` is not null is checked explicitly in the if-condition. SonarJava seems not to be able to build a proper control flow graph or it assumes, that the function is not deterministic.

Issue #1527, opened 2018-07-27
rules repositories: solve inconsistent rules activation


ivangalkin:
make all rules disabled by default
see #1521

guwirth:
@ivangalkin agree: make all rules disabled by default

guwirth:
The default rules are defined in the default-profile.xml files:
* https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-c-plugin/src/main/resources/default-profile.xml
* https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/resources/default-profile.xml

They are activated in:
* https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxDefaultProfile.java#L54
* https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/sonar-c-plugin/src/main/java/org/sonar/plugins/c/CDefaultProfile.java#L54

I will keep the files (mechanism) but remove all items. Default is then that all rules from external checkers are off per default.

Issue #1365, opened 2017-12-27
Add Compilation Database to handle different configuration sources in an unique way


guwirth:
At the moment we have different solution to handle macros and include directories from different sources:
* config file
* Visual Studio LOG file
* JsonCompilationDatabase

Would be the best to have only one internal handling for this:
* read the source only once, store result in CxxConfiguration
* Visual Studio LOG file: reading the configuration should be separated from other functionality. At the moment reading issues and settings is mixed behind one config value.

**Proposal:**
* create a database (map) in CxxConfiguration
* should be possible to define global settings (valid for all files)
* should be possible to define source file specific macros and includes
   * key is the source file name
   * values are:
      * file specific include paths
      * file specific macros
* different config sources should use this one DB as backend
* config file should be read only once

see also #1325, #1326,

Bertk:
Hi @guwirth,

will this proposal also support C++ Modules? 

Visual Studio supports definitions for VS C++ projects and for C++ implementation files. As far as I know does the C++ Community plug-in not supports all of this levels. Will this proposal consider the definitions on file level?

Best regards
Bert

guwirth:
Hi @Bertk,

in C++ typically only the C or C++ files are compiled (and the header files only indirect). So it should be sufficient to create a database with c/cpp files and the related macros and includes per file?

Regards,
Günter

PR #1533, opened 2018-08-05
Split compiler sensor into GCC and MSVC compiler to use it at the same time

### LINKED ISSUE #1180: Refactoring: Split compiler sensor into GCC and MSVC compiler to use it at the same time
**sonar.cxx.compiler.parser** option allows either GCC or Visual C++, not both.

guwirth:
Users like to use the VC and GCC compiler results in parallel.
* refactoring of compiler sensors
* close #1180

**Attention:**
* This is a breaking change - `sonar.cxx.compiler` settings are no more supported.
* you have to define the named groups `<file>, <line>, <id>, <message>` in your regex now: <br>`(.*>)?(?<file>.*)\((?<line>\d+)\)\x20:\x20warning\x20(?<id>C\d+):(?<message>.*)`
* Reading messages line by line:  `^...$` no more needed in the regex
* tbd: update wiki after merge

Remove:
```
sonar.cxx.compiler.parser
sonar.cxx.compiler.reportPath
sonar.cxx.compiler.regex
sonar.cxx.compiler.charset
```
Add:
```
sonar.cxx.vc.reportPath
sonar.cxx.vc.regex
sonar.cxx.vc.charset

sonar.cxx.gcc.reportPath
sonar.cxx.gcc.regex
sonar.cxx.gcc.charset
```
To read includes and defines from MSBuild log use:
```
sonar.cxx.msbuild.reportPath
sonar.cxx.msbuild.charset
```

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1533)


cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/gcc/CxxCompilerGccParser.java
```diff
@@ -0,0 +1,70 @@
...
+
+    LOG.info("Parsing 'GCC' format ({})", charset);
+
+    try (Scanner scanner = new Scanner(report, charset)) {
+      Pattern p = Pattern.compile(reportRegEx, Pattern.MULTILINE);
+      LOG.debug("Using pattern : '{}'", p);
+      MatchResult matchres;
+      while (scanner.findWithinHorizon(p, 0) != null) {
+        matchres = scanner.match();
+        String filename = matchres.group(1).trim();
```
guwirth: In case group not available could be null

cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/gcc/CxxCompilerGccSensor.java
```diff
@@ -0,0 +1,75 @@
...
+import org.sonar.cxx.CxxMetricsFactory;
+import org.sonar.cxx.sensors.compiler.CxxCompilerSensor;
+
+public class CxxCompilerGccSensor extends CxxCompilerSensor {
+
+  public static final String KEY = "GCC";
+  public static final String REPORT_PATH_KEY = "gcc.reportPath";
+  public static final String REPORT_REGEX_DEF = "gcc.regex";
+  public static final String REPORT_CHARSET_DEF = "gcc.charset";
+  public static final String DEFAULT_CHARSET_DEF = "UTF-8";
```
guwirth: What is default charset for GCC?

cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/gcc/CxxCompilerGccSensor.java
```diff
@@ -0,0 +1,75 @@
...
+  // search for single line with compiler warning message - order for groups: 1=file, 2=line, 3=message, 4=id
+  public static final String DEFAULT_REGEX_DEF = "^(.*):([0-9]+):[0-9]+:\\x20warning:\\x20(.*)\\x20\\[(.*)\\]$";
```
guwirth: Is this latest regex for GCC?

cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/vc/CxxCompilerVcParser.java
```diff
@@ -96,7 +67,7 @@ public void processReport(final SensorContext context, File report, String chars
 
   private static String removeMPPrefix(String fpath) {
```
guwirth: Do we really need this MP prefix removal as an extra step? Could be part of global regex?

cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/vc/CxxCompilerVcSensor.java
```diff
@@ -0,0 +1,73 @@
...
+import org.sonar.cxx.CxxMetricsFactory;
+import org.sonar.cxx.sensors.compiler.CxxCompilerSensor;
+
+public class CxxCompilerVcSensor extends CxxCompilerSensor {
+
+  public static final String KEY = "Visual C++";
+  public static final String REPORT_PATH_KEY = "vc.reportPath";
+  public static final String REPORT_REGEX_DEF = "vc.regex";
+  public static final String REPORT_CHARSET_DEF = "vc.charset";
+  public static final String DEFAULT_CHARSET_DEF = "UTF-8"; // use "UTF-16" for VS2010 build log or TFS Team build log file
```
guwirth: What is the latest charset for VS2017?

cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/vc/CxxCompilerVcSensor.java
```diff
@@ -0,0 +1,73 @@
...
+  // search for single line with compiler warning message VS2008 - order for groups: 1=file, 2=line, 3=ID, 4=message
+  public static final String DEFAULT_REGEX_DEF = "^(.*)\\((\\d+)\\)\\x20:\\x20warning\\x20(C\\d+):(.*)$";
```
guwirth: What is the regex for VS2017?

cxx-squid/src/main/java/org/sonar/cxx/CxxMetricsFactory.java
```diff
@@ -52,7 +52,8 @@ private CxxMetricsFactory() {
     // Introduce metric keys for sensors - number of issues per file / per module
     CLANG_SA_SENSOR_ISSUES_KEY("ClangSA"),
     CLANG_TIDY_SENSOR_ISSUES_KEY("Clang-Tidy"),
-    COMPILER_SENSOR_ISSUES_KEY("Compiler"),
+    VC_SENSOR_ISSUES_KEY("Compiler-VC"),
```
guwirth: Which metric keys should we use?
Compiler-XXX, or only XXX?

cxx-squid/src/main/java/org/sonar/cxx/CxxMetricsFactory.java
```diff
@@ -104,7 +105,8 @@ public String toString() {
 
     addSensorMetric(Key.CLANG_SA_SENSOR_ISSUES_KEY, "ClangSA issues", domain, langPropertiesKey, metrics);
     addSensorMetric(Key.CLANG_TIDY_SENSOR_ISSUES_KEY, "ClangTidy issues", domain, langPropertiesKey, metrics);
-    addSensorMetric(Key.COMPILER_SENSOR_ISSUES_KEY, "Compiler issues", domain, langPropertiesKey, metrics);
+    addSensorMetric(Key.VC_SENSOR_ISSUES_KEY, "VC compiler issues", domain, langPropertiesKey, metrics);
```
guwirth: Same here?
VC compiler issues
-or- only
VC issues

cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/gcc/CxxCompilerGccParser.java
ivangalkin: Since findWithinHorizon() is used as pre-requirement for match(), the MatchResult contains the proper results. Maybe we could assert the number of groups at runtime (matchres.groupCount() >=5). However it's sufficient to double-check the corresponding expression IMHO

cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/gcc/CxxCompilerGccParser.java
```diff
@@ -0,0 +1,70 @@
...
+  @Override
+  public void processReport(final SensorContext context, File report, String charset, String reportRegEx, List<Warning> warnings)
+    throws java.io.FileNotFoundException {
+
+    LOG.info("Parsing 'GCC' format ({})", charset);
+
+    try (Scanner scanner = new Scanner(report, charset)) {
+      Pattern p = Pattern.compile(reportRegEx, Pattern.MULTILINE);
+      LOG.debug("Using pattern : '{}'", p);
+      MatchResult matchres;
```
ivangalkin: I would move MatchResult matchres into the loop

cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/gcc/CxxCompilerGccSensor.java
ivangalkin: IMHO UTF-8 is de-facto standard for all OSs and can be used as default charset for each input file.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/gcc/CxxCompilerGccSensor.java
```diff
@@ -0,0 +1,75 @@
...
+  // ToDo: as long as java 7 API is not used the support of named groups for regular expression is not possible
```
ivangalkin: currently we use java8 and named regexp groups are supported since java7
so we can rewrite the comment as // TODO use named groups

cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/gcc/CxxCompilerGccSensor.java
```diff
@@ -0,0 +1,75 @@
...
+    super(language, REPORT_PATH_KEY, CxxCompilerGccRuleRepository.getRepositoryKey(language),
+      new CxxCompilerGccParser());
+  }
+
+  @Override
+  public void describe(SensorDescriptor descriptor) {
+    descriptor
+      .name(getLanguage().getName() + " CxxCompilerGccSensor")
+      .onlyOnLanguage(getLanguage().getKey())
+      .createIssuesForRuleRepositories(getReportPathKey())
```
ivangalkin: **ERROR**: please use `getRuleRepositoryKey()` (embarrassingly I've introduced it by myself)

cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/vc/CxxCompilerVcSensor.java
```diff
@@ -0,0 +1,73 @@
...
+  public CxxCompilerVcSensor(CxxLanguage language) {
+    super(language, REPORT_PATH_KEY, CxxCompilerVcRuleRepository.getRepositoryKey(language), new CxxCompilerVcParser());
+  }
+
+  @Override
+  public void describe(SensorDescriptor descriptor) {
+    descriptor
+      .name(getLanguage().getName() + " CxxCompilerVcSensor")
+      .onlyOnLanguage(getLanguage().getKey())
+      .createIssuesForRuleRepositories(getReportPathKey())
```
ivangalkin: **ERROR**: `getReportPathKey()` -> `getRuleRepositoryKey()`

cxx-sensors/src/test/java/org/sonar/cxx/sensors/compiler/gcc/CxxCompilerGccSensorTest.java
```diff
@@ -0,0 +1,64 @@
...
+import static org.mockito.Mockito.when;
+import org.sonar.api.batch.fs.internal.DefaultFileSystem;
+import org.sonar.api.batch.fs.internal.TestInputFileBuilder;
+import org.sonar.api.batch.sensor.internal.SensorContextTester;
+import org.sonar.api.config.internal.MapSettings;
+import org.sonar.cxx.CxxLanguage;
+import org.sonar.cxx.sensors.compiler.CxxCompilerSensor;
+import org.sonar.cxx.sensors.utils.TestUtils;
+
+public class CxxCompilerGccSensorTest {
```
ivangalkin: since we already have an issue in Sensor::describe() method, I would suggest to add 
```JAVA
  @Test
  public void sensorDescriptor() {
   ...
  }
```

both for GCC and VC sensors

cxx-squid/src/main/java/org/sonar/cxx/CxxMetricsFactory.java
ivangalkin: Compiler-XXX look more consistent to me

cxx-squid/src/main/java/org/sonar/cxx/CxxMetricsFactory.java
ivangalkin: VC compiler issues IMHO

guwirth:
@ivangalkin 

> @guwirth the following diff is mandatory. Should I push it as a small fix right-away? or will you include it in your PR? In any way the bug should be fixed before the next release. Thank you.

Please create an own PR. Think takes longer to finalize this one:
* default values
* regex group names
* compatibility
...

guwirth:
@ivangalkin need an regex expert :-).

Like to introduce named groups for the regex parser. What's the best approach to:
* read a complete file with an encoding
* search for an regex
* use  named groups instead of index

MatchResult does not support named groups.
Pattern is `Pattern.MULTILINE`, so reading it line by line is also not working.

Current code:
```Java
 try (Scanner scanner = new Scanner(report, reportCharset)) {
      Pattern p = Pattern.compile(reportRegEx, Pattern.MULTILINE);
      LOG.info("Using pattern : '{}'", p);
      while (scanner.findWithinHorizon(p, 0) != null) {
        MatchResult match = scanner.match();
        String filename = alignFilename(match.group(gFn).trim());
        String line = alignLine(match.group(gLine).trim());
        String id = alignId(match.group(gId).trim());
        String msg = alignMessage(match.group(gMsg).trim());
        if (isInputValid(filename, line, id, msg)) {
          if (LOG.isDebugEnabled()) {
            LOG.debug("Scanner-matches file='{}' line='{}' id='{}' msg={}", filename, line, id, msg);
          }
          CxxReportIssue issue = new CxxReportIssue(id, filename, line, msg);
          saveUniqueViolation(context, issue);
        } else {
          LOG.warn("Invalid compiler warning: '{}''{}'", id, msg);
        }
      }

```

ivangalkin:
Hi @guwirth,

unfortunately, I am not a big regex expert (TBH I am not a big Java expert either), but I'll try to help:

* read a complete file with an encoding + search for an regex: according to what I read, usage of `java.util.Scanner` seems to be a good choice. `Scanner` doesn't load the whole file into the memory, but rather buffers it and matches chunk-by-chunk (see https://stackoverflow.com/questions/10336478/does-the-scanner-class-load-the-entire-file-into-memory-at-once). So in case of large compiler logs, this should be less memory-intensive.

* use named groups instead of index: I don't have anything special to mention here. There are [a couple](https://www.robert-franz.com/2015/11/15/how-to-use-regex-groups-in-java-with-ease/) of [simple examples](https://www.robert-franz.com/2015/11/15/how-to-use-regex-groups-in-java-with-ease/), which might be helpful.

The only a little bit confusing pattern I found in your code snippet is `match.group(<group name>).trim()`. Why do we have to use `trim()`? Cannot we implement a group regexp in the way, that it doen't have leading/trailing spaces? E.g like https://stackoverflow.com/questions/38934328/regex-match-a-string-without-leading-and-trailing-spaces. But that's a minor comment.

guwirth:
@ivangalkin, @Bertk, @jmecosta changes done - ready to review again.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/squid/CxxSquidSensor.java
```diff
@@ -79,6 +76,14 @@
   public static final String MISSING_INCLUDE_WARN = "missingIncludeWarnings";
   public static final String JSON_COMPILATION_DATABASE_KEY = "jsonCompilationDatabase";
 
+  /**
+   * the following settings are in use by the feature to read configuration settings from the VC compiler report
+   */
+  public static final String REPORT_PATH_KEY = "compiler.reportPath";
+  public static final String REPORT_CHARSET_DEF = "compiler.charset";
+  public static final String PARSER_KEY_DEF = "compiler.parser";
```
ivangalkin: the only supported value for `compiler.parser` is "Visual C++"; (see sonar-cxx/cxx-squid/src/main/java/org/sonar/cxx/CxxConfiguration.java).  Maybe we could deprecate this property too?

sonar-c-plugin/src/main/java/org/sonar/plugins/c/CPlugin.java
```diff
@@ -339,57 +336,73 @@
   private static List<PropertyDefinition> compilerWarningsProperties() {
     String subcateg = "(4) Compiler warnings";
     return new ArrayList<>(Arrays.asList(
-      PropertyDefinition.builder(LANG_PROP_PREFIX + CxxCompilerSensor.REPORT_PATH_KEY)
-        .name("Compiler report(s)")
+      PropertyDefinition.builder(LANG_PROP_PREFIX + CxxCompilerVcSensor.REPORT_PATH_KEY)
```
ivangalkin: it looks like we don't we miss the following properties now:

```JAVA
  public static final String REPORT_PATH_KEY = "compiler.reportPath";
  public static final String REPORT_CHARSET_DEF = "compiler.charset";
  public static final String PARSER_KEY_DEF = "compiler.parser";
```

as far as I see, they are still used in order to set the project configuration. Is it your intent to hide them?

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxPlugin.java
```diff
@@ -45,13 +44,10 @@
...
 import org.sonar.cxx.sensors.clangtidy.CxxClangTidyRuleRepository;
 import org.sonar.cxx.sensors.clangtidy.CxxClangTidySensor;
-import org.sonar.cxx.sensors.compiler.CxxCompilerGccParser;
-import org.sonar.cxx.sensors.compiler.CxxCompilerGccRuleRepository;
-import org.sonar.cxx.sensors.compiler.CxxCompilerGccSensor;
-import org.sonar.cxx.sensors.compiler.CxxCompilerSensor;
-import org.sonar.cxx.sensors.compiler.CxxCompilerVcParser;
-import org.sonar.cxx.sensors.compiler.CxxCompilerVcRuleRepository;
-import org.sonar.cxx.sensors.compiler.CxxCompilerVcSensor;
+import org.sonar.cxx.sensors.compiler.gcc.CxxCompilerGccRuleRepository;
```
ivangalkin: same for c++ plugin

sonar-c-plugin/src/main/java/org/sonar/plugins/c/CPlugin.java
guwirth: Thx for this hint. They are really missing now.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/CxxCompilerSensor.java
```diff
@@ -20,73 +20,136 @@
...
+   * Derived classes can overload this method
+   *
+   * @param filename
+   * @param line
+   * @param id
+   * @param msg
+   * @return true, if valid
+   */
+  protected boolean isInputValid(String filename, String line, String id, String msg) {
+    return !filename.isEmpty() || !line.isEmpty() || !id.isEmpty() || !msg.isEmpty();
```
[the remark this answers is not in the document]
guwirth: @ivangalkin  The idea of the plugin is to break if there is a configuration error and to give only warnings if there are (partial) problems with a report. So if there is one (several) wrong messages at least the other are reported.

Or what exactly do you expect in such a case?

cxx-sensors/src/main/java/org/sonar/cxx/sensors/squid/CxxSquidSensor.java
guwirth: @ivangalkin idea was to change only one feature with this PR and not also the "read configuration". In general there is still #1365.
@Bertk What are you thinking?

guwirth:
@ivangalkin I added missing UI properties "compiler.reportPath" and "compiler.charset". Please review again.

@Bertk maybe we should rename them as well (compiler => msbuild):

```JAVA
  public static final String REPORT_PATH_KEY = "msbuild.reportPath";
  public static final String REPORT_CHARSET_DEF = "msbuild.charset";
```

jmecosta:
@guwirth if this is braking change, that it is. Maybe its best to get ride of
sonar.cxx.compiler.reportPath
sonar.cxx.compiler.charset
altogether.

guwirth:
> if this is braking change, that it is. Maybe its best to get ride of sonar.cxx.compiler.reportPath sonar.cxx.compiler.charset altogether.

Would be renamed to then
```
sonar.cxx.msbuild.reportPath
sonar.cxx.msbuild.charset 
```

klimkin:
For some reason I am unable to activate compiler rules anymore. Any idea? All other sensors activate ok.

    20:09:33.581 DEBUG: 'C++ (Community) CxxCompilerGccSensor' skipped because there is no related rule activated in the quality profile
    20:09:33.581 DEBUG: 'C++ (Community) CxxCompilerVcSensor' skipped because there is no related rule activated in the quality profile

    .scannerwork/scanner-report/analysis.log:  - sonar.cxx.gcc.charset=yyy
    .scannerwork/scanner-report/analysis.log:  - sonar.cxx.gcc.regex=xxx
    .scannerwork/scanner-report/analysis.log:  - sonar.cxx.gcc.reportPath=build/build.log

guwirth:
@klimkin what do you see in the rules section of the UI? Are the rules still available and activated?

klimkin:
@guwirth Administration->General Settings->C++(Community) all settings are there.

Surprisingly,  sonar.cxx.cppcheck.reportPath is just enough to see DEBUG logs trying to load the file. sonar.cxx.gcc.reportPath and sonar.cxx.clangtidy.reportPath do not have any effect. I was staring at cppcheck and clangtidy sensor sources for quite time and don't see anything that might cause the difference in behavior!

klimkin:
Ahh, you are right. The rules were inactive by default. Thanks!

ivangalkin:
> I was staring at cppcheck and clangtidy sensor sources for quite time and don't see anything that might cause the difference in behavior!

@guwirth that's one more reason to make the behavior consistent: see #1527

Issue #1547, opened 2018-08-27
Complexity metric / checks must be performed on original source code


ivangalkin:
* complexity metrics "Cognitive Complexity", "Cyclomatic Complexity" as well as derivate from cyclomatic complexity metrics "Complex Function (NR)", "Complex Functions (%)", "Complex Functions Lines of Code", "Complex Functions Lines of Code" and...
* complexity checks "FunctionCognitiveComplexityCheck", "FunctionComplexity", "ClassComplexity", "FileComplexity"...

...are calculated/performed on preprocessed code. The preprocessor falsifies the code complexity

* expansion of macros makes the code more complex
* evaluation of preprocessor directives like `#if` `#ifdef`, `#else` etc makes the code less complex

Example of how macro expansion affects the check `FunctionComplexity`: `EXPECT_EQ` macro adds +3 extra cyclomatic complexity points because it's expanded as `switch-case` + `if`.

![complexity_vs_macro](https://user-images.githubusercontent.com/11503715/44644971-37427d80-a9d6-11e8-86ac-9290147d9551.png)

Possible solutions:
* keep information, which AST parts were added while preprocessing (marking of some AST nodes as generated, helps only by macro expansions)
* keep both original and preprocessor code for each compile unit (create two distinct ASTs)
* perform some set of visitors before preprocessor and some set after it (there might be more visitors spoiled by preprocessor, e.g. the syntax-highlighting one)

Possible problems:
* creation of AST for non-preprocessed (original) source code might be tricky: While macro expansion can be mapped to the function-invocation-AST (?), not-evaluation of preprocessor directives can lead to unparseable code (?)

guwirth:
@ivangalkin thx for investigations. Meanwhile I also tend to evaluate only "what you see later on in the source code view" (means without preprocessing the code).

> perform some set of visitors before preprocessor and some set after it (there might be more visitors spoiled by preprocessor, e.g. the syntax-highlighting one)

@all: Where do you think do we still need preprocessed code?

jmecosta:
I dont think so, does anyone know when do we need a "full" ast?

ivangalkin:
@guwirth @guwirth I came to conclusion, that current visitors (all metric calculators and checks are AST visitors) cannot be applied to the original (non-preprocessed code). Original code might be non-parseable at all (there are far too many tricks one can make with macros and complier directives).

[...]

PR #1552, opened 2018-09-05
exclude macro expansions from complexity metrics/checks

ivangalkin:
* partial solution for #1547

MOTIVATION:

* #1494 made visible the fact, that preprocessor affects the complexity
metric. E.g. if function uses macros, the underlying (hidden) complexity
of theses macros increases the complexity of the function. This is wrong
because of many reasons:

1. Cognitive complexity:
Macros do not increase the cognitive complexity of the source code. To
the reader they are similar to the function invocations.
Example ASSERT_XX/EXPECT_XX macros of GTest.

2. Cyclomatic complexity:
Similar logic might be applied to the cyclomatic complexity: macro expansions
are similar to the function calls (which become inlined). In this case
the cyclomatic complexity of the caller should not be increased.
[Cyclomatic] complexity of the callee must be calculated separately.

3. Preprocessor doesn't distinguish between the macros, which
were declared inside of the analyzed project or were included from the
external header. Obviously the external source code (e.g. again very
complex GTest macros) should not affect the complexity of someone's
custom SonarQube project.

SOLUTION:

In general it's impossible to build a valid C/C++ AST without
performing of preprocessor step. The original source code might
be syntactically incorrect (e.g. if macros are used to "invent" some
domain specific language out of C/C++). All metrics of sonar-cxx
are however AST-based. (This is a very questionably decision
e.g. because #if/#ifdef etc might affect the resulting complexity or [NC]LOC
in dramatic way). So we cannot skip preprocessor step entirely and/or
perform metrics calculation/checks on the non-preprocessed code.

Sonar SSLR has a possibility to mark some tokens as generated.
This property (`Token::isGeneratedCode()`) is for example used
in order to ignore generated while
* [NC]LOC counting (see `AbstractLineLengthCheck<>`)
* CDP analysis (see `CxxCpdVisitor`) or
* Highlighting (see `CxxHighlighterVisitor`)

This patch ensures, that
* all `AstNode`s, which were created while macro expansion are marked as
generated
* all complexity related visitors ignore generated `AstNode`s

RESULT:

PRO:
1. The complexity of "callers" is not affected by macro anymore.
   This fact makes the complexity check look logical again.

CON:
1. the complexity of "callees" is not calculated at all. This
is correct in case of external macros, but wrong w.r.t. to the
project intern macros. Possible solution: preprocessor must try
to extract the function-like macro declaration into some special AST-subtree.
Afterwards complexity metrics/check will be able to calculate/check
macros as separate language construct.

By the way: it must be analyzed if and how macros affect other metrics.
For example LOC etc.

2. This patch might change the complexity related metrics. Depending
on how many macros were used in the analyzed project, the reduction
can be radical.

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1552)

### DIFF
--- a/cxx-checks/src/main/java/org/sonar/cxx/checks/metrics/CxxCyclomaticComplexityCheck.java
+++ b/cxx-checks/src/main/java/org/sonar/cxx/checks/metrics/CxxCyclomaticComplexityCheck.java
@@ -97,6 +97,10 @@ public abstract class CxxCyclomaticComplexityCheck<G extends Grammar> extends Mu
 
   @Override
   public void visitNode(AstNode astNode) {
+    if (astNode.getToken().isGeneratedCode()) {
+      return;
+    }
+
     final Optional<AstNodeType> scopeType = getScopeType();
     if (scopeType.isPresent() && astNode.is(scopeType.get())) {
       complexityScopes.addFirst(new CxxComplexityScope(astNode.getTokenLine()));
@@ -113,6 +117,10 @@ public abstract class CxxCyclomaticComplexityCheck<G extends Grammar> extends Mu
 
   @Override
   public void leaveNode(AstNode astNode) {
+    if (astNode.getToken().isGeneratedCode()) {
+      return;
+    }
+
     final Optional<AstNodeType> scopeType = getScopeType();
     if (scopeType.isPresent() && astNode.is(scopeType.get())) {
       analyzeScopeComplexity();
--- a/cxx-checks/src/test/resources/checks/FunctionCognitiveComplexity.cc
+++ b/cxx-checks/src/test/resources/checks/FunctionCognitiveComplexity.cc
@@ -110,3 +110,31 @@ std::string func_with_trycatch(int i) {
       return (i % 2 == 0) ? "exotic exception" : "strange exception";
    }
 }
+
+#define IMPLEMENTATION()                                                      \
+   try {                                                                      \
+      if ((i % 2 == 0 && i % 3 == 0) || (i % 6 == 0)) {                       \
+         return std::to_string(i);                                            \
+      }                                                                       \
+   } catch (const std::logic_error& e) {                                      \
+      return "std::logic_error";                                              \
+   } catch (const std::bad_cast& e) {                                         \
+      return std::string { "std::bad_cast" };                                 \
+   } catch (const std::invalid_argument& e) {                                 \
+      return std::string { "std::invalid_argument" };                         \
+   } catch (const std::length_error& e) {                                     \
+      return std::string { "std::length_error" };                             \
+   } catch (const std::out_of_range& e) {                                     \
+      return std::string { "std::out_of_range" };                             \
+   } catch (const std::exception& e) {                                        \
+      return std::string("std::exception");                                   \
+   } catch (...) {                                                            \
+      while (i >= 0) {                                                        \
+         return (i % 3 == 0) ? "unknown exception" : "unexpected exception";  \
+      }                                                                       \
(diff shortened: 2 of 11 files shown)


guwirth:
> the complexity of "callees" is not calculated at all. This is correct in case of external macros, but wrong w.r.t. to the project intern macros.

@ivangalkin not sure if this is really a problem. 'Complexity' is normally 'Readability' (for sure Cognitive complexity). Not sure how other tools are doing this for 'Cyclomatic complexity' in case of C/C++ (with / without prepossessing it). Think it's quite intuitive to calculate only the 'visible' code. Imaging the macro would be a function/method call it's also not counting. So why counting a macro?

guwirth:
@ivangalkin also like to make a sample:

```C++
// currently we don't calculate the complexity of macros
// we expand it and calculate the complexity inside user of the macro
// ==> new: handle this like a function?
#define MACRO(a, b) \
   if (a)  { \
      //... \
   else { \
     //... \
   }

// complexity of func1 is calculated
void func1(int a, int b)
{
   if (a)  { 
      //...
   else {
     //...
   }
}

// in this sample only the complexity of func2 is calculated
// complexity of called functions / methods do not count inside of this function
void func2()
{
    int a=0, b=0;
    if ( /*..*/ ) {
       func1(a, b);
    }
}

// in the past we did a macro expansion and were calculating the resulting complexity
// but I think there is (should be) no difference to func2?
void func3 ()
{
    int a=0, b=0;
    if ( /*..*/ ) {
       MACRO(a, b);
    }
}
```

ivangalkin:
@guwirth I agree with your example. Complexity of `func2()` and `func3()` should be the same (and it will be the same as result of this PR). I believe, that complexity of `MACRO` should be somehow added to the overall complexity of the project e.g. by treating them as function-definition-like nodes. The advantage of such one-time calculation will be that complexity of `MACRO` is added only once. Currently it might be misleadingly multiplied. Example:

```CPP
void func2()
{
    int a=0, b=0;
    if ( /*..*/ ) {
       func1(a, b); // function-call is not considered to be a complexity source
       func1(a, a); // so 3 functions calls still add nothing to the function complexity
       func1(b, b);
    }
}

void func3 ()
{
    int a=0, b=0;
    if ( /*..*/ ) {
       MACRO(a, b); // we "incline" the complexity of MACRO, but not just once but 3 times (!)
       MACRO(a, a); // this might increase the overall project complexity and cause complexity warnings
       MACRO(b, b); // it also contradicts the func2 example 
    }
}
```

guwirth:
@all someone a different opinion?

> ...complexity of MACRO should be somehow added to the overall complexity of the project

@ivangalkin do you see a chance to do this?

guwirth:
@all like to merge this this evening. No feedback means ok 😄.

ivangalkin:
@guwirth sorry for the delay and thank you for giving this PR a chance

> @ivangalkin do you see a chance to do this?

There is a little chance, that `token-stringopt` (in terms of [MSDN](https://msdn.microsoft.com/de-de/library/teas0593.aspx)) is a parseable AST subtree. My naive idea was to instrument the preprocessor in the way, that...

```C++
#define identifier( identifieropt, ... , identifieropt ) token-stringopt 
```
... becomes ...

```C++
void identifier( void* identifieropt, ... , void* identifieropt ) { token-stringopt; }
```

... so that our parser could try to parse it. If it succeeds, we have a function declaration, which complexity might be analyzed in the standard way. If not - we just ignore it.

I've played around with the preprocessor, but could not manage to get some results yet. We can create an user story / issue as a reminder. Maybe somebody has a better idea.

PR #1558, opened 2018-09-21
Update clang-tidy rules; Tune severities

### LINKED ISSUE #1536: US: Clang-Tidy 8 support (update clangtidy.xml)
### Description

https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/resources/clangtidy.xml is a bit out of date with http://clang.llvm.org/extra/clang-tidy/checks/list.html

### Questions for implementation

Most of the required data is pulled straight from clang.llvm.org with a couple of exceptions:
- we seem to only have a subset of the more relevant ones (I'm happy to make a call on selecting the more generic C/C++ rules rather than random specific project rules like `abseil-string-find-startswith`)
- severity (INFO, MINOR, MAJOR, CRITICAL) and type (BUG or CODE_SMELL) is where I come unstuck - is this decided just fairly arbitrarily?

ivangalkin:
Apply custom severities, suggested by @nathanaelg

**ATTENTION!** severity of the following exiting rule was lowered
  * cppcoreguidelines-no-malloc

please review!

misc: original documentation contained internal hyperlinks, change them to external ones

#1536 ( still doesn't close the issue, we need more rules mentioned in #1560)

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1558)

### DIFF
--- a/cxx-sensors/src/main/resources/clangtidy.xml
+++ b/cxx-sensors/src/main/resources/clangtidy.xml
@@ -218,8 +218,8 @@ struct some_struct* s = (struct some_struct*) malloc(sizeof(struct some_struct))
 ]]></description>
   <tag>bad-practice</tag>
   <tag>suspicious</tag>
-  <severity>MAJOR</severity>
-  <type>BUG</type>
+  <severity>INFO</severity>
+  <type>CODE_SMELL</type>
 </rule>
 <rule>
   <key>cppcoreguidelines-pro-bounds-array-to-pointer-decay</key>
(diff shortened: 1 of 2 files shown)


guwirth:
@nathanaelg please add also your comments

nathanaelg:
I had put `cppcoreguidelines-no-malloc` as info code smell because it was one of the "don't use this feature of the language" even though if you use it properly it isn't necessarily problematic.

On further thought though, it is one of those things that is fairly easy to get wrong and the consequences are fairly bad so might be best to have it as a major code smell - I'm still not convinced it should be classed as a bug rather than code smell.

PR #1563, opened 2018-10-02
reduce amount of tracing for CxxParseErrorLoggerVisitor

ivangalkin:
* I profiled our visitors (#1507) and came to the conclusion,
  that CxxParseErrorLoggerVisitor might require a terrible amount
  of time in order to analyze and to print out all the parsing errors

* The ordinary user should not be interested in the parsing process.
  He/she has no influence on that. Such information is important
  if parsing affects some internal checks or metrics. If user decides
  to submit a bug report, the full debug trace will include the
  parsing errors.

* In general I believe, that our trace is too verbose. So this
  change address this problem too.

**UPDATE**:
* review comments were applied: once per module there will be a
  WARN trace about parsing errors (if relevant). Errors details
  will be printed only in DEBUG

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1563)


guwirth:
@ivangalkin not really sure what we should do here:
* Providing metrics is a basic functionality of the plugin. Showing wrong metrics is an error. Showing the root causes as warnings is already a compromise.
* On the other hand you are right: if someone don’t provide the right includes he is also not interested in these warnings.

Maybe a compromise could be to show **one** warning that there were parsing problems and show the details only if debug informations are on?

cxx-squid/src/main/java/org/sonar/cxx/visitors/CxxParseErrorLoggerVisitor.java
```diff
@@ -64,7 +74,7 @@ public void visitNode(AstNode node) {
       } else if (type.equals(CxxPunctuator.CURLBR_LEFT)) {
         // part with CURLBR_LEFT is typically an ignored declaration
         if (identifierLine != -1) {
-          LOG.warn("[{}:{}]: skip declaration: {}",
+          LOG.info("[{}:{}]: skip declaration: {}",
```
guwirth: would use LOG.debug

cxx-squid/src/main/java/org/sonar/cxx/visitors/CxxParseErrorLoggerVisitor.java
```diff
@@ -79,7 +89,7 @@ public void visitNode(AstNode node) {
 
     if (identifierLine != -1 && sb.length() > 0) {
       // part without CURLBR_LEFT is typically a syntax error
-      LOG.warn("[{}:{}]:    syntax error: {}",
+      LOG.info("[{}:{}]:    syntax error: {}",
```
guwirth: would use LOG.debug

cxx-squid/src/main/java/org/sonar/cxx/visitors/CxxParseErrorLoggerVisitor.java
```diff
@@ -42,6 +43,15 @@ public CxxParseErrorLoggerVisitor(SquidAstVisitorContext<?> context) {
     this.context = context;
   }
 
+  /**
+   * The only thing this visitor does is logging with severity == INFO; Disable
+   * the entire visitor if logging level was set to the level > INFO
+   */
+  public static boolean isLoggingEnabled() {
```
guwirth: logger level is global, don’t need local method?

cxx-squid/src/main/java/org/sonar/cxx/CxxAstScanner.java
```diff
@@ -229,7 +229,9 @@ public SourceCode createSourceCode(SourceCode parentSourceCode, AstNode astNode)
     builder.withSquidAstVisitor(new CxxFileVisitor<>(context));
 
     // log syntax errors
-    builder.withSquidAstVisitor(new CxxParseErrorLoggerVisitor<>(context));
+    if (CxxParseErrorLoggerVisitor.isLoggingEnabled()) {
```
guwirth: would remove this

guwirth:
@ivangalkin will make a proposal

guwirth:
use #1565 instead

PR #1565, opened 2018-10-04
make CxxParseErrorLoggerVisitor less verbose

guwirth:
- alternative solution for #1563

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1565)

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/visitors/CxxParseErrorLoggerVisitor.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/visitors/CxxParseErrorLoggerVisitor.java
@@ -37,8 +37,7 @@ public class CxxParseErrorLoggerVisitor<GRAMMAR extends Grammar>
 
   private static final String SYNTAX_ERROR_MSG
     = "Source code parser: {} syntax error(s) detected. Syntax errors could cause invalid software metric values."
-    + " Root cause are typically missing includes, missing macros or compiler specific extensions."
-    + " Turn debug info on to get more details.";
+    + " Root cause are typically missing includes, missing macros or compiler specific extensions.";
   private static final Logger LOG = Loggers.get(CxxParseErrorLoggerVisitor.class);
   private final SquidAstVisitorContext<?> context;
   private static int errors = 0;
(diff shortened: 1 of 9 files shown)


ivangalkin:
BTW there is test error on some older (?) Sonar/API versions

```
[ERROR] handleParseErrorLoggerLevelDebugTest(org.sonar.cxx.visitors.CxxParseErrorLoggerVisitorTest)  Time elapsed: 0.061 s  <<< FAILURE!
org.junit.ComparisonFailure: expected:<1[3]> but was:<1[2]>
	at org.sonar.cxx.visitors.CxxParseErrorLoggerVisitorTest.handleParseErrorLoggerLevelDebugTest(CxxParseErrorLoggerVisitorTest.java:59)
```

Not sure, but maybe it's because of `org.sonar.api.utils.log.LogTester::logs()` method. The separation of log entries by level seemed to work for me:

```JAVA
List<String> warnEntries = logTester.logs(LoggerLevel.WARN);
List<String> debugEntries = logTester.logs(LoggerLevel.DEBUG);
```

guwirth:
@ivangalkin I added a post job. Should ensure that warning is at the end.
Maybe we can use the same pattern to reduce "#include" warnings?

cxx-squid/src/main/java/org/sonar/cxx/visitors/CxxParseErrorLoggerVisitor.java
```diff
@@ -36,28 +36,31 @@
...
   private final SquidAstVisitorContext<?> context;
-  private static boolean firstError = true;
+  private static int errors = 0;
 
   public CxxParseErrorLoggerVisitor(SquidAstVisitorContext<?> context) {
     this.context = context;
   }
 
+  public static void finalReport() {
+    if (!LOG.isDebugEnabled() && (errors > 0)) {
```
ivangalkin: I think it would be more consistent to log the WARN message always: 
instead of `if (!LOG.isDebugEnabled() && (errors > 0)) {` just `if (errors != 0)` ?

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxPlugin.java
```diff
@@ -585,7 +583,7 @@ public void define(Context context) {
     l.add(CxxClangTidySensorImpl.class);
     l.add(CxxClangSASensorImpl.class);
     l.add(CxxExternalRulesSensorImpl.class);
-
+
```
ivangalkin: trailing whitespaces

guwirth:
@ivangalkin thx for your comments. Made an update with your comments.

ivangalkin:
@guwirth thank you! I have a couple of additional minor comments; feel free to ignore

PR #1575, opened 2018-10-12
path normalization / absolutization is not allowed to use File::getCanonicalPath()

### LINKED ISSUE #1465: coverage cache vs multi-module projects

### LINKED ISSUE #1574: Sensors have problem to import results because of symbolic links and path normalization
I use sonarqube 7.3 and sonar-cxx 1.1.0  and cppcheck 1.84

I use cppcheck scan src code result with cppcheck-result-1.xml  eg:
image
When i startup sonar-scnaner to parse cppcheck-result-1.xml , raise the issue: 
Cannot find the file 'src/targeting_service_dumpads.cpp', skipping violations

I lookup the official wiki: but can not resolve the problem
my sonar-project.conf such as:
image

Then My sonar root path such as:
image

any one can help me sovle this issue？  Thank you very much

@guwirth  can help me ?

ivangalkin:
Path normalization / absolutization is not allowed to use `File::getCanonicalPath()`. If performed on Linux (Unix), this method resolves the symbolic links. This fact might make the queries like "is this normalized path in SonarQube project" go wrong.

Fixed #1574

* review the SonarQube's filesystem queries
* no path absolutization is necessary; filesystem deals perfectly with relative paths

Fixes #1465 

* `CoberturaParser` used  `CxxUtils::normalizePathFull()`
* This made all file paths in parsed report to  absolute ones. Base directory of the first processed module was used in order to make paths absolute
* Since parsed reports were cached, this would lead to the wrong absolute paths in multi-module project
* Store original paths instead!

Misc:

* minor refactoring
* remove error-prone implementation of `CxxUtils::normalizePathFull()`

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1575)

### DIFF
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/BullseyeParser.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/BullseyeParser.java
@@ -28,7 +28,6 @@ import javax.annotation.Nullable;
 import javax.xml.stream.XMLStreamException;
 import org.codehaus.staxmate.in.SMHierarchicCursor;
 import org.codehaus.staxmate.in.SMInputCursor;
-import org.sonar.api.batch.sensor.SensorContext;
 import org.sonar.api.utils.PathUtils;
 import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
@@ -45,14 +44,14 @@ public class BullseyeParser extends CxxCoverageParser {
   private static int totalcoveredconditions;
 
   public BullseyeParser() {
-    // no operation but necessary for list of coverage parsers 
+    // no operation but necessary for list of coverage parsers
   }
 
   /**
    * {@inheritDoc}
    */
   @Override
-  public void processReport(final SensorContext context, File report, final Map<String, CoverageMeasures> coverageData)
+  public void processReport(File report, final Map<String, CoverageMeasures> coverageData)
     throws XMLStreamException {
     LOG.debug("Parsing 'Bullseye' format");
     StaxParser topLevelparser = new StaxParser(new StaxParser.XmlStreamHandler() {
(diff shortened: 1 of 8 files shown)


guwirth:
@jmecosta ok for me. Do you have any comments? Think you had some special cases in your CI?

jmecosta:
@guwirth i think we have integration tests for bullseye in place, it should be ok by me also. i will raise issue if something goes bad in our instances. i dont expect so

ericlemes:
Hi @guwirth and @ivangalkin . I've recently bumped in this issue in 1.1.0 and seems to be fixed by this PR. Any plans for releasing this?

guwirth:
Hi @ericlemes, will be part of 1.2, see #1579. You are welcome to try it out and give feedback.

Issue #1544, opened 2018-08-24
API check: Method defined in header file, but out of class, marked as undocumented


jspam:
### Description

A method that is declared and documented inside a class, but defined outside the class in the header file, is marked as undocumented API.

```c++
class A {
public:
  /** Documented */
  void f();
}

// wrongly marked as undocumented
inline void A::f() {}
```

PR #1543 contains an example of such a definition.

In Doxygen, the documentation is correctly taken from the definition inside the class.

### Steps to reproduce the problem

Analyze `public_api.h` from #1543 with sonar-cxx

### Expected behavior

The method definition in line 111 (`inline void testClass::publicDeclaredMethod() {`) is not marked as undocumented.

### Actual behavior

The method definition in line 111 is marked as undocumented (`Undocumented API: testClass`).

### LOG file

```
...
06:45:44.554 DEBUG - node: declaratorId line: 108 id: 'inlineCommentedLastAttr' documented: true
06:45:44.555 DEBUG - ***** Node: functionDefinition tokenValue='inline' tokenLine=111 tokenColumn=0
06:45:44.556 DEBUG - Public API: testClass
06:45:44.557 DEBUG - node: declaratorId line: 111 id: 'testClass' documented: false
06:45:44.557 DEBUG - ***** Node: functionDefinition tokenValue='inline' tokenLine=111 tokenColumn=0
06:45:44.558 DEBUG - Public API: testClass
06:45:44.559 DEBUG - node: declaratorId line: 111 id: 'testClass' documented: false
06:45:44.560 DEBUG - ***** Node: classSpecifier tokenValue='struct' tokenLine=117 tokenColumn=0
06:45:44.561 DEBUG - Doc: /// testStruct doc
06:45:44.562 DEBUG - Public API: testStruct
06:45:44.563 DEBUG - node: className line: 117 id: 'testStruct' documented: true
06:45:44.563 DEBUG - ***** Node: classSpecifier tokenValue='struct' tokenLine=117 tokenColumn=0
06:45:44.564 DEBUG - Doc: /// testStruct doc
...
```

### Related information

* cxx plugin version? 1.1.0
* SonarQube version? 6.7.4.38452

guwirth:
Hi @jspam,

thanks for providing this sample.

I see two possible solutions:
* The easy one: ignore all function/methods marked "inline". Think will ignore too much?
[...]

PR #1580, opened 2018-10-27
Public API: ignore methods outside of class

guwirth:
* add exception: Function definitions with a qualified name outside of a class. Assumption is that they are already documented inside of the class.
* close #1543, #1544

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1580)


ivangalkin:
@guwirth detection of outside-of-the-class declarations are similar to the one from `MethodNameCheck`. So the change looks good to me. My comments are very minor. Feel free to ignore.

guwirth:
@ivangalkin thx for your review. Tried to fix it. Is the description now better?

PR #1599, opened 2018-11-24
support xunit reports without testsuite

### LINKED ISSUE #1595: xUnit sensor couldn't parse test report with 0 testsuite tag
### Description

When an xUnit test report does not have any testsuite tag, xUnit sensor throws exception and could not continue to parse other reports.

### Steps to reproduce the problem

Analyze a test report as follows:
```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites tests="0" failures="0" disabled="0" errors="0" timestamp="2018-11-20T11:37:41" time="0" name="AllTests">
</testsuites>
```
The test program contains no actual tests but it serves as a stub for future development.
This report was created by gtest with option --gtest_output=xml:test_report.xml

### Expected behavior

Expect the report should be parsed without error.

### Actual behavior
xUnit sensor throws exception and could not continue to parse other reports.

### Known workarounds
Perhaps adding a null check at cxx-sensors/src/main/java/org/sonar/cxx/sensors/tests/xunit/XunitReportParser.java:61 may fix this problem but I haven't verified it yet. 

    do {
      parseTestSuiteTag(testSuiteCursor);
    } while (testSuiteCursor.getNext() != null);

### LOG file

```
11:37:56.061 INFO: Scanner found '3' report files
11:37:56.061 INFO: Parser will parse '3' report files
11:37:56.075 INFO: Processing report '/var/jenkins_home/workspace/playground-build-test-scan/report/blank_test_report.xml'
11:37:56.075 DEBUG: Transformation skipped: no xslt given
11:37:56.110 ERROR: Cannot feed the data into SonarQube, details: 'javax.xml.stream.XMLStreamException: ParseError at [row,col]:[4,1]
Message: Can not call 'getAttributeValue(): cursor does not point to a valid node (curr event [null]; cursor state CLOSED)'
11:37:56.111 INFO: Sensor C++ (Community) XunitSensor [cxx] (done) | time=79ms
```

### Related information

* cxx plugin version? 1.1.0
* SonarQube version? 7.4

guwirth:
- close #1595

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1599)

### DIFF
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/tests/xunit/XunitReportParser.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/tests/xunit/XunitReportParser.java
@@ -52,14 +52,12 @@ public class XunitReportParser implements XmlStreamHandler {
   public void stream(SMHierarchicCursor rootCursor) throws XMLStreamException {
     SMInputCursor testSuiteCursor = rootCursor.constructDescendantCursor(new ElementFilter("testsuite"));
     try {
-      testSuiteCursor.getNext();
+      while (testSuiteCursor.getNext() != null) {
+        parseTestSuiteTag(testSuiteCursor);
+      }
     } catch (com.ctc.wstx.exc.WstxEOFException eofExc) {
       throw new EmptyReportException("Cannot read Xunit report", eofExc);
     }
-
-    do {
-      parseTestSuiteTag(testSuiteCursor);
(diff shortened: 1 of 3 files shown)


guwirth:
@ivangalkin think the difference is
* empty / corrupt XML => exception
* or iterating over testsuites (with special case no testsuite)

PR #1605, opened 2018-11-27
Eliminate wrongly used static variables

ivangalkin:
There is no need to share state between classes. This is error-prone and
prohibits any possible parallelization of code analysis

* BullseyeParser
* HardcodedAccountCheck - refactored
* HardcodedIpCheck - refactored

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1605)

### DIFF
--- a/cxx-checks/src/main/java/org/sonar/cxx/checks/HardcodedAccountCheck.java
+++ b/cxx-checks/src/main/java/org/sonar/cxx/checks/HardcodedAccountCheck.java
@@ -24,6 +24,8 @@ import com.sonar.sslr.api.Grammar;
 import java.util.Objects;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
 import org.sonar.check.Priority;
 import org.sonar.check.Rule;
 import org.sonar.check.RuleProperty;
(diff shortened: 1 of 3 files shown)


cxx-checks/src/main/java/org/sonar/cxx/checks/HardcodedIpCheck.java
```diff
@@ -54,42 +56,35 @@
...
 
-  @SuppressWarnings("squid:S2696") // ... initialize SquidAstVisitor
   @Override
   public void init() {
-    String regEx = getRegularExpression();
-    Objects.requireNonNull(regEx, "getRegularExpression() should not return null");
+    Objects.requireNonNull(regularExpression, "getRegularExpression() should not return null");
 
-    if (!regEx.isEmpty()) {
+    if (!regularExpression.isEmpty()) {
```
ivangalkin: same here

cxx-checks/src/main/java/org/sonar/cxx/checks/HardcodedAccountCheck.java
```diff
@@ -55,45 +57,35 @@
...
+  private Pattern pattern;
 
-  @SuppressWarnings("squid:S2696") // ... initialize SquidAstVisitor
   @Override
   public void init() {
-    String regEx = getRegularExpression();
     Objects.requireNonNull(regularExpression, "getRegularExpression() should not return null");
 
-    if (null != regEx && !regEx.isEmpty()) {
+    if (!regularExpression.isEmpty()) {
```
ivangalkin: probably throwing an `IllegalStateException` in case of empty regular expression would be more correct

guwirth: @ivangalkin general rule we had in the past:
* configuration errors should result in errors and "stop" the analysis
* runtime errors: https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Supported-configuration-properties#sonarcxxerrorrecoveryenabled

Wrong regex is configuration error.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/BullseyeParser.java
```diff
@@ -39,9 +39,9 @@
 public class BullseyeParser extends CxxCoverageParser {
 
   private static final Logger LOG = Loggers.get(BullseyeParser.class);
-  private static String prevLine;
```
guwirth: @ivangalkin don't know why they were static in the past? Seems ok now.

cxx-checks/src/main/java/org/sonar/cxx/checks/HardcodedAccountCheck.java
ivangalkin: Ok. Then I'll throw an exception if regular expression was empty

PR #1608, opened 2018-11-30
organize members

guwirth:
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1608)


guwirth:
Next try. Should be in SonarCloud order now.

PR #1609, opened 2018-12-01
Test FP gherkin:missing-data-table-column

Bertk:
- FP gherkin:missing-data-table-column
- add escape sequence for scenarios with Examples (workaround option)

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1609)


Bertk:
Check was OK.

PR #1611, opened 2018-12-02
Fix quality flaws 3

guwirth:
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1611)

### DIFF
--- a/cxx-checks/src/main/java/org/sonar/cxx/checks/HardcodedAccountCheck.java
+++ b/cxx-checks/src/main/java/org/sonar/cxx/checks/HardcodedAccountCheck.java
@@ -60,7 +60,7 @@ public class HardcodedAccountCheck extends SquidCheck<Grammar> {
     key = "regularExpression",
     description = "literal regular expression rule",
     defaultValue = DEFAULT_REGULAR_EXPRESSION)
-  private String regularExpression = DEFAULT_REGULAR_EXPRESSION;
+  private final String regularExpression = DEFAULT_REGULAR_EXPRESSION;
 
   private Pattern pattern;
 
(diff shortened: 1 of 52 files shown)


ivangalkin:
@guwirth you've done huge effort converting all for-loops into streams with lambdas. What was the reason for that? In my opinion, the code become more complex, less readable. The new code doesn't use any of stream benefits (declarative/functional approach with maps, filters, collects, distincts, not to mention parallelism). If we want such conversion to be done for the loops, I would prefer more careful approach: rewrite only those loops, where comprehensive `stream()` operation gains some [readability] benefits. Remaining for-loops are perfectly ok to stay just for-loops IMHO .

guwirth:
@ivangalkin thanks for your feedback. In general: I  always try to keep the changes as small as possible and try to follow a pattern.
* most of your proposals can be done in a next PR
* SonarCloud issues: don't have SQ support in NetBeans, so it's try an error.
* will fix isDebugEnabled
* For my understanding stream.forEach is new and more generic way in Java 8. Readability is a matter of taste. Think after reading this several times it's as easy to understand as for-loops?

ivangalkin:
@guwirth w.r.t. to the streams: they have their own benefits and disadvantages. Some more details:

https://blog.jooq.org/2015/12/08/3-reasons-why-you-shouldnt-replace-your-for-loops-by-stream-foreach/
https://stackoverflow.com/questions/44180101/in-java-what-are-the-advantages-of-streams-over-loops

In my opinion your change introduces all the disadvantages of streams without using any possible advantages.

guwirth:
@ivangalkin I integrated most of your hints.
- removed stream()... usage
- removed most isDebugEnabled
- For more detailed samples I had no time. May be you can do it in a PR?

PR #1619, opened 2018-12-08
add 'cannot find the file...' only once

guwirth:
- reduce the noise in LOG file
- add 'cannot file the file...' only once per file in CxxIssuesReportSensor

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1619)

### DIFF
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java
@@ -142,10 +142,8 @@ public abstract class CxxIssuesReportSensor extends CxxReportSensor {
     final InputFile inputFile = sensorContext.fileSystem().inputFile(sensorContext.
       fileSystem().predicates().hasPath(path));
     if (inputFile == null) {
-      if (LOG.isDebugEnabled()) {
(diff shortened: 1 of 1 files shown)


guwirth:
@ivangalkin this PR is an improvement to the current state. Current solution add warnings  even if files are already in notFoundFiles. I moved the warning to getInputFileIfInProject. Maybe not perfect but much better.

ivangalkin:
@guwirth I agree, it's better. After some second look at the functions 

*  `getInputFileIfInProject()` and
*  `createNewIssueLocationFile()`

I believe, that the first one will produce less noise. Some background.

* function `getInputFileIfInProject()` is a service function, which might be called multiple times per project/module. It's used in all sensors implicitly (because `CxxIssuesReportSensor` is a base class) and in some of them explicitly:
    * CppcheckParserV2.java
    * CxxDrMemorySensor.java
    * CxxValgrindSensor.java

That means, that the same warning for some particular file path might appear multiple times per module/project. But it appears only once per sensor, because results of `getInputFileIfInProject()` are cached in `notFoundFiles`;

* function `createNewIssueLocationFile()` has a similar problem. It might be also called from multiple sensors + each sensor might call this function multiple time. However the warning was not associated with the cache. So it's potentially more verbose.

So it's ok from my side.

PR #1637, opened 2018-12-21
CxxPreprocessor: reduce run-time overhead, extract invariants

ivangalkin:
There were several actions, which were executed for each preprocessed source file
Among others all standard macros were re-created for each preprocessor invocation.
Create those macros statically in order to reduce the run-time overhead.

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1637)


guwirth:
@ivangalkin not sure if I wanna have it in 1.2.1. Is there a measurable difference in the performance?

ivangalkin:
@guwirth I haven't measured it (yet). It won't provide any huge boost, but is a right thing to do imho - to make invariants to the constants

guwirth:
@ivangalkin I like to merge it. I'm only not sure if we should do it in 1.2.1. Therefore the question for the risk.

ivangalkin:
@guwirth this change doesn't introduce any risk; this is refactoring only.

I am investigating memory consumption problem. I am sure, that CxxPreprocessor uselessly collects data. I'll provide one more PR soon. That one will be more invasive.

PR #1638, opened 2018-12-21
CxxPreprocessor: remove memory bloat, delete MissingIncludeFileCheck

ivangalkin:
* CxxPreprocessor was misused as a global container, which
  unconditionally collected
     * file paths to all detected include directives
     * file paths to include directives, which couldn't be resolved

* MissingIncludeFileCheck was the only user of these structures

There are several reasons to remove this functionality all together

* MissingIncludeFileCheck is highly error-prone and doesn't provide any
  valuable information (any compiler can do this job better than we)
* Global `HashMultimap`s prevent any attempts to make the analysis parallel
* They introduce run-time overhead and bloat the heap

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1638)


guwirth:
Hi @ivangalkin I’m not sure if we should remove the missing include file check. This is to give hints what’s missing in the includeDirectories. You are right that code should be always compilable and error free as a precondition. But in case you want to reduce parsing errors to create a valid AST you need includeDirectories.

ivangalkin:
@guwirth in my opinion we have to differentiate between two problems:

a) my application administrator is unable to setup SonarQube/sonar-cxx properly: we address this issue with tracing. If necessary we can trace each misconfigured include. However one missing include directory can cause thousands of warnings.  For administrator such huge number of issues might raise his/her attention.

b) my software developers create poor quality code: there are panty of useful checks which can help to avoid some hidden errors (e.g. control-flow analysis or OOP best practices, which are hard to find manually of by means of just compiling the code). Such warning are visualized in the Web-UI, they affect quality rank, they are sent via email notifications etc. Do we want, that hundreds or even thousands of "missing include issues" exposed in such way? Is it really a sign of poor software quality or just a configuration problem? How do normal developer benefits from this issues, which he/she receives daily in their mailboxes?

All false-positives decrease the trust in the analysis tools. But this particular "missing include" check is the worst one. It's incredible hard to setup sonar-cxx properly. Compilation database (JSON) or MSVC log is the only reliable source of compiler configuration. But even with the compilation database I still face 13000 include issues. So if false-positives are guaranteed, why to provide such check at all?

ivangalkin:
@guwirth 

>  please rebase it

done

I accept the argument, that some users might want to enable `MissingIncludeFileCheck` in order to visualize wrong (or missing) plugin settings. Cppcheck has similar diagnostic rules `missingInclude` and `missingIncludeSystem`. 

I pushed an alternative PR #1650, which collects missing includes, but only if corresponding check was enabled. I am ok if you prefer the alternative PR over the original one. 

Just a couple of words, why I don't like `MissingIncludeFileCheck`. Speaking e.g. about cppcheck: it's a command line tool. All results are piped to STDOUT, STDERR or maybe to the file. In comparison to this tool we have different stakeholders by design. SonarQube has the concept of scanners, application server and the web-UI. So there are

a) developer(s), which maintain the scanners (or rather scanners' jobs),
b) administrator(s) or the application server and
c) just usual developers, which are interested in the defects

The  `MissingIncludeFileCheck` addresses the target group a), but uses the Web-UI / email notification. Those are the communication channel of the group c). So for the usual developers the missing-include notifications are false-positives. They shatter confidence in SonarQube. Of course `MissingIncludeFileCheck` can be disabled, but again the "normal" developers do not have power to change any settings. So IMHO it's better not to have this check at all.

We already have a trace. That's the communication channel for the group a).

guwirth:
@ivangalkin what are we doing with cxx:ParsingErrorRecovery (e.g. #1648)? Should also be in LOG file only if we follow this approach?

PR #1645, opened 2018-12-27
CxxPreprocessor: fix handling of SquidAstVisitorContext

ivangalkin:
* `SquidAstVisitorContext::getFile()` never returns `null` during productive execution
* the only exception is preprocessing inside of preprocessor's constructor
  (error-prone design in my humble opinion)

* Fixes:
   * wrong assumptions in `CxxPreprocessor` were fixed
   * documentation was added
   * poorly parametrized mocks were fixed in the unit tests
   * minor refactoring

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1645)


guwirth:
@ivangalkin looks good. And yes, maybe there is a better solution how to add default macros and config settings to preprocessor.

Issue #1415, opened 2018-02-01
Performance: Slow analysis of arrays


ryosa:
With SQ 5.6.7 and sonar-cxx 0.9.8, analysis of  multi dimensional arrays is very slow. A 22-hour run (as a Jenkins job on CentOS7) didn't analyze a file which contains 512 2D arrays, the file size is near 5MB.
It seems that the analysis uses only one CPU. This may be a matter of Jenkins, but I wonder if the performance would be improved.
Thanks.

guwirth:
Hi @ryosa can you provide a small sample please to reproduce the issue. Regards

ryosa:
@guwirth  Attached should take a long time to analyze. Please change the file name. Thanks.
[dim_test.cpp.txt](https://github.com/SonarOpenCommunity/sonar-cxx/files/1693789/dim_test.cpp.txt)

guwirth:
@ryosa trying this in the SSLR toolkit it needs about 5s on my computer to parse it and create an AST.

> It seems that the analysis uses only one CPU

Normally compiler are compiling one cpp file always with one core only. To reduce total time we can think about compiling more than one cpp file in parallel but that would help nothing here. Think we can have a look why total time is 5s.

ryosa:
@guwirth Thank you for your try and comment. If analysis of arrays should be improved, that will be a big advantage for sonar-cxx. I expect it will be done. :-)

guwirth:
@ryosa thinks I'm wondering:

* See no reason why there should be differences between one and multi dimensional arrays? Do you really have problems with multi dimensional arrays only?
* Maybe it's a problem with hex numbers? Do you have the same problem with decimal numbers?
* Maybe it's a problem with parsing (the grammar)?
* Maybe it's a problem with AST creation (AST too big)? Memory problem?

Can you give some more hints?

ryosa:
@guwirth  Analysis is slow with both 1D and 2D arrays, and the analyzing time seems to be in proportion to the size of array. It is not a matter of hex numbers because analysis is also slow with 1D floating point number arrays. I'm not sure about  decimal number array, as the project contains only hex and float arrays. The server has been equipped with 128GB memory. The shell script below runs prior to each analysis. 

> export SONAR_SCANNER_OPTS="-Xmx4096m"
export SONAR_RUNNER_OPTS="-Xmx1G -XX:MaxPermSize=512m"
ulimit -s 409600

[...]

PR #1646, opened 2018-12-29
improve number lexer

guwirth:
- reduce the number of regex rules to detect a number
- use only one channel to detect numbers
- improve #1415

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1646)


ivangalkin:
@guwirth Your change looks good to (although I am not sure, that I understand it in detail). I am not available for reviews until the the 2nd week of January.

I wish you happy new year! My best wishes to all contributors too!

guwirth:
@ivangalkin 
> Your change is valid. However the original regexp is not really clear to me. Does it make sense to have * followed by +? Wouldn’t + be sufficient?

Good explanation is here: https://docs.oracle.com/javase/tutorial/essential/regex/quant.html.
See also in the source code https://github.com/SonarSource/sslr/blob/master/sslr-core/src/main/java/com/sonar/sslr/impl/channel/RegexpChannelBuilder.java.

```Java
  public static String opt(String regexpPiece) {
    return regexpPiece + "?+";
  }

  public static String one2n(String regexpPiece) {
    return regexpPiece + "++";
  }

 public static String o2n(String regexpPiece) {
    return regexpPiece + "*+";
}
```

PR #1649, opened 2019-01-01
use Maven 3.5.4

guwirth:
- set 4 digit version in script (old version was renaming JAR only)
- update sonar-maven-plugin
- update maven-surefire-plugin
- update maven-compiler-plugin

This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1649)


guwirth:
@ivangalkin it’s Python glob: 
https://docs.python.org/3/library/glob.html

Issue #1326, opened 2017-11-20
Refactoring: cleanup CxxSquidSensor - remove getJsonCompilationDatabaseFile


guwirth:
Remove `getJsonCompilationDatabaseFile` call from `CxxSquidSensor`:  `cxxConf.getJsonCompilationDatabaseFile` in `createConfiguration`.

This is slowing down the plugin because `CxxSquidSensor` is called for each source file. `getJsonCompilationDatabaseFile`  should be called only once and save parsed setting at a central place e.g. `CxxConfiguration`.

see also #1325

Issue #1325, opened 2017-11-20
Refactoring: cleanup CxxSquidSensor - remove CxxVCppBuildLogParser


guwirth:
Remove `CxxVCppBuildLogParser` call from `CxxSquidSensor`:  `cxxConf.setCompilationPropertiesWithBuildLog` in `createConfiguration`.

This is slowing down the plugin because `CxxSquidSensor` is called for each source file. `CxxVCppBuildLogParser`  should be called only once and save parsed setting at a central place e.g. `CxxConfiguration`.

Bertk:
I checked the behavior and the log file parser is called once per module (per VC++ project) which is correct. 
```
2017-12-18T07:12:00.8478356Z INFO: build log parser baseDir='C:\agent\_work\1\s\Source\SQCxxTestData'
2017-12-18T07:12:00.9572044Z INFO: Parse build log 'C:\agent\_work\1\s\CppTesting.sln.log' added includes: '16', added defines: '16'
2017-12-18T07:12:04.3476589Z INFO: build log parser baseDir='C:\agent\_work\1\s\Source\MotorController'
2017-12-18T07:12:04.4570271Z INFO: Parse build log 'C:\agent\_work\1\s\CppTesting.sln.log' added includes: '16', added defines: '16'
2017-12-18T07:12:23.2740779Z INFO: build log parser baseDir='C:\agent\_work\1\s\Source\RootFinder'
2017-12-18T07:12:23.3834404Z INFO: Parse build log 'C:\agent\_work\1\s\CppTesting.sln.log' added includes: '16', added defines: '16'
2017-12-18T07:12:42.4299502Z INFO: build log parser baseDir='C:\agent\_work\1\s\Source\Calculation'
2017-12-18T07:12:42.5393214Z INFO: Parse build log 'C:\agent\_work\1\s\CppTesting.sln.log' added includes: '16', added defines: '16'
```
The parser will analyze each line of the log file which has a performance impact for large multi module log files and may also create wrong settings because the parser does not handle the module settings separately e.g. it consumes all include paths or preprocessor definitions.

sample log file parser results
```
12:24:22.091 DEBUG org.sonar.cxx.CxxVCppBuildLogParser - build log parser baseDir='.'
[...]

Issue #1215, opened 2017-08-24
JSON Compilation Database class does not handle "-isystem" include path references


apriori:
Title should be self-explanatory

guwirth:
@vesajaaskelainen do you have some time to have a look to this?

guwirth:
Not sure if adding system include paths is a good idea. This will slow down the parser and is typically not needed. Better and much faster is typically to provide the macros creating syntax errors with defines or force include (see https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Dealing-with-compiler-specific-code-pieces). 

@Bertk @jmecosta how are you doing this with your VS analysis?

jmecosta:
we have system headers, both in server and local analysis the time spend on
analysis is not that significant for us so that we could consider disabling
those headers.

vesajaaskelainen:
Compiler's headers are more or less mandatory in order to get good level of analysis. Even thou the analysis is rather slow it is better to analyze correctly than incorrectly. Aka. I would sacrifice speed over accuracy. 

Making it configurable for user I suppose is OKish.

vesajaaskelainen:
Link to -isystem documentation in case someone was wondering:
https://gcc.gnu.org/onlinedocs/gcc-7.2.0/gcc/Directory-Options.html#Directory-Options

I suppose -isystem would need more in-depth support to handle different classes of headers.

Minimal compatible version probably would be to treat -isystem <path> as -i <path> but it is slightly incorrect.

guwirth:
Hi @tisoft, you seems to be familiar with JSON Compilation Database. Do we need the -isystem support?

tisoft:
I have a local fork with isystem enabled and it helps. Still testing though.

guwirth:
@tisoft If that works, it would be nice if you could provide the solution as PR.

PR #1665, opened 2019-01-11
CxxPreprocessor: cache forced includes, forced & global defines

ivangalkin:
Also:

apply correct order macro definitions
    
1. standard macro values (e.g.` __LINE__`)
2. configured defines (e.g. `-D__LINE__=123`)
3. forced include files (e.g. header contains `#define __LINE__ 456`)

-> 2 overrides 1
-> 3 overrides 2 and 1

This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1665)

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
@@ -90,28 +90,39 @@ public class CxxPreprocessor extends Preprocessor {
 
   private final Parser<Grammar> pplineParser;
   /**
-   * Contains pre-parsed forced defines (see
-   * {@link CxxConfiguration#getDefines()}), {@link Macro#STANDARD_MACROS} and
-   * forced includes (see {@link CxxConfiguration#getForceIncludeFiles()}). Used
-   * if there is no compilation unit settings. All hi-prio macros are pre-parsed
-   * while construction of {@link CxxPreprocessor}
+   * Contains a standard set of pre-compiled macros, which are defined for each
+   * compilation unit. This map is used if there is no specific compilation unit
+   * settings. The map contains
+   * <ol>
+   * <li>{@link Macro#STANDARD_MACROS}</li>
+   * <li>{@link CxxConfiguration#getDefines()} and</li>
+   * <li>forced includes (see
+   * {@link CxxConfiguration#getForceIncludeFiles()})</li>
+   * </ol>
+   * All hi-prio macros are pre-parsed while construction of
+   * {@link CxxPreprocessor}
    */
   private final MapChain<String, Macro> fixedMacros = new MapChain<>();
   /**
    * If CxxConfiguration contains any compilation unit settings, this map is
-   * filled with pre-parsed forced defines, which must be applied for each
-   * compilation unit: {@link CxxConfiguration#getDefines()}),
-   * {@link Macro#UNIT_MACROS} and forced includes (see
-   * {@link CxxConfiguration#getForceIncludeFiles()}); Immutable; macros are
-   * pre-parsed while construction of {@link CxxPreprocessor}
+   * filled with a set of pre-compiled macros. Those macros must be defined for
+   * each compilation unit:
+   * <ol>
+   * <li>{@link Macro#UNIT_MACROS}</li>
+   * <li>{@link CxxConfiguration#getDefines()} and</li>
+   * <li>forced includes (see
+   * {@link CxxConfiguration#getForceIncludeFiles()})</li>
+   * </ol>
+   * The map is immutable; macros are pre-parsed while construction of
+   * {@link CxxPreprocessor}
    */
-  private final Map<String, Macro> forcedUnitMacros;
+  private final Map<String, Macro> predefinedUnitMacros;
   /**
    * If current processed file has some specific configuration settings, this
    * map will be filled with relevant macros and defines:
    * <ol>
-   * <li>pre-parsed forced macros will be added (see
-   * {@link CxxPreprocessor#forcedUnitMacros}</li>
+   * <li>predefined compilation unit macros will be added (see
+   * {@link CxxPreprocessor#predefinedUnitMacros}</li>
    * <li>specific unit settings will be parsed and added (see
    * {@link CxxConfiguration#getCompilationUnitSettings(String)}</li>
    * </ol>
@@ -162,11 +173,11 @@ public class CxxPreprocessor extends Preprocessor {
 
     pplineParser = CppParser.create(conf);
 
-    final List<String> forcedDefines = conf.getDefines();
-    Map<String, Macro> forcedMacros = Collections.emptyMap();
-    if (!forcedDefines.isEmpty()) {
-      LOG.debug("parsing forced defines");
-      forcedMacros = parseMacroDefinitions(forcedDefines);
+    final List<String> configuredDefines = conf.getDefines();
+    Map<String, Macro> configuredMacros = Collections.emptyMap();
+    if (!configuredDefines.isEmpty()) {
+      LOG.debug("parsing configured defines");
+      configuredMacros = parseMacroDefinitions(configuredDefines);
     }
 
     // fill CxxPreprocessor.fixedMacros
@@ -176,14 +187,14 @@ public class CxxPreprocessor extends Preprocessor {
     }
     try {
       getMacros().setHighPrio(true);
-      getMacros().putAll(forcedMacros);
       getMacros().putAll(Macro.STANDARD_MACROS);
+      getMacros().putAll(configuredMacros);
       parseForcedIncludes();
     } finally {
       getMacros().setHighPrio(false);
     }
 
-    // fill CxxPreprocessor.forcedUnitMacros() if relevant
+    // fill CxxPreprocessor.predefinedUnitMacros if relevant
     final CxxCompilationUnitSettings globalCUSettings = conf.getGlobalCompilationUnitSettings();
     if (!conf.getCompilationUnitSourceFiles().isEmpty() || (globalCUSettings != null)) {
       unitMacros = new MapChain<>();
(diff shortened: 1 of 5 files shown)


ivangalkin:
@guwirth I found out, that you marked `sonar.cxx.jsonCompilationDatabase` as deprecated (see https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Supported-configuration-properties/_compare/2a3859fbbe6e030b32a06245b61b0764310b8807...eef90bd839398469d58d4495de2fc803d90e051b). I strongly disagree with the deprecation. At the moment JSON compilation database it the only

* convenient and
* compiler independent way

to add the includes and defines.

SonarCFamily also uses some similar but proprietary solution: intercept the build and store the compiler arguments into the JSON database.

guwirth:
@ivangalkin maybe that is a misunderstanding. 

* Json compilation database support sonar.cxx.jsonCompilationDatabase is also experimental only
sonar.cxx.scanOnlySpecifiedSources is no more supported. There were conflicts with sonar.sources and sonar.tests.

The problem I see is that we have too many ways to handle macros and includes. Having different input formats is ok for me but internally we should handle it in an unique way. That was the reason for #1365, #1325 and #1326. 

There seems to be also some open questions like #1215. 

I had also less feedback that someone is using it. That’s the reason why the functionality is marked experimental.

ivangalkin:
@guwirth right now the page "[Supported configuration properties](https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Supported-configuration-properties#sonarcxxjsoncompilationdatabase)" looks as following:

```
sonar.cxx.jsonCompilationDatabase

Scope: project
Default: ---
History: no more supported with v1.0.0+

Enable Sonar C++ analysis to utilize JSON compilation database support. Specifies file to be used as JSON compilation database. JSON compilation database is used to improve C++ symbol and include directory knowledge. Sonar C++ plugin also supports extension to allow importing of used compiler's internal symbols and includes.

For more information please see: JSON compilation database format specification support

Example:
sonar.cxx.scanOnlySpecifiedSources

Scope: project
Default: False

When using JSON compilation database support limit Sonar C++ plugin analysis to specified files in JSON compilation database. This setting can be used to limit analysis to only compiled files.

Example:
```
According to you comment, the deprecation note  "History: no more supported with v1.0.0+" should be moved from `sonar.cxx.jsonCompilationDatabase` to `sonar.cxx.scanOnlySpecifiedSources`?

cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
```diff
@@ -127,34 +162,56 @@ public CxxPreprocessor(SquidAstVisitorContext<Grammar> context,
 
     pplineParser = CppParser.create(conf);
 
+    final List<String> forcedDefines = conf.getDefines();
+    Map<String, Macro> forcedMacros = Collections.emptyMap();
```
guwirth: don't like force

cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
```diff
@@ -127,34 +162,56 @@ public CxxPreprocessor(SquidAstVisitorContext<Grammar> context,
...
+    if (!forcedDefines.isEmpty()) {
+      LOG.debug("parsing forced defines");
```
guwirth: ... it's not really forced. These are macros from configuration?

cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
```diff
@@ -90,8 +89,44 @@
...
   private final MapChain<String, Macro> fixedMacros = new MapChain<>();
+  /**
+   * If CxxConfiguration contains any compilation unit settings, this map is
+   * filled with pre-parsed forced defines, which must be applied for each
+   * compilation unit: {@link CxxConfiguration#getDefines()}),
+   * {@link Macro#UNIT_MACROS} and forced includes (see
+   * {@link CxxConfiguration#getForceIncludeFiles()}); Immutable; macros are
+   * pre-parsed while construction of {@link CxxPreprocessor}
+   */
+  private final Map<String, Macro> forcedUnitMacros;
```
guwirth: Why call it forced?

cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
```diff
@@ -127,34 +162,56 @@ public CxxPreprocessor(SquidAstVisitorContext<Grammar> context,
...
-        LOG.debug("parsing external macro: '{}'", define);
-        if (!"".equals(define)) {
-          Macro macro = parseMacroDefinition("#define " + define);
-          LOG.debug("storing external macro: '{}'", macro);
-          getMacros().put(macro.name, macro);
-        }
-      }
-
-      // set standard macros
+      getMacros().putAll(forcedMacros);
```
guwirth: Is this the right order? First standard macros and then forced? Think should be possible to overwrite standard macros with forced.

cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
```diff
@@ -127,34 +162,56 @@ public CxxPreprocessor(SquidAstVisitorContext<Grammar> context,
...
+    // fill CxxPreprocessor.forcedUnitMacros() if relevant
+    final CxxCompilationUnitSettings globalCUSettings = conf.getGlobalCompilationUnitSettings();
+    if (!conf.getCompilationUnitSourceFiles().isEmpty() || (globalCUSettings != null)) {
+      unitMacros = new MapChain<>();
+      if (getMacros() != unitMacros) {
+        throw new IllegalStateException("expected unitMacros as active macros map");
+      }
+      try {
+        getMacros().setHighPrio(true);
+        getMacros().putAll(forcedMacros);
```
guwirth: Same here, also not sure about the order.

cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
```diff
@@ -478,54 +535,32 @@ public void init() {
...
         }
       } else {
         // Use global settings
         LOG.debug("global settings for: '{}'", currentContextFile);
-        if (isCFile(currentContextFile.getAbsolutePath())) {
+        if (isCFile(currentContextFile.getName())) {
           //Create macros to replace C++ keywords when parsing C files
           getMacros().putAll(Macro.COMPATIBILITY_MACROS);
-          fixedMacros.disable(CPLUSPLUS);
+          getMacros().disable(Macro.CPLUSPLUS);
```
guwirth: think there are more macros different between C and C++?

cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
```diff
@@ -127,34 +162,56 @@ public CxxPreprocessor(SquidAstVisitorContext<Grammar> context,
...
-
-      // parse the configured force includes and store into the macro library
-      for (String include : conf.getForceIncludeFiles()) {
-        LOG.debug("parsing force include: '{}'", include);
-        if (!"".equals(include)) {
-          parseIncludeLine("#include \"" + include + "\"", "sonar." + this.language.getPropertiesKey()
-            + ".forceIncludes", conf.getEncoding());
-        }
-      }
+      parseForcedIncludes();
```
guwirth: Order also not clear for me?
* predefined
* configuration
* forceInclude

If yes we should document it in wiki.

cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
ivangalkin: the map `Macro.STANDARD_MACROS_IMPL` contains only two C++ specific macros: 

* __cplusplus
* __has_include (since C++17)

so theoretically we have to disable `__has_include` to. BUT

1. I tested GCC in the ANSI-C mode and it still provides this define (https://coliru.stacked-crooked.com/a/30a53f2c2a9a323d)
2. Our implementation of `__has_include` is hard-wired in the grammar and the expansion of `__has_include()` is not restricted to C++. So even if `#ifdef __has_include` evaluates to `0`, we still can't prevent `__has_include(...)` from evaluation.
3. This C vs C++ was not scope of my change, so please let's consider to change it in other PR

cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
```diff
@@ -127,34 +173,56 @@ public CxxPreprocessor(SquidAstVisitorContext<Grammar> context,
...
+        throw new IllegalStateException("expected unitMacros as active macros map");
+      }
+      try {
+        getMacros().setHighPrio(true);
+        getMacros().putAll(Macro.UNIT_MACROS);
+        getMacros().putAll(configuredMacros);
+        parseForcedIncludes();
+        predefinedUnitMacros = new HashMap<>(unitMacros.getHighPrioMap());
+      } finally {
+        unitMacros = null;
```
guwirth: @ivangalkin is here `getMacros().setHighPrio(false)` missing?

cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
```diff
@@ -127,34 +173,56 @@ public CxxPreprocessor(SquidAstVisitorContext<Grammar> context,
 
     pplineParser = CppParser.create(conf);
 
+    final List<String> configuredDefines = conf.getDefines();
```
guwirth: @ivangalkin ctor code is difficult to read. Maybe you can split it into smaller methods.

cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
ivangalkin: no, in this case we don't have to disable the high-prio mode

1. unitMacros = new MapChain<>(); -> make getMacro() return unitMacros
2. fill the map getMacro() directly and (more important) indirectly through parseForcedIncludes()
3. all precompiled macros for compile units are now in unitMacros.getHighPrioMap(); we copy them into predefinedUnitMacros
4. unitMacros = null; -> destroy the unitMacros; they won't be filled from constructor anymore, no need to toggle the high-prio mode

cxx-squid/src/main/java/org/sonar/cxx/preprocessor/CxxPreprocessor.java
ivangalkin: yes, I'll do some refactoring; I'll remove the 'try{...}finally {}' code from the ctor. It doesn make much sense here. Exceptions are not handled -> `operator new` doesn't create any object instance -> there will be no sense in house-keeping. The case in CxxPreprocessor::init() is different: there is an object, it is reused for all the processed files, the object must be kept in consistent state.

PR #1667, opened 2019-01-19
optimize AST

guwirth:
- skip some nodes to make resulting AST more readable and improve performance
- improve grammar performance

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1667)

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/parser/CxxGrammarImpl.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/parser/CxxGrammarImpl.java
@@ -1118,7 +1118,7 @@ public enum CxxGrammarImpl implements GrammarRuleKey {
             b.sequence(b.optional(attributeSpecifierSeq), b.optional(nestedNameSpecifier), IDENTIFIER) // C++
           )
         ),
-        b.sequence(b.optional(cliAttributes), CxxKeyword.ENUM, b.optional(nestedNameSpecifier), IDENTIFIER) // C++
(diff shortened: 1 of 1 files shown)


ivangalkin:
@guwirth I believe there is an error in CxxGrammarImpl->elaboratedTypeSpecifier. The remaining refactoring looks good to me.

guwirth:
@ivangalkin thx will have a look.

cxx-checks/src/test/resources/checks/UsingNamespaceInHeader.h
```diff
@@ -8,7 +8,7 @@ namespace Foo
     void fooFunc( );
 }
 
-using namespace std;
+using namespace std; // error
```
ivangalkin: just for your information: this error is equivalent to clang-tidy's rule `google-global-names-in-headers` aka "using declarations in the global namespace in headers are prohibited"

cxx-checks/src/test/resources/checks/UsingNamespaceInHeader.h
```diff
@@ -30,3 +30,13 @@ namespace Foo
         privateFunction( );
     }
 }
+
+void fooFunc()
```
ivangalkin: clang-tidy causes warning here and below too. Both have rule IDs `google-build-using-namespace` aka  "do not use namespace using-directives; use using-declarations instead". So that's some separate warning, which is not specific to the headers. Therefore I think your fix works as expected. Thank you.

PR #1668, opened 2019-01-20
Improves Visual studio/MSBUILD log file analysis

rudolfgrauberger:
Uses regular expression to get the project file indenpendently of the language of the log file. I've tested this with the existing englisch log file and with a new german log file.

This is important, among other things, the following points:
- Special folder structure
- Simplifies configuration

### 1. Special folder structure
The detecting of correct project file is nessesary if you have following project folder structure

    .
    ├── Build                                # Project files (.vxcproj)
    │   ├── project.vcxproj
    │   ├── buildlog.log                     # MSBUILD log file
    ├── Interface                            # Public Header files
    ├── Source                               # Source files (and internal header files)
    ├── Test                                 # Tests
    └── sonar-project.properties

The relative include paths are then correctly converted relative to the project file.

### 2. Simplifies configuration
It is very helpful because all INCLUDE and DEFINES are taken from the project file and do not have to be defined manually.

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1668)


guwirth:
Hi @rudolfgrauberger, tests on Travis (Linux) are failing:
```
[ERROR] Errors: 
[ERROR]   CxxVCppBuildLogParserTest.shouldExtractProjectFileFromEnglishVCpp2013Log:73 NullPointer
[ERROR]   CxxVCppBuildLogParserTest.shouldExtractProjectFileFromGermanVCpp2017Log:105 NullPointer
```

rudolfgrauberger:
@guwirth Oh, okay, thanks for the info. I will correct it in the next few days.

rudolfgrauberger:
@guwirth Ready :-)

ivangalkin:
@rudolfgrauberger Thank you for the providing of this PR. I have no strong opinion about the idea of your change. My review addresses the code style only.

rudolfgrauberger:
@ivangalkin Thank you very much for your feedback/review. I am currently in the exam phase and will answer/work on the points immediately afterwards.

ivangalkin:
@rudolfgrauberger there is no hurry, good luck with your exams

guwirth:
@rudolfgrauberger 

>> to get the project file indenpendently of the language of the log file. 

Solutions makes it more generic. Think in the past it was no issue because most developer are using an Englisch VS.

>> Could we use sonar-cxx/cxx-sensors/src/test/resources/org/sonar/cxx/sensors/reports-project/compiler-reports/VC-report.vclog instead? could we extend the existing log instead of adding a new one?

Think it's ok to use two files German/English. But would also shorten it a little bit to the relevant parts.

rudolfgrauberger:
> Ok.
> 
> 1. The file should be as large as necessary to cover the tested functionality. So some dummy project with (for example) 3 compile units (CPP files), the different types of includes (""-include, <>-include etc) and the different types of defines (-Dstuff, -Dstuff=stuff, -Dstuff(x,y)=(x)+(y)) etc. would be more than enough IMHO.
> 2. If you want to validate that your parsing is language-independent you'll need the compiler logs of the same project with different language settings. So basically you use your first test (see above) as a reference. The second test parses the German log. Afterwards it asserts that the set of source files, the set of includes and the set of defines are equal to the corresponding sets of the English log.

That sounds good. I'll remove the big log file (vc++2017-de.txt) and create a new and smaller project file for the log file. In this project I will set the different cases for include specifications (relative and absolute paths, with and without quotes) and definitions.

Then I will create a German and an English log file from this project and reduce it to the necessary size.

@ivangalkin and @guwirth  Is that all right? Should I do all that in this branch and then just make a new commit?

> Solutions makes it more generic. Think in the past it was no issue because most developer are using an Englisch VS.

Could be or most of them didn't use this feature at all and have manually specified all definitions and includes in the properties file (at least I've seen this in some projects before), but that doesn't really matter. :-)

rudolfgrauberger:
@guwirth Can you restart the build on travis-ci manually, please? The one error is not my fault...

ivangalkin:
@rudolfgrauberger I like the tests much more! Thank you. There are still some little thing to improve, but my comments are minor.

ivangalkin:
@guwirth @rudolfgrauberger BTW there were some important discussion in #1678
As far as I get it @rudolfgrauberger suggested to improve the regular expressions.

Shouldn't we fix them?

rudolfgrauberger:
```powershell
git clone -q https://github.com/SonarOpenCommunity/sonar-cxx.git C:\projects\sonar-cxx
fatal: unable to access 'https://github.com/SonarOpenCommunity/sonar-cxx.git/': Could not resolve host: github.com
```
@guwirth, @ivangalkin Also this error is not related to my changes. Can you please repeat/restart the build?

rudolfgrauberger:
> @guwirth @rudolfgrauberger BTW there were some important discussion in #1678
> As far as I get it @rudolfgrauberger suggested to improve the regular expressions.
> 
> Shouldn't we fix them?

Should I create a separate issue for this discussion?

guwirth:
@rudolfgrauberger I merge it. Let’s see if problem with AppVeyor is still there.

Issue #1636, opened 2018-12-20
cppcheck missing "nullPointerArithmeticRedundantCheck" rule


VirtualTim:
### Description
Sonarqube cppcheck is missing the "nullPointerArithmeticRedundantCheck" rule. I believe it should be under https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/resources/cppcheck.xml

The following is logged (filenames removed):

```
ERROR: Could not add the issue 'CxxReportIssue [ruleRepoKey=cppcheck, ruleId=nullPointerArithmeticRedundantCheck, locations=CxxReportLocation [file=C:████/████.cpp, line=448, info=Either the condition '!████' is redundant or there is pointer arithmetic with NULL pointer.], CxxReportLocation [file=C:████/████.cpp, line=448, info=Null pointer addition], CxxReportLocation [file=C:████/████.cpp, line=450, info=Assuming that condition '!████' is not redundant]]':The rule 'cppcheck:nullPointerArithmeticRedundantCheck' does not exist.
', skipping issue
```

If it helps here is an exert from the cppcheck xml report:
```
<results version="2">
<cppcheck version="1.84"/>
  <errors>
    <error id="nullPointerArithmeticRedundantCheck" severity="warning" msg="Either the condition '!████' is redundant or there is pointer arithmetic with NULL pointer." verbose="Either the condition '!████' is redundant or there is pointer arithmetic with NULL pointer." cwe="682">
      <location file0="C:████/████.cpp" file="C:████/████.cpp" line="448" info="Null pointer addition"/>
      <location file0="C:████/████.cpp" file="C:████/████.cpp" line="450" info="Assuming that condition '!████' is not redundant"/>
    </error>
  </errors>
</results>
```

guwirth:
Hi @VirtualTim, thx for your feedback.

ivangalkin:
Hi @VirtualTim & @guwirth,

cppcheck has a CLI switch, which allows to list all accessible rules. Unfortunately from time to time they miss one or another ID there. In our rules repository we use only officially published rules, which I believe makes sense.

I've (kind of) reported the missing rule by cppcheck. Please see the thread https://github.com/danmar/cppcheck/pull/1292.

Regards,

ivangalkin:
Btw. I changed the subject, because it's not only the version 1.84 of cppcheck.
[...]

PR #1672, opened 2019-01-23
Added cppcheck rule keys for "iterator2" and "internalAstError". These were sh…

gjd6640:
…owing up in a cppcheck report and the plugin was printing a warning to the console about errors being present in the report file that this plugin didn't recognize.

### DIFF
--- a/cxx-sensors/src/main/resources/cppcheck.xml
+++ b/cxx-sensors/src/main/resources/cppcheck.xml
@@ -2582,6 +2582,26 @@ Same iterator is used with two different containers.
     <remediationFunctionGapMultiplier>5min</remediationFunctionGapMultiplier>
   </rule>
   <rule>
+    <key>iterators2</key>
+    <name>Same iterator is used with two different containers</name>
+    <description>
+      <![CDATA[
+<p>
+Same iterator is used with two different containers.
+</p>
+<h2>References</h2>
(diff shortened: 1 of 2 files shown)


guwirth:
Hi @gjd6640,

thanks for providing this patch. See #1636 for a description of the problem. Think right way would be to ask Cppcheck to fix the issue.

Regards,

guwirth:
@gjd6640 you can download and test it from here: [Download latest snapshot](https://ci.appveyor.com/project/SonarOpenCommunity/sonar-cxx/branch/master/artifacts)

ivangalkin:
@gjd6640 some short explanation from my side: we import only the ruleIDs which are exported by cppcheck officially. For example w.r.t. the `internalAstError` there should be the decision of cppcheck team first, whether they will export this ID or won't (https://github.com/danmar/cppcheck/pull/1576). If there is some public ruleID missing, please report it by cppcheck directly.

@guwirth the missing rule IDs are fixed in cppcheck (I can speak for sure for `iterator2`, `internalAstError` was under discussion). After cppcheck 1.87 will be officially released, we'll import the new IDs to our rules repository. They are not updated in the snapshot yet.

gjd6640:
Sounds good. I'm fine with waiting for the next release to pick up these
two rules.

PR #1674, opened 2019-01-25
test with SQ 7.6

guwirth:


guwirth:
Assertion Failed:  directories is actually 5 [expected: 4.0]
```
@SqApi75
  Scenario: Smoketest 
    Given the project "smoketest_project"
    And the following metrics have following values
      | directories              | 4     |
      Assertion Failed: 
      	directories is actually 5 [expected: 4.0]
```

PR #1675, opened 2019-01-26
fix #1671

### LINKED ISSUE #1671: Scans fail with java.nio.charset.MalformedInputException in ReservedNamesCheck.visitFile
### Description

Scan fails with error:  ERROR: Error during SonarQube Scanner execution
org.sonar.squidbridge.api.AnalysisException: Unable to analyze file: /my/company/somefile.H

This error comes from here: org.sonar.cxx.checks.**ReservedNamesCheck**.visitFile(ReservedNamesCheck.java:64)

See the attached text file for the stack trace that printed when I used the "-X" option.

Character set the scanner expects is UTF-8. It appears that my file actually conforms to the "CP 1252" character set and the scan chokes on the "left quote" (0x223) and "right quote" (0x224) characters.

### Steps to reproduce the problem

Please provide the steps required to reproduce the problem

1. Create a text file that contains the 0x223 and 0x224 characters
2. Scan that text file with the sonarqube scanner and using the default UTF-8 encoding
3. Observe that the scan exits with the error message given at the top of this issue.

If possible then please create a git repository with a repro sample or attach a zip to the issue.

### Expected behavior

This rule should just skip past any non-translatable characters instead of exiting. There's another rule to raise an issue when bytes that don't map to characters.

### Actual behavior

Scan stops / fails.

### Known workarounds

I specified the correct character set by setting the options `-Dsonar.sourceEncoding=CP1252 -Dsonar.cxx.gcc.charset=CP1252`. I'm still seeing the warning text below BUT the error message has gone away and the scan does complete. Probably that WARN output is written by some other check class.

> "WARN: Invalid character encountered in file /myprojectpath/MySourceFile.C at line 7 for encoding UTF-8. Please fix file content or configure the encoding to be used using property 'sonar.sourceEncoding'."

### Related information

This looks similar to issue #371 but isn't the exact same issue.

* sonar-cxx-plugin-1.2.2-SNAPSHOT (source cloned from master around 4:30pm on 1/22/2019)
* SonarQube version:  Version 6.7.3 (build 38370)

guwirth:
- support malformed input and unmappable character in checks
- close #1671

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1675)

### DIFF
--- a/cxx-checks/src/main/java/org/sonar/cxx/checks/ReservedNamesCheck.java
+++ b/cxx-checks/src/main/java/org/sonar/cxx/checks/ReservedNamesCheck.java
@@ -21,11 +21,12 @@ package org.sonar.cxx.checks;
 
 import com.sonar.sslr.api.AstNode;
 import com.sonar.sslr.api.Grammar;
+import java.io.BufferedReader;
+import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
-import java.util.List;
 import java.util.Locale;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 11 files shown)


ivangalkin:
@guwirth basically you should replace all `java.nio.file.File.readAllBytes()` with `java.nio.file.File.readAllLines()`. This function is implemented through the `BufferedReader` and would solve the problem too.

In some cases it's more efficient to apply the checks line-by-line. So your change might even decrease the run-time / the memory footprint.

guwirth:
@ivangalkin tried 
`final BufferedReader br = Files.newBufferedReader(getContext().getFile().toPath(), charset);`
not working. Seems
`onMalformedInput(CodingErrorAction.REPLACE) / onUnmappableCharacter(CodingErrorAction.REPLACE)`
missing

PR #1688, opened 2019-02-12
use real path in coverage sensors also

jmecosta:
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1688)

### DIFF
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/CxxCoverageSensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/coverage/CxxCoverageSensor.java
@@ -23,9 +23,11 @@ import java.io.File;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import javax.xml.stream.XMLStreamException;
 import org.sonar.api.batch.fs.InputFile;
 import org.sonar.api.batch.sensor.SensorContext;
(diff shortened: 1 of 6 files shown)


jmecosta:
@ivangalkin inst each sensor a own instance during the analysis stage. Even in a multi module scenario, each sensor is instantiated again and again? so the notFoundFiles is always empty when the sensor executes?

If thats the case, maybe notFoundFiles could be injected in similar way we handle the coverage cache data. It makes sense since notFoundFiles should be constant for all sensor, so it should speed up the analysis?

ivangalkin:
@jmecosta before #1522, all sensors were derived from `CxxReportSensor` and **all** of them had a copy of `notFoundFiles`. Unfortunately the logic of the methods `CxxReportSensor::execute()`, `CxxReportSensor::executeReport()` etc. was absolutely not applicable to the coverage sensors (among others). So I had to create a new class `CxxIssuesReportSensor` (`CxxIssuesReportSensor extends CxxReportSensor`) and move a large block of code to this new file. But if parts of logic became useful for all sensors again, why not to move these logic back? I prefer consistency over optimization so I personally don't see any problem.

> If thats the case, maybe notFoundFiles could be injected in similar way we handle the coverage cache data.

This could bring some optimization indeed. But again 1) the handling of `notFoundFiles` should be available to all sensors (and therefore moved back to `CxxReportSensor`) 2) we'll introduced some shared state between sensors. Right now such shared state doesn't play any role (all sensors are executed successively), but it could hinder the future parallelization.

jmecosta:
So, for now just move the thing back to CxxReportSensor and leave optimization to some later stage?

ivangalkin:
@jmecosta that would be the right solution in my opinion;

guwirth:
@jmecosta how should we continue here? Do you make another PR with the proposal of @ivangalkin?

guwirth:
replaced by #1701

PR #1690, opened 2019-02-15
optimize Travis

guwirth:
- download only needed SQ ZIP
- remove travis_retry: makes no sense -q suppress all output and default timeout setting of wget is 900s (>10min)

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1690)


guwirth:
@ivangalkin yes you are right...

PR #1710, opened 2019-04-18
Add path/flow support for Clang Static Analyzer *.plist reports

### LINKED ISSUE #1707: Clang Static Analyzer: add support for full path/flow to a given issue
The goal of this task is to extend Clang Static Analyzer support in sonar-cxx plugin with the full path to a given issue (which can be found under the "path" key in CSA's *.plist files; it's also called "flow" in SQ terms).

romanek-adam:
ClangStaticAnalyzer is good at finding issues which involve a certain path in code, with given constraints and values of particular variables. Its HTML reports visualize the full path from the start to the final location, showing all the intermediate steps and the assumptions taken. The path and conditions are very often critical to understand why CSA reports a given issue. Without them one can easily judge CSA for reporting a false positive, while in fact the issue is there, but the conditions are non-obvious.

Currently sonar-cxx only reports the final location of a given issue. This PR addresses this shortcoming and closes the gap to CSA's HTML reports. With this PR SonarQube is now able to present the full path to a given issue, which provides similar UX to CSA's HTML reports.

Closes #1707.

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1710)

### DIFF
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java
@@ -178,6 +178,7 @@ public abstract class CxxIssuesReportSensor extends CxxReportSensor {
 
     Set<InputFile> affectedFiles = new HashSet<>();
     List<NewIssueLocation> newIssueLocations = new ArrayList<>();
+    List<NewIssueLocation> newIssueFlow = new ArrayList<>();
 
     for (CxxReportLocation location : issue.getLocations()) {
       if (location.getFile() != null && !location.getFile().isEmpty()) {
@@ -192,11 +193,27 @@ public abstract class CxxIssuesReportSensor extends CxxReportSensor {
       }
     }
 
+    for (CxxReportLocation location : issue.getFlow()) {
+      NewIssueLocation newIssueLocation = createNewIssueLocationFile(sensorContext, newIssue, location, affectedFiles);
+      if (newIssueLocation != null) {
+        newIssueFlow.add(newIssueLocation);
+      } else {
+        LOG.debug("Failed to create new issue location from flow location {}", location);
+        newIssueFlow.clear();
+        break;
+      }
+    }
+
     if (!newIssueLocations.isEmpty()) {
       try {
         newIssue.at(newIssueLocations.get(0));
         for (int i = 1; i < newIssueLocations.size(); i++) {
           newIssue.addLocation(newIssueLocations.get(i));
+        }
+        // paths with just one element are not reported as flows to avoid
+        // presenting 1-element flows in SonarQube UI
+        if (newIssueFlow.size() > 1) {
+          newIssue.addFlow(newIssueFlow);
         }
         newIssue.save();
 
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensor.java
@@ -107,6 +107,9 @@ public class CxxClangSASensor extends CxxIssuesReportSensor {
         String filePath = ((NSString) sourceFiles[fileIndex]).getContent();
 
         CxxReportIssue issue = new CxxReportIssue(checkerName, filePath, Integer.toString(line), description);
+
+        addFlowToIssue(diag, sourceFiles, issue);
+
         saveUniqueViolation(context, issue);
       }
     } catch (Exception e) {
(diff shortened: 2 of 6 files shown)


romanek-adam:
Unfortunately my change "conflicts" with maven-license-plugin plugin. I need the copyright notice preserved as I've submitted this change on behalf of the company that I work for.

@guwirth do you know a way to handle this?

guwirth:
Hello @romanek-adam,

thanks for providing this PR. This plugin is using the LGPL license which is copyleft. If you are modifying an already existing file you can’t get the copyright. The right way would be to add you (or your company) to the contributor list in the POM file. The resulting work is again under LGPL. 

Regards,

romanek-adam:
@guwirth I wasn't aware of all this, thanks for explaining it. Please hold on a moment with merging, I need to get approval for the final shape of this change.

guwirth:
@romanek-adam maybe you can provide some sentences what your PR is doing / is intended to do.
Think developer section would be better than contributor because you are providing code: https://maven.apache.org/pom.html#Developers

romanek-adam:
@guwirth, I updated the description of this PR.

Regarding the section in pom.xml - I'd rather leave it as is. According to Maven's POM Reference 

> A healthy open source project will likely have more contributors than developers.

IMO this is a one-time contribution so I should be considered a contributor rather than a regular developer.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java
```diff
@@ -267,12 +268,26 @@ private void saveViolation(SensorContext sensorContext, CxxReportIssue issue) {
...
+      }
+    }
+
     if (!newIssueLocations.isEmpty()) {
       try {
         newIssue.at(newIssueLocations.get(0));
         for (int i = 1; i < newIssueLocations.size(); i++) {
           newIssue.addLocation(newIssueLocations.get(i));
         }
+        if (newIssueFlow.size() > 1) {
```
guwirth: Please add also text from unit test here to explain why > 1

cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java
```diff
@@ -267,12 +268,26 @@ private void saveViolation(SensorContext sensorContext, CxxReportIssue issue) {
       }
     }
 
+    for (CxxReportLocation location : issue.getFlow()) {
+      NewIssueLocation newIssueLocation = createNewIssueLocationFile(sensorContext, newIssue, location, affectedFiles);
+      if (newIssueLocation != null) {
+        newIssueFlow.add(newIssueLocation);
+      } else {
+        LOG.error("Failed to create new issue location from flow location {}", location);
```
guwirth: Should be debug not error.
Think makes no sense to mark whole analysis false in such a case.

guwirth:
@romanek-adam please also rebase it.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java
romanek-adam: Done

cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java
romanek-adam: It should either be an error or a warning - otherwise such an error would go unnoticed.

Also the following clear() and break do not mark the whole analysis false - they just make sure we don't report an incomplete flow, which would be confusing. Other than that the issue remains perfectly valid.

romanek-adam:
> @romanek-adam please also rebase it.

Done.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java
guwirth: Hi, please have a look to https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Supported-configuration-properties#sonarcxxerrorrecoveryenabled how current error behaviour is. Realization is `in CxxUtils.validateRecovery(...)`.

romanek-adam: Hi @guwirth, I wasn't aware of this flag. I made some changes in error handling. Please have a look.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java
```diff
@@ -192,12 +193,28 @@ private void saveViolation(SensorContext sensorContext, CxxReportIssue issue) {
       }
     }
 
+    for (CxxReportLocation location : issue.getFlow()) {
+      NewIssueLocation newIssueLocation = createNewIssueLocationFile(sensorContext, newIssue, location, affectedFiles);
+      if (newIssueLocation != null) {
+        newIssueFlow.add(newIssueLocation);
+      } else {
+        String msg = String.format("Failed to create new issue location from flow location %s", location);
+        LOG.debug(msg);
```
guwirth: @romanek-adam in case you log debug you don't need the validateRecovery. In case you wanna create an error it should be followed by validateRecovery. Still not sure what's the right way: Question is if such an "error" is so critical that you wanna stop analysis? Is only flow missing or is report wrong / displaying wrong data?
```Java
LOG.error(msg);
CxxUtils.validateRecovery(msg, getLanguage());
```
or
```Java
LOG.debug(msg);
```

guwirth:
@romanek-adam please rebase to head you can see it here https://github.com/SonarOpenCommunity/sonar-cxx/network that you were rebasing to an older PR.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java
romanek-adam: I assume this issue is not critical, I wouldn't stop the analysis. I'd rather clear the flow as in my initial commit, as to avoid displaying incomplete data.

```
LOG.debug("Failed to create new issue location from flow location {}", location);
newIssueFlow.clear();
break;
```

Is that OK?

guwirth: Yes, looks good.

romanek-adam:
> @romanek-adam please rebase to head you can see it here https://github.com/SonarOpenCommunity/sonar-cxx/network that you were rebasing to an older PR.

My bad. I'll do that as soon as we agree on the final look of the flow handling code.

romanek-adam:
I'm glad I could contribute. Thanks for merging it.

When do you expect v1.3.0 to be released?

guwirth:
> When do you expect v1.3.0 to be released?

You can download always latest snapshot from here: https://ci.appveyor.com/project/SonarOpenCommunity/sonar-cxx/branch/master/artifacts

Final release depends a little bit when other parts are done and how many API changes are in next SQ versions. SQ 7.8 is planned for 29th April, SQ 7.9 for 27th May. So latest end of May...

Issue #1720, opened 2019-05-13
clang-tidy: Support diagnostic notes


jspam:
clang-tidy often provides additional context for its diagnostic in the form of notes. These are currently lost when parsing a clang-tidy log file.

SonarQube supports issue flows, which seem like the appropriate way to display that information. See https://docs.sonarqube.org/display/SONARQUBE67/Issues , section "Understanding issue context"

Example 1:

```cpp
#include <cstdio>
  
int X;

int main(int argc, char** argv)
{
    int* x = nullptr;
    if (argc > 1) {
        x = &X;
    }
    printf("%d", *x);
}
```

clang-tidy output:

```
1 warning generated.
/tmp/static-analyzer.cpp:11:18: warning: Dereference of null pointer (loaded from variable 'x') [clang-analyzer-core.NullDereference]
    printf("%d", *x);
                 ^
/tmp/static-analyzer.cpp:7:5: note: 'x' initialized to a null pointer value
    int* x = nullptr;
    ^
/tmp/static-analyzer.cpp:8:9: note: Assuming 'argc' is <= 1
    if (argc > 1) {
        ^
/tmp/static-analyzer.cpp:8:5: note: Taking false branch
    if (argc > 1) {
    ^
/tmp/static-analyzer.cpp:11:18: note: Dereference of null pointer (loaded from variable 'x')
    printf("%d", *x);
                 ^
```

Example 2:

```cpp
struct A {
    void foo(int bar);
};

void A::foo(int baz) {
}
```

clang-tidy output:

```
/tmp/name-mismatch.cpp:2:10: warning: function 'A::foo' has a definition with different parameter names [readability-inconsistent-declaration-parameter-name]
    void foo(int bar);
         ^
/tmp/name-mismatch.cpp:5:9: note: the definition seen here
void A::foo(int baz) {
        ^
/tmp/name-mismatch.cpp:2:10: note: differing parameters are named here: ('bar'), in definition: ('baz')
    void foo(int bar);
         ^
```

mfenjiro:
Hey there,

I see you are using almost the same setup as me. I am trying to use clang-tidy, Jenkins, sonarqube and gerrit as a combo for my project. What is your experience with them ? What is your compile_commmands.json generator tool ? Have you find problems with it, clang-tidy and the parser of sonarscanner ?
Thank you in advance.

Best regards,

guwirth:
[...]

PR #1728, opened 2019-07-01
PC-lint: support supplemental messages

bloodlee:
Sometimes PCLint plus will generate a lot of useful supplemental messages, which helps us to locate why the issue was found. 
My change loads the supplemental into the issue flow. The following is an example.

Thanks.

image

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1728)


bloodlee:
BTW, I noticed that - if the flow only has one location, this flow will be ignored. So the newer commit uses the secondary locations, not the flow.

Please help to review this update for PC-Lint issues.

Thanks very much.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/pclint/CxxPCLintSensor.java
```diff
@@ -109,21 +133,67 @@ public void stream(SMHierarchicCursor rootCursor) throws XMLStreamException {
...
+            file = matcher.group(1);
+            line = matcher.group(2);
+          }
+        }
+
+        if (file == null || file.isEmpty() || line == null || line.isEmpty()) {
+          return;
+        }
+
+        assert !currentIssue.getLocations().isEmpty() : "current issue must have a location.";
```
guwirth: @bloodlee you should not assert here. Write a warning or error to the LOG file instead. Please have a look to other sensors how this works. In case of error seems also https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Supported-configuration-properties#sonarcxxerrorrecoveryenabled.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/pclint/CxxPCLintSensor.java
```diff
@@ -109,21 +133,67 @@ public void stream(SMHierarchicCursor rootCursor) throws XMLStreamException {
...
 
+      private void addMoreLocationsToCurrentIssue(@Nonnull CxxReportIssue currentIssue,
+                                                String file,
+                                                String line,
+                                                String msg) {
+        if (file != null && file.isEmpty() && msg != null && msg.startsWith(PREFIX_DURING_SPECIFIC_WALK_MSG)) {
+          String walkedSrcFile =
+                  msg.substring(PREFIX_MSG_LENGTH, msg.indexOf(" ", PREFIX_MSG_LENGTH));
+
+          Matcher matcher = FILE_NAME_WITH_LINE_AND_COL_PATTERN.matcher(walkedSrcFile);
```
guwirth: @bloodlee why not including the prefix into the Regel?

guwirth:
Hi @bloodlee,

do you have some time to have a look to the review comments?

Regards,

bloodlee:
Yes. Thanks for your comments. I will fix the issue soon.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/pclint/CxxPCLintSensor.java
bloodlee: The pattern is only for extracting the filename, line number. And I think matching prefix only is faster.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/pclint/CxxPCLintSensor.java
bloodlee: Agreed! I will update my implementation.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/pclint/CxxPCLintSensor.java
guwirth: Hi @bloodlee,

think there is no difference in performance.
Now you have to:
- search for string
- create a new string (copy rest of line)
- search with regex again

Regex would break immediately after one sign does not fit (same as with your compare). For my understanding code would be much better to read.

Regards,

bloodlee: You convinced me! I will commit the update.

cxx-sensors/src/main/java/org/sonar/cxx/sensors/pclint/CxxPCLintSensor.java
```diff
@@ -1,171 +1,239 @@
...
+            file = matcher.group(1);
+            line = matcher.group(2);
+          }
+        }
+
+        if (file == null || file.isEmpty() || line == null || line.isEmpty()) {
+          return;
+        }
+
+        if (currentIssue.getLocations().isEmpty()) {
```
guwirth: @bloodlee shouldn’t this test not be the first one?

guwirth:
@bloodlee name of your PR should not be `master`. This is confusing the CI.

guwirth:
@bloodlee 

> And "addFlowElement" is not working

The `CxxClangSASensor`.java is also using `addFlowElement`:
https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/clangsa/CxxClangSASensor.java#L189

If the `addFlowElement `is not generic enough we have to change it in the base class.

https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java#L215

Issue: Clang Static Analyzer: add support for full path/flow to a given issue #1707
  PR: Add path/flow support for Clang Static Analyzer *.plist reports #1710

Think you can replace the "> 1" with "> 0" and fix the test.

guwirth:
@romanek-adam do you remember why there is this `> 1`?
https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java#L215

guwirth:
>The logic in PCLint sensor is slightly different. 

@bloodlee think there are two poinzs to take into account:
1. What happens if you add an flow item which is outside of the module (no source exists)? Does this work or do you get an exception.
2. What do you/the user see in the UI? If you click an flow issue with no source what happens?

romanek-adam:
> @romanek-adam do you remember why there is this `> 1`?
> https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java#L215

Yes, I do. I tried to document it in the code as a comment but... it's only useful to report a flow when it has more than one element, like on this screenshot: https://pbs.twimg.com/media/C-Z53ywXcAAhJzA.jpg:large . Here the flow has 4 elements. Otherwise when there is just one element in a given flow we would get just one step in a flow on SonarQube UI which is a bit misleading and redundant. We would see red squares with "1" in them in a few places on the UI. IMO such issues without a multi-step path are expected to be presented on SonarQube UI like this: https://www.ndepend.com/Doc/CI_SonarQube/BrowseNDependIssuesInSQUI.png (without these red squares indicating the number of steps on the flow/path).

(I took these screenshots from Google, with no particular preference, just to show the idea)

@guwirth  does this answer your question?

guwirth:
@romanek-adam tanks for your fast response. 
* What I understood: The real value of the flow issues are these `1,2,3...` in the code to give more details to the user. In the left hand panel I see the corresponding messages.
* What's still a question for me: If I add a message and then flow issue(s), there is the original issue (in the red box) and also this first flow item (in your case: `'visitTerminal() returns null`) provides additional informaton? So I would not remove it?

The other discussion which is ongoing is: Does it make sense to add issues (locations) to the flow where no source code is available (outside of the project)? What do you think?

romanek-adam:
@guwirth 
ad 1. correct
ad 2. Clang Static Analyzer provides two types of outputs
- one which contains the full path to a given issue (like for an uninitialized variable - first when the variable is declared, then possibly some conditionals if statements, function calls etc., and finally the location when the uninitialized variable is actually used)
- another one which provides a specific location in the code, just a single line (like for dead stores, i.e. when a variable is declared and initialized on a given line and not used anywhere else) - in this case CSA provides a description like "Value stored to 'result' during its initialization is never read"; this description is presented on SonarQube UI in a red wide block; there is no need to add flow steps, even just one because the report (the .plist) file does not provide any additional information; uff, quite hard to explain without a specific example to refer to

Regarding flows for issues where there is no source code is available - I'm not sure SonarQube would accept such issues/flows - I'd expect an exception to be thrown but maybe they are silently skipped. Or maybe I'm wrong and SonarQube is actually able to present such issues/flows somehow, even without source code(?) - this needs to be verified through experimentation.

guwirth:
@romanek-adam thanks for your explanation.

Looking to #1720 (clang-tidy) there it is clear what is the warning and what is the flow data (labeled note). Wondering if this is not the same in clang sa reports? C**ould you provide a report (plist) with a flow?** In our repository I found only this report: https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/test/resources/org/sonar/cxx/sensors/reports-project/clangsa-reports/clangsa-report.plist.

romanek-adam:
> (...) In our repository I found only this report: https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/test/resources/org/sonar/cxx/sensors/reports-project/clangsa-reports/clangsa-report.plist.

Actually this report contains an issue with a two-element flow. I was using a helper script to pretty print the plist file so that I could examine the structure more easily, see https://gist.github.com/romanek-adam/70b4534a780e5cb624fd982b0b3fdc8f. The first issue in the report provides the following "path":

                  'path': [{'edges': [{'end': [{'col': 5,
                                                'file': 0,
                                                'line': 31},
                                               {'col': 7,
                                                'file': 0,
                                                'line': 31}],
                                       'start': [{'col': 5,
                                                  'file': 0,
                                                  'line': 28},
                                                 {'col': 7,
                                                  'file': 0,
                                                  'line': 28}]}],
                            'kind': 'control'},
                           {'depth': 0,
                            'extended_message': "'a' declared without an "
                                                'initial value',
                            'kind': 'event',
                            'location': {'col': 5, 'file': 0, 'line': 31},
                            'message': "'a' declared without an initial value",
                            'ranges': [[{'col': 5, 'file': 0, 'line': 31},
                                        {'col': 9, 'file': 0, 'line': 31}]]},
                           {'edges': [{'end': [{'col': 5,
                                                'file': 0,
                                                'line': 32},
                                               {'col': 6,
                                                'file': 0,
                                                'line': 32}],
                                       'start': [{'col': 5,
                                                  'file': 0,
                                                  'line': 31},
                                                 {'col': 7,
                                                  'file': 0,
                                                  'line': 31}]}],
                            'kind': 'control'},
                           {'edges': [{'end': [{'col': 8,
                                                'file': 0,
                                                'line': 32},
                                               {'col': 8,
                                                'file': 0,
                                                'line': 32}],
                                       'start': [{'col': 5,
                                                  'file': 0,
                                                  'line': 32},
                                                 {'col': 6,
                                                  'file': 0,
                                                  'line': 32}]}],
                            'kind': 'control'},
                           {'depth': 0,
                            'extended_message': 'Branch condition evaluates to '
                                                'a garbage value',
                            'kind': 'event',
                            'location': {'col': 8, 'file': 0, 'line': 32},
                            'message': 'Branch condition evaluates to a '
                                       'garbage value',
                            'ranges': [[{'col': 8, 'file': 0, 'line': 32},
                                        {'col': 8, 'file': 0, 'line': 32}]]}],

The code in CxxClangSASensor.java only parses path elements of kind 'event' which provide the exact location of a given flow step. The remaining elements of kind 'control' provide additional information which I wasn't able to map to SonarQube APIs easily.

In the report example two elements of type 'event' exist:
-                            {'depth': 0,
                            'extended_message': "'a' declared without an "
                                                'initial value',
                            'kind': 'event',
                            'location': {'col': 5, 'file': 0, 'line': 31},
                            'message': "'a' declared without an initial value",
                            'ranges': [[{'col': 5, 'file': 0, 'line': 31},
                                        {'col': 9, 'file': 0, 'line': 31}]]},
-                            {'depth': 0,
                            'extended_message': 'Branch condition evaluates to '
                                                'a garbage value',
                            'kind': 'event',
                            'location': {'col': 8, 'file': 0, 'line': 32},
                            'message': 'Branch condition evaluates to a '
                                       'garbage value',
                            'ranges': [[{'col': 8, 'file': 0, 'line': 32},
                                        {'col': 8, 'file': 0, 'line': 32}]]}

This is enough to form a flow.

guwirth:
> Do you prefer to use flow or secondary locations?

@bloodlee If you want to add a flow you should use the flow method. I'm still wondering why `addLocation` is working? Did you verify this in the UI? I still had not the time to verify the differences between flow and additional location (I mean not our implementation in the plugin, I mean the SQ API differences). Or is it at the end the same?

PR #1742, opened 2019-07-23
clang-tidy: support diagnostic notes

guwirth:
- add error, warning, note samples and tests
- add flow support: note items
- flow items are without ruleId
- refactoring
- close #1720
- close #1719 

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1742)

PR #1746, opened 2019-08-11
Correction for expression used to check if a compiler warning is valid.

stalb:
I think there is a bug in the expression used to check if a compiler warning is valid (CxxCompilerSensor.java).

Currently a compiler warning is valid if there is either a filename,  a line number, an id or a warning message, I think it should be valid only if there are both a filename,  a line number, an id and a warning message..

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1746)

### DIFF
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/CxxCompilerSensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/CxxCompilerSensor.java
@@ -106,6 +106,10 @@ public abstract class CxxCompilerSensor extends CxxIssuesReportSensor {
 
   /**
    * Derived classes can overload this method
+   * 
+   * A valid issue must have an id 
+   * and, if it has a line number, a filename.
+   * 
    *
    * @param filename
    * @param line
@@ -113,9 +117,15 @@ public abstract class CxxCompilerSensor extends CxxIssuesReportSensor {
    * @param msg
    * @return true, if valid
    */
-  protected boolean isInputValid(String filename, String line, String id, String msg) {
-    return !filename.isEmpty() || !line.isEmpty() || !id.isEmpty() || !msg.isEmpty();
-  }
+	protected boolean isInputValid(String filename, String line, String id, String msg) {
+		if ((id == null) || id.isEmpty()) {
+			return false;
+		}
+		if ((line != null) && !line.isEmpty() && ((filename == null) || filename.isEmpty())) {
+			return false;
+		}
+		return true;
+	}
 
   /**
    * Derived classes can overload this method to align filename


Bertk:
Hi Stéphane,

thank you for the PR.  The updated condition will hide any project issues (no file name or line number). Did you check this?

see https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java line 199

stalb:
In fact, II did not check.  
However, compiler sensors currently don't use issue flow...  
so I think the pertinent lines are here [CxxIssuesReportSensor.java line 183-194](https://github.com/SonarOpenCommunity/sonar-cxx/blob/d410ab59ab3e94c2341378275e081db4c5a7efd3/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java#L183-L194) and won't report any error:
```
    for (CxxReportLocation location : issue.getLocations()) {
      if (location.getFile() != null && !location.getFile().isEmpty()) {
        NewIssueLocation newIssueLocation = createNewIssueLocationFile(sensorContext, newIssue, location,
                affectedFiles);
        if (newIssueLocation != null) {
          newIssueLocations.add(newIssueLocation);
        }
      } else {
        NewIssueLocation newIssueLocation = createNewIssueLocationModule(sensorContext, newIssue, location);
        newIssueLocations.add(newIssueLocation);
      }
    }
```

If the input is not valid, a warning will be emitted at [ CxxCompilerSensor.java line 75](https://github.com/SonarOpenCommunity/sonar-cxx/blob/d410ab59ab3e94c2341378275e081db4c5a7efd3/cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/CxxCompilerSensor.java#L75), so project issues will not be completely hidden.

And, in practice, the regex used with gcc or visual c++ make sure that there is at least a non null line number, so currently the [`CxxCompilerSensor.isInputValid`](https://github.com/SonarOpenCommunity/sonar-cxx/blob/d410ab59ab3e94c2341378275e081db4c5a7efd3/cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/CxxCompilerSensor.java#L116) method returns always `true`, even if there is no id, file name or message...

The point is what should this method do ? @guwirth   
And what is a valid compiler warning input ?

guwirth:
@stalb is there a concrete problem you like to solve with this PR?

stalb:
@guwirth I was wanted to add flow support for note items in compiler sensors like you have done for clang-tidy (#1742) and had problem with this.

I didn't want to burry this inside another PR, so I thought I would first make a PR for this and another one for flow support for gcc warning notes.

guwirth:
@stalb not sure if we should change this. There are many dependencies and corner cases.
* issue for source code file
* for directory
* for module
* for project
...

stalb:
@guwirth Actually this function is doing nothing except always returning true.  
I was hopping you would remember what this function you have written should do and if it was doing it correctly.

However I don't want to break everything by changing a so crucial function.

guwirth:
Hi @stalb,

to make you happy  ;-) I created the table below:

| id | msg | fn | line | valid | comment |
|---|---|---|---|---|---|
| - | - | - | - | no | message id missing | 
| - | - | - | X | no | message id missing | 
| - | - | X | - | no | message id missing | 
| - | - | X | X | no | message id missing |
| - | X | - | - | no | message id missing |
| - | X | - | X | no | message id missing |
| - | X | X | - | no | message id missing |
| - | X | X | X | no | message id missing |
| X | - | - | - | yes | project issue; special case: only message id without concrete message |
| X | - | - | X | no | line without file makes no sense |
| X | - | X | - | yes | file issue (without concrete line number |
| X | - | X | X | yes | file issue |
| X | X | - | - | yes | project issue |
| X | X | - | X | no | line without file makes no sense |
| X | X | X | - | yes | file issue with message |
| X | X | X | X | yes | file issue in line with message |

So if we wanna change it looks for me like this:

```Java
if( id.empty() ) {
    return false;
}
if( !line.empty() && filename.empty() ) {
    return false;
}
return true;
```
Regards,

stalb:
@guwirth Thanks for the clarification.  
I will modify the PR using the proposed change if you're ok.

PR #1776, opened 2020-01-02
handle duplicate issues with different relative paths

### LINKED ISSUE #1777: clang-tidy creates duplicate issues with different relative paths
Some sensors are creating same issues with different relative paths. These issues are added several times to source files.

Sample:

    xxx/aaa/bbb/../ddd/eee/source.cpp
    xxx/aaa/ccc/../ddd/eee/source.cpp
    => xxx/aaa/ddd/eee/source.cpp

guwirth:
Some sensors are creating same issues with different relative paths. These issues were added several times to source files in the past.

This PR normalize paths before adding them to avoid this.

Sample:
- xxx/aaa/bbb/../ddd/eee/source.cpp
- xxx/aaa/ccc/../ddd/eee/source.cpp
- => xxx/aaa/ddd/eee/source.cpp

close #1777 

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1776)

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/utils/CxxReportLocation.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/utils/CxxReportLocation.java
@@ -19,6 +19,7 @@
  */
 package org.sonar.cxx.utils;
 
+import java.nio.file.Paths;
 import java.util.Objects;
 import javax.annotation.Nullable;
 
@@ -33,7 +34,11 @@ public class CxxReportLocation {
 
   public CxxReportLocation(@Nullable String file, @Nullable String line, String info) {
     super();
-    this.file = file;
+    if (file != null) {
+      this.file = Paths.get(file).normalize().toString();
+    } else {
+      this.file = null;
+    }
     this.line = line;
     this.info = info;
   }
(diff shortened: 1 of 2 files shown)


cxx-squid/src/main/java/org/sonar/cxx/utils/CxxReportLocation.java
```diff
@@ -33,7 +34,11 @@
 
   public CxxReportLocation(@Nullable String file, @Nullable String line, String info) {
     super();
-    this.file = file;
+    if (file != null) {
+      this.file = Paths.get(file).normalize().toString();
+    } else {
```
Bertk: file is already ˋnullˋ
Please remove else block.

guwirth:
Hi @ivangalkin, 

Thanks for your feedback.

The reason why I did it in CxxReportLocation is [saveUniqueViolation](https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java#L130). Here we check with a HashSet for duplicates of CxxReportIssue. This does not work with not normalized paths (and was in the past the reason for the duplicates). Think better would be to compare later in saveViolation after path is absolute. But this would be much slower?

Could also normalize inside of saveUniqueViolation?

Regards,

ivangalkin:
@guwirth your explanation makes sense completely; caching of `CxxReportIssue` with non-normalized paths is wrong.

> saveUniqueViolation

would be a better place IMHO since we don't scatter the path transforming logic across different 
location. However I see that it requires rather bigger change.

So LGTM

PR #1782, opened 2020-01-06
fix multi-values/property warning

### LINKED ISSUE #1775: How  to  solve "The SonarQube plugin declaring this property should be updated"
I have met a problem at using ths plugin sonar-c-plugin-1.3.1.1807.jar .

When the jenkins start analyzing prossess,this problem will happen as follow logs.

WARN: Property 'sonar.c.includeDirectories' is not declared as multi-values/property set but was read using 'getStringArray' method. The SonarQube plugin declaring this property should be updated.
WARN: Property 'sonar.c.forceIncludes' is not declared as multi-values/property set but was read using 'getStringArray' method. The SonarQube plugin declaring this property should be updated.
WARN: Property 'sonar.c.cFilesPatterns' is not declared as multi-values/property set but was read using 'getStringArray' method. The SonarQube plugin declaring this property should be updated.

The prossess seized up.

What should I to do ? Logs tell me to update this plugin.But the version 1.3.1 is the lastest.

Thanks.

guwirth:
- WARN: Property 'xxx' is not declared as multi-values/property set but was read using 'getStringArray' method. The SonarQube plugin declaring this property should be updated.
- close #1775

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1782)

### DIFF
--- a/sonar-c-plugin/src/main/java/org/sonar/plugins/c/CPlugin.java
+++ b/sonar-c-plugin/src/main/java/org/sonar/plugins/c/CPlugin.java
@@ -95,7 +95,8 @@ public final class CPlugin implements Plugin {
   private static List<PropertyDefinition> generalProperties() {
     String subcateg = "(1) General";
     return Collections.unmodifiableList(Arrays.asList(
-      PropertyDefinition.builder(SOURCE_FILE_SUFFIXES_KEY).multiValues(true)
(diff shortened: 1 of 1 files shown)


guwirth:
@ivangalkin 

Double checked this with Java. They also call it "comma-separated list". Think this is because it is in the config file still a comma-separated list and only in the UI different fields:

grafik

PR #1799, opened 2020-01-27
JSON Compilation Database: support -isystem includes

### LINKED ISSUE #1215: JSON Compilation Database class does not handle "-isystem" include path references

tisoft:
* -isystem includes are treated the same as -I

Fixes #1215

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1799)

### DIFF
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/JsonCompilationDatabase.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/JsonCompilationDatabase.java
@@ -132,8 +132,12 @@ public class JsonCompilationDatabase {
         }
       } else if ("-I".equals(arg)) {
         nextInclude = true;
+      } else if ("-isystem".equals(arg)) {
+        nextInclude = true;
       } else if (arg.startsWith("-I")) {
         includes.add(makeRelativeToCwd(cwd, arg.substring(2)));
+      } else if (arg.startsWith("-isystem")) {
+        includes.add(makeRelativeToCwd(cwd, arg.substring(8)));
       } else if ("-D".equals(arg)) {
         nextDefine = true;
       } else if (arg.startsWith("-D")) {


tisoft:
According to spec -isystem includes would need to be sorted after the -I includes, and there are also -iquote and -idirafter. Do we need to support all that?

tisoft:
Will add a testcase as soon as we have decided if we need the correct sorting, or if it's ok to combine -I and -isystem for now.

guwirth:
@tisoft thanks for this PR. 

> According to spec -isystem includes would need to be sorted after the -I includes, and there are also -iquote and -idirafter. Do we need to support all that?

That's always difficult to answer. Additional includes means always slower analysis times. On the other hand you need often the includes to get valid C++ code. Small sample:

You can parse this code without knowing details of the type:
```C++
void MyType func() {}
```
No valid C++ without knowing EXPORT:
```C++
void EXPORT int func() {}
```

The order of the includes is always important. To get same results as the compiler you have to ensure that search order of includes is the same as in your original environment. So I think "sort" is the wrong way.
* In best case you can use it in the same order as they are in the json file
* Makes also sense that you put all `-isystem` after the other includes, but my feeling is that you have to keep the order inside of `-isystem` and '-I'.

Maybe you can verify this in your build environment in which order the compiler gets it.

guwirth:
@tisoft 

> According to spec -isystem includes would need to be sorted after the -I includes, and there are also -iquote and -idirafter. Do we need to support all that?

Would collect in three separate lists:
- includes
- sysincludes
- dirafter 

And merge at the end in the order above. 

For `iquote` we have no support. Would handle it like `-I`. 

Maybe you can also fix this, should be 1 (not empty)
`-D<macro>=<value>`
Define `<macro>` to `<value>` (or 1 if `<value>` omitted)

PR #1801, opened 2020-01-31
Imrove compatibility of default warning regex

fandjelo:
Some older GCC compilers (f.e. 4.3.2) don't print column per default in the warning message. This leads to problems detecting the reports. This fix would allow to detect both variants without the need to redefine the regex in the configuration.

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1801)


guwirth:
Hi @fandjelo,

thanks for providing this.

Think this regex is not working on Windows with absolute paths?
`(?<file>[^:]*):`  ==> c:\path\file.ext

Regards

fandjelo:
Oh, indeed, I didn't think about that.... I'll try to improve this, if possible.

PR #1865, opened 2020-05-13
Add exception when Lint file is invalid XML

Dudenman:
The PCLint Sensor does not error out when it encounters a malformed XML, but just continues with a simple error message. In our scenarios, this has lead to developers thinking all Lint warnings had been removed, but in fact still persisted and were just not parsed. In my pull request,, I added a runtime exception when a malformed XML is detected

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1865)


guwirth:
Hello @Dudenman,

thanks for providing this. I will review it.

Regards,

Dudenman:
Hello @guwirth,

thanks for the review comments. I used the brand new InvalidReportException instead. I also used CxxUtils.validateRecovery to reroute the exception via the configuration check. Since I don't want to leave the other thrown runtime exception just hanging in there, I also took care of this one. The other exceptions / log.error can be left as is IMHO.

Cheers!

guwirth:
@Dudenman thanks for your solution. I generalized your proposal as part of #1870 and close this one.

PR #1913, opened 2020-08-06
Added a test for macros that shouldn't be substituted.

VirtualTim:
I suspect this is an issue, so just throwing it against the build system to see if it causes problems.

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/1913)

### DIFF
--- a/cxx-squid/src/test/resources/parser/preprocessor/variadic_macros.cpp
+++ b/cxx-squid/src/test/resources/parser/preprocessor/variadic_macros.cpp
@@ -7,6 +7,9 @@
 #define CHECK3(...) { printf(__VA_ARGS__); }
 #define MACRO(s, ...) printf(s, __VA_ARGS__)
 
+//Check macro substutiuon doesn't happen in the middle of a string.
+int mainMACRO();
+
 int main() {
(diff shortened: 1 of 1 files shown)


VirtualTim:
Note sure how to interpret the test failure here. Basically I've noticed issues that appear to come from functions who's name partially matches a macro, and I thought I'd add a test for it.

Minimal example of what I'm talking about (untested):
```
#include <iostream>
#define MACRO(s, ...) printf(s, __VA_ARGS__)

int mainMACRO();  //'MACRO' should not be substituted here

int main() {}
```

guwirth:
Hi @VirtualTim,

better and easier way to test this is to download and use the https://github.com/SonarOpenCommunity/sonar-cxx/releases/download/untagged-9a1e946acb840347c02e/sslr-cxx-toolkit-1.3.3.2051.jar. 
* download it
* start it
* copy your code snippet into it 

In case you get an AST the parser is able to handle it. In your case:

sample

VirtualTim:
Thanks @guwirth.
Using sslr-cxx-toolkit-1.3.3.2051.jar. It looks like the macro substitution is doing the right thing here, so I'm not sure what's triggering the case I'm seeing. I'll have to look into it more.

Issue #1685, opened 2019-02-09
Parsing of nested templates is slow: template <... template<...>>


guwirth:
Parsing of nested templates is slow if no blanks are between closing >. In extreme case user are thinking plugin is hanging.

Known workaround: add blanks between closing >.

Some real world samples:
```C++
UARTTask<BL6523GXController<UART<UART_NUM_2, 4800, 8, 'N', 1>>> uartTask;
using residual_down = add_prev2<avg_pool<2,2,2,2,skip1<tag2<block<N,BN,2,tag1<SUBNET>>>>>>;

bool ok1(templateName1<templateName<string>> declarator); // without const
bool ok2(templateName1<const templateName<string> > declarator); // blank between >>
bool failure(templateName1<const templateName<string>> declarator); // failure
```

MarcoWagner:
Is this the same problem which is slowing down the following?
[main.cpp.txt](https://github.com/SonarOpenCommunity/sonar-cxx/files/3677311/main.cpp.txt)

ericlemes:
I bumped in this issue in 1.3.2, just for information I thought would be fixed.
@guwirth  is this parser something maintained by sonar-cxx? I'd like to give a stab and see if I can fix it.

guwirth:
Hello @ericlemes,

We are currently working on a 2.0 release and your support would be very welcome.

Regards,

guwirth:
Other samples:
```C++
using List= TypeA<A, B>; // fast: 1
using List= TypeA<A, TypeB<B, C>>; // fast: 2
using List= TypeA<A, TypeB<B, TypeC<C, D>>>; // fast: 3
using List= TypeA<A, TypeB<B, TypeC<C, TypeD<D, E>>>>; // fast: 4
using List= TypeA<A, TypeB<B, TypeC<C, TypeD<D, TypeE<E, F>>>>>; // fast: 5
using List= TypeA<A, TypeB<B, TypeC<C, TypeD<D, TypeE<E, TypeF<F, G>>>>>>; // fast: 6
using List= TypeA<A, TypeB<B, TypeC<C, TypeD<D, TypeE<E, TypeF<F, TypeG<G, H>>>>>>>; // 3s: 7
using List= TypeA<A, TypeB<B, TypeC<C, TypeD<D, TypeE<E, TypeF<F, TypeG<G, TypeH<H, I>>>>>>>>; // 127s: 8
```
[...]

PR #2080, opened 2021-03-12
CxxSquidSensor: run sensor only when related rules are active

andreydanin:
SquidSensor can take significant amount of time on big projects.
Skip execution if all appropriate rules are disabled.

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/2080)

### DIFF
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxRuleRepository.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxRuleRepository.java
@@ -26,10 +26,11 @@ import org.sonar.squidbridge.annotations.AnnotationBasedRulesDefinition;
 public class CxxRuleRepository implements RulesDefinition {
 
   private static final String REPOSITORY_NAME = "SonarQube";
+  protected static final String KEY = "cxx";
 
   @Override
   public void define(Context context) {
-    NewRepository repository = context.createRepository("cxx", CxxLanguage.KEY).
+    NewRepository repository = context.createRepository(KEY, CxxLanguage.KEY).
       setName(REPOSITORY_NAME);
     new AnnotationBasedRulesDefinition(repository, CxxLanguage.KEY).addRuleClasses(false, CheckList.getChecks());
     repository.done();
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxSquidSensor.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxSquidSensor.java
@@ -269,6 +269,7 @@ public class CxxSquidSensor implements ProjectSensor {
     descriptor
       .name("CXX")
       .onlyOnLanguage("cxx")
+      .createIssuesForRuleRepository(getRuleRepositoryKey())
       .onlyOnFileType(InputFile.Type.MAIN);
   }
 
@@ -504,4 +505,7 @@ public class CxxSquidSensor implements ProjectSensor {
       .save();
   }
 
+  protected String getRuleRepositoryKey() {
+      return CxxRuleRepository.KEY;
+  }
 }


guwirth:
Hi @andreydanin, think that is not working because squid is also used for indexing, calculating metrics, syntax highlighting ...

andreydanin:
@guwirth probably you are right. I created this patch to avoid long squid scan for our pull request jobs (I disabled all cxx rules for that job in a Quality Profile). However I still can run squid if at least one cxx rule is enabled in a Quality Profile (for nightly builds in my case).

Another solution that I had in mind is to implement an option that will stop squid from running.
What do you think?

guwirth:
@andreydanin think squid in general is not slow, the problem is the preprocessor and the include files. Easiest to try is to set no (less) sonar.cxx.includeDirectories. Have you tried this?

The general issue with this is, that SonarQube needs always all source files and reports, also for a PR. There is no „incremental analysis“. Maybe you can ask this as a general question here: https://community.sonarsource.com/. 

To understand this better. How big is your code base? What execution time we are talking about?

andreydanin:
> To understand this better. How big is your code base? What execution time we are talking about?
2+ M lines of code. Execution time is 40+ minutes.
Squid is ok, but if project is big enough execution time is significant.

> The general issue with this is, that SonarQube needs always all source files and reports, also for a PR. There is no „incremental analysis“. Maybe you can ask this as a general question here: https://community.sonarsource.com/.
We use other analyzers for PR as well. SonarQube (with witsonar-cxx plugin) is one of the tools. It would be nice to be able exclude specific checks/ plugins.

> Easiest to try is to set no (less) sonar.cxx.includeDirectories. Have you tried this?
Not yet. I will try different parameters as you suggested.

guwirth:
@andreydanin 

> It would be nice to be able exclude specific checks/ plugins.

- you can turn off a plugin by setting an invalid file extension `sonar.cxx.file.suffixes` (same for other plugins)
- you can turn off checks by creating a Quality Profile with more/less rules and use this for the PR

Did you turn debug info on to see where exactly the time is used up?
https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Get-Debug-Information

> 2+ M lines of code. Execution time is 40+ minutes.

That‘s slow. Guess it’s because of many includes or #1685

Which version of the cxx plugin and SonarQube are you using?

andreydanin:
>     * you can turn off a plugin by setting an invalid file extension `sonar.cxx.file.suffixes` (same for other plugins)

Good suggestion. I will try.
 
>     * you can turn off checks by creating a Quality Profile with more/less rules and use this for the PR

No, squid for c/c++ is executed even if all rules are disabled.

> Did you turn debug info on to see where exactly the time is used up?
> https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Get-Debug-Information

Yes. This is how I found that squid for c/c++ is slowest part of check in our environment.

> That‘s slow. Guess it’s because of many includes or #1685

Thanks, I will check the discussion.

> Which version of the cxx plugin and SonarQube are you using?

SonarQube 7.9.4.35981, cxx plugin - 1.3.2 (with my patch).

guwirth:
@andreydanin in case you wanna proceed in this direction
* you should add a switch turning the preprocessor off https://github.com/SonarOpenCommunity/sonar-cxx/blob/76e71ae78a025581cf5146e88ceff5e9ac543722/cxx-squid/src/main/java/org/sonar/cxx/parser/CxxParser.java#L54. Try it without the preprocessor. 
* an internal switch registering only the active rules to be activated https://github.com/SonarOpenCommunity/sonar-cxx/blob/76e71ae78a025581cf5146e88ceff5e9ac543722/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxSquidSensor.java#L283.  To play you can forward an empty list here to see if this makes any difference.

guwirth:
@andreydanin maybe you can try and give feedback. Is this faster?

https://github.com/SonarOpenCommunity/sonar-cxx/blob/76e71ae78a025581cf5146e88ceff5e9ac543722/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxSquidSensor.java#L283

```Java
@Override
  public void execute(SensorContext context) {
    this.context = context;

    // add vistor only if corresponding rule is active
    var visitors = new ArrayList<SquidAstVisitor<Grammar>>();
    for (var check : checks.all()) {
      RuleKey key = checks.ruleKey(check);
      if (key != null) {
        if (context.activeRules().find(key) != null) {
          visitors.add(check);
        }
      }
    }

    AstScanner<Grammar> scanner = CxxAstScanner.create(createConfiguration(),
                                                       visitors.toArray(new SquidAstVisitor[visitors.size()]));
...
```

guwirth:
@andreydanin see also #2082. will wait for you feedback...

andreydanin:
> @andreydanin see also #2082, will wait for you feedback...
Thanks for the quick responses. I really appreciate it!
Unfortunately I can try your patch not earlier than this weekend.

guwirth:
@andreydanin any news?

andreydanin:
> @andreydanin any news?
Sorry for the delay. It took more time than I expected. I've performed few tests and unfortunately I don't see big difference on full and empty CXX profile with "optimize squid sensor #2082" patch. Full test is much faster (8-9 minutes vs ~36 minutes) on my machine (i7 with SSD) then on machine where I performed tests initially.
Here is short version of logs (-X -Dsonar.verbose=true): https://pastebin.com/qAZnSSwQ

"-Dsonar.exclusions=**/*.cpp" disables CXX (execution time is much faster than 8-9 minutes) but I haven't checked if cppcheck report is processed in this case.

Please let me know if you need more information or have other ideas about things that I should check.

guwirth:
@andreydanin thx for testing. It's hard for me to see from the log file which log is with/without the patch?

> I don't see big difference 

How big is the difference?

andreydanin:
> @andreydanin thx for testing. It's hard for me to see from the log file which log is with/without the patch?
> How big is the difference?

Branch:
**optimize-squid-visitors-and-squid-conditional** - contains both patches (my and yours)
**optimize-squid-visitors** - contains only your patch

Mode:
**cpp** means that I enabled only .cxx and .cpp extensions (I ignored .c, .h, etc.)

Profile:
**empty** - "Sonar way", 0 enabled rules
**full** - 5425 cxx rules are enabled

CXX_time:
value from log string "Sensor CXX [cxx] (done) | time=\<value\>"
270000ms = 4.5 min

Total_time:
value from log string "INFO: Total time: \<value\> s"

```
Branch                                         Mode  Profile  CXX_time  Total_time
optimize-squid-visitors-and-squid-conditional  cpp   empty    none      2:40.589
optimize-squid-visitors-and-squid-conditional  cpp   full     277759ms  7:36.565
optimize-squid-visitors                        cpp   empty    340701ms  9:19.440
optimize-squid-visitors                        cpp   empty    267705ms  7:25.019
optimize-squid-visitors                        cpp   empty    280359ms  7:37.511
optimize-squid-visitors                        cpp   empty    265813ms  7:29.725
optimize-squid-visitors                        cpp   empty    262919ms  7:20.962
optimize-squid-visitors                        cpp   full     270328ms  7:21.942
optimize-squid-visitors                        cpp   full     334975ms  8:40.202
optimize-squid-visitors                        cpp   full     278051ms  7:44.835
optimize-squid-visitors                        cpp   full     276215ms  7:30.724
```

guwirth:
@andreydanin thx

Still wondering which are two comparable lines in the list for
**empty** - "Sonar way", 0 enabled rules
**full** - 5425 cxx rules are enabled

What is the time difference between full and empty?

```
Branch                                         Mode  Profile  CXX_time  Total_time
optimize-squid-visitors                        cpp   empty    340701ms  9:19.440
optimize-squid-visitors                        cpp   empty    267705ms  7:25.019
optimize-squid-visitors                        cpp   empty    280359ms  7:37.511
optimize-squid-visitors                        cpp   empty    265813ms  7:29.725
optimize-squid-visitors                        cpp   empty    262919ms  7:20.962
optimize-squid-visitors                        cpp   full     270328ms  7:21.942
optimize-squid-visitors                        cpp   full     334975ms  8:40.202
optimize-squid-visitors                        cpp   full     278051ms  7:44.835
optimize-squid-visitors                        cpp   full     276215ms  7:30.724
```

andreydanin:
I performed multiple tests in each configuration to eliminate interference with background tasks on my machine.
If we remove longest test in each group we can see that total execution time is close to 7 minutes 30 seconds for both configurations. IMHO it's hard to tell that one configuration was faster than the other.

guwirth:
Hi @andreydanin, I merged the PR #2082 to 2.0. Don‘t know if faster but can not hurt...

andreydanin:
Hi @guwirth 

I've rebased my patch on top of 2.0.5 and performed some tests on our projects. For tests I enabled debug log (-X parameter for scanner). 

Longes execution was found for "Sensor CXX". I checked output like this:
> INFO: Sensor CXX [cxx]
> ...
> INFO: Sensor CXX [cxx] (done) | time=1958588ms

Note: for cppcheck related part we expect log line:
> INFO: Sensor CXX Cppcheck report import [cxx]

I checked 3 configurations:

**1. Without "Sensor CXX"**
No rules for squid. Only cppcheck related rule is enabled.

	'INFO: Sensor CXX [cxx]' was not found in log
	Execution time: ~ 1.5-2 min


**2. With "Sensor CXX" (rule "Avoid too many code lines in source file" is enabled)**

	'INFO: Sensor CXX [cxx]' was found in log
	Execution time: ~ 50 min


**3. With "Sensor CXX" (rule "Avoid too many code lines in source file" is enabled, Preprocessor::handlePreprocessorDirective is disabled)**
I checked process stack during execution (using jstack) and found that in most cases process was within Preprocessor::handlePreprocessorDirective.
I replaced this function with "return oneConsumedToken(token);" to check how long it will process files without it. But I don't know what functionality is affected. Probably it's a bad idea to disable this function.

	'INFO: Sensor CXX [cxx]' was found in log
	Execution time: ~ 9 min


Do you have any ideas what should I check next?

guwirth:
Hi @andreydanin,

thx for testing. 

Like to understand it better:
* In case of 1) did you do any other changes on the source code? 
* 2) is same as 1) except you activated rule rule "Avoid too many code lines in source file" is enabled? Also no changes on the sources code?
* With 3) you changed the source code. For my understanding this turns off the preprocessor. Are the activated rules here same as 1) or 2)?

Regards,

andreydanin:
Hi. Thanks for the quick response.

> In case of 1) did you do any other changes on the source code?

Yes. All three test cases are tested with the patch from this pull request applied. It should only have effect in test case 1 (when squid execution is skipped because it's rules are not active).

> is same as 1) except you activated rule rule "Avoid too many code lines in source file" is enabled?

Yes.

> Also no changes on the sources code?

Source code was changed. I can run all three test cases without my changes but I expect that test case 2 will be the same and case 1 will become same as 2 (squid will be active even rules are disabled).

> With 3) you changed the source code. For my understanding this turns off the preprocessor.

Yes. I believe I disabled preprocessor in test case 3.

> Are the activated rules here same as 1) or 2)?

Rules were same as in test case 2.


Test case 1 - only cppcheck rules are active. Squid ("Sensor CXX") was not executed as expected because patch from this pull request is applied.

Test cases 2 and 3 - rules from 1 plus rule "Avoid too many code lines in source file". Additional rule activated squid ("Sensor CXX") in test cases 2 and 3.

guwirth:
Hi @andreydanin,

my main concern is still that the software metrics are not working with your patch? Did you compare the values, e.g. LoC, Complexity, … with the cases 1) to 3)…

Regards

guwirth:
Hi @andreydanin,

think there are still different ways:
1. Improve the performance of the squid sensor. Maybe you can help me by doing some measurements with a profiler.
2. Disable the squid sensor in the configuration.  Here I would prefer ‚sonar.cxx.squid.disabled‘ (default =false) over your solution, this is more explizit.
3. Disable the preprozessor only. Here I’m not sure if completely or only the include-file handling part?

If you can provide a solution for 1) I would merge it. 

If you could provide more detailed numbers with https://www.ej-technologies.com/products/jprofiler/overview.html it would help to take a better decision. If you ask them they will provide a free license for open source projects. 

Regards

PR #2198, opened 2021-06-16
C++ Parser can't read code. Declaration is skipped.

guwirth:
- Add last known token with position. Problem is most likely before/after this token

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/2198)


guwirth:
@dbochard here is a version adding more information:
https://ci.appveyor.com/project/SonarOpenCommunity/sonar-cxx/builds/39629198/artifacts

Problem is: could be before after the last token. Please have also a look to the sample in the PR to understand the problem.

Issue #2241, opened 2021-09-01
SonarQube 9.x compatibility


arthepsy:
Latest release doesn't seem to work on latest SonarQube (9.0.1):
```
java.lang.NoSuchMethodError: 'org.sonar.api.server.rule.RulesDefinition$NewRule org.sonar.api.server.rule.RulesDefinition$NewRule.setEffortToFixDescription(java.lang.String)'
        at org.sonar.cxx.squidbridge.annotations.AnnotationBasedRulesDefinition.setupSqaleModel(AnnotationBasedRulesDefinition.java:184)
        at org.sonar.cxx.squidbridge.annotations.AnnotationBasedRulesDefinition.addRuleClasses(AnnotationBasedRulesDefinition.java:102)
        at org.sonar.cxx.squidbridge.annotations.AnnotationBasedRulesDefinition.addRuleClasses(AnnotationBasedRulesDefinition.java:88)
        at org.sonar.plugins.cxx.CxxRuleRepository.define(CxxRuleRepository.java:34)
        at org.sonar.server.rule.RuleDefinitionsLoader.load(RuleDefinitionsLoader.java:53)
        at org.sonar.server.rule.RegisterRules.start(RegisterRules.java:117)
        at org.sonar.core.platform.StartableCloseableSafeLifecyleStrategy.start(StartableCloseableSafeLifecyleStrategy.java:40)
        at org.picocontainer.injectors.AbstractInjectionFactory$LifecycleAdapter.start(AbstractInjectionFactory.java:84)
        at org.picocontainer.behaviors.AbstractBehavior.start(AbstractBehavior.java:169)
        at org.picocontainer.behaviors.Stored$RealComponentLifecycle.start(Stored.java:132)
        at org.picocontainer.behaviors.Stored.start(Stored.java:110)
        at org.picocontainer.DefaultPicoContainer.potentiallyStartAdapter(DefaultPicoContainer.java:1016)
        at org.picocontainer.DefaultPicoContainer.startAdapters(DefaultPicoContainer.java:1009)
        at org.picocontainer.DefaultPicoContainer.start(DefaultPicoContainer.java:767)
        at org.sonar.core.platform.ComponentContainer.startComponents(ComponentContainer.java:136)
        at org.sonar.server.platform.platformlevel.PlatformLevel.start(PlatformLevel.java:90)
[...]

PR #2274, opened 2021-11-04
Cleanup deprecated code in 8.9 to make plugin run on 9.1

yachoor:
I needed the plugin running on latest sonar 9.1 without errors (#2241)

I've updated targeted version to latest LTS - 8.9.3.4875 and fixed compilation problems (internal code used in tests moved to sonar-plugin-api-impl dependency, missing gson dependency in one place, `org.sonar.api.issue.NoSonarFilter` now abstarct class - replaced usage in tests with `org.sonar.api.batch.sensor.issue.internal.DefaultNoSonarFilter`).

Then I moved to fix deprecated functions in 8.9, that are missing in 9.1 - most were simply renamed, the only significant change is that [org.sonar.api.server.rule.RulesDefinition.NewRule.setDebtSubCharacteristic](https://javadocs.sonarsource.org/8.9.3.48735/org/sonar/api/server/rule/RulesDefinition.NewRule.html#setDebtSubCharacteristic(java.lang.String)) is gone in 9.1 without replacement (it's documented to do nothing in 8.9), so I removed the call, marked `@SqualeSubCharacteristic`, which was used to set it, as deprecated and fixed logic to check for `@Squale*Remediation` annotations instead

The plugin runs on 9.1 and builds with passing tests targeting 8.9.3.4875

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/2274)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ jdk:
 env:
   - SONARLABEL=sonarqube-7.9.6 SONARAPI=SqApi79
   - SONARLABEL=sonarqube-8.9.2.46101 SONARAPI=SqApi79
+  - SONARLABEL=sonarqube-9.1.0.47736 SONARAPI=SqApi79
 
 addons:
   # shorten the VM hostname with the new workaround
--- a/cxx-checks/pom.xml
+++ b/cxx-checks/pom.xml
@@ -42,6 +42,11 @@
       <artifactId>commons-lang</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.sonarsource.sonarqube</groupId>
+      <artifactId>sonar-plugin-api-impl</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.sonarsource.sslr</groupId>
       <artifactId>sslr-testing-harness</artifactId>
       <scope>test</scope>
--- a/cxx-sensors/pom.xml
+++ b/cxx-sensors/pom.xml
@@ -29,6 +29,15 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.sonarsource.sonarqube</groupId>
+      <artifactId>sonar-plugin-api-impl</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.sonarsource.sonarqube</groupId>
       <artifactId>sonar-testing-harness</artifactId>
       <scope>test</scope>
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/StaxParser.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/StaxParser.java
@@ -63,12 +63,9 @@ public class StaxParser {
    */
   public StaxParser(XmlStreamHandler streamHandler, boolean isoControlCharsAwareParser) {
     this.streamHandler = streamHandler;
-    var xmlFactory = XMLInputFactory.newInstance();
-    if (xmlFactory instanceof WstxInputFactory) {
-      var wstxInputfactory = (WstxInputFactory) xmlFactory;
-      wstxInputfactory.configureForLowMemUsage();
-      wstxInputfactory.getConfig().setUndeclaredEntityResolver(new UndeclaredEntitiesXMLResolver());
-    }
+    WstxInputFactory xmlFactory = new WstxInputFactory();
+    xmlFactory.configureForLowMemUsage();
+    xmlFactory.getConfig().setUndeclaredEntityResolver(new UndeclaredEntitiesXMLResolver());
     xmlFactory.setProperty(XMLInputFactory.IS_VALIDATING, Boolean.FALSE);
     xmlFactory.setProperty(XMLInputFactory.SUPPORT_DTD, Boolean.FALSE);
     xmlFactory.setProperty(XMLInputFactory.IS_NAMESPACE_AWARE, Boolean.FALSE);
--- a/cxx-squid-bridge/src/main/java/org/sonar/cxx/squidbridge/annotations/AnnotationBasedProfileBuilder.java
+++ b/cxx-squid-bridge/src/main/java/org/sonar/cxx/squidbridge/annotations/AnnotationBasedProfileBuilder.java
@@ -23,7 +23,6 @@
  */
 package org.sonar.cxx.squidbridge.annotations;
 
-import org.sonar.api.profiles.ProfileDefinition;
 import org.sonar.api.profiles.RulesProfile;
 import org.sonar.api.rules.RuleFinder;
 import org.sonar.api.utils.AnnotationUtils;
@@ -31,8 +30,7 @@ import org.sonar.api.utils.ValidationMessages;
 
 /**
  * Utility class to build an instance of {@link RulesProfile} based on a list of classes annotated
- * with {@link ActivatedByDefault}. It can be used to implement the <code>createProfile</code>
- * method of {@link ProfileDefinition}.
+ * with {@link ActivatedByDefault}.
  *
  * @since 2.5
  */
--- a/cxx-squid-bridge/src/main/java/org/sonar/cxx/squidbridge/annotations/AnnotationBasedRulesDefinition.java
+++ b/cxx-squid-bridge/src/main/java/org/sonar/cxx/squidbridge/annotations/AnnotationBasedRulesDefinition.java
@@ -44,15 +44,11 @@ import org.sonar.cxx.squidbridge.rules.ExternalDescriptionLoader;
 /**
  * Utility class which helps setting up an implementation of {@link RulesDefinition} with a list of
  * rule classes annotated with {@link Rule}, {@link RuleProperty} and SQALE annotations:
- * <ul>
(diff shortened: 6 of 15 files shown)


yachoor:
Hmm, seems like the builds are failing because cxx-sslr-toolkit-2.0.6-SNAPSHOT.jar size (7799261) too small. Min. is 8000000
Probably because sonar-plugin-api size was reduced from 4MB to 788KB by moving some code into sonar-plugin-api-impl

guwirth:
Hello @yachoor,

thanks a lot for providing this update. This seems to be a bigger change which needs some additional review effort. Will try it as fast as possible.

Regards,

guwirth:
@yachoor in case you wanna test with SQ 9.1 you have to adapt https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/.travis.yml. Please add 9.1 and comment out 7.9 and 8.9 to get faster results and use up not some much Travis credits.

grafik

```INI
env:
  # - SONARLABEL=sonarqube-7.9.6 SONARAPI=SqApi79
  # - SONARLABEL=sonarqube-8.9.2.46101 SONARAPI=SqApi79
  - SONARLABEL=sonarqube-9.1.0.47736 SONARAPI=SqApi79
```

yachoor:
@guwirth ahh, thanks
Also it seems by using latest LTS 8.9.3.48735 `sonar-plugin-api` version I'm preventing it from running on SonarQube 8.9.2.46101
What's the policy? Should I target earliest LTS version - 8.9.0.43852?

yachoor:
Ok, figured it out - I think I should set Sonar-Version property manually to 8.9 and not the default (version of sonar-plugin-api dependency)

guwirth:
@yachoor for 9.1: Easiest is always to compare with other plugins in https://github.com/SonarSource. Python and Java are always good samples to compare:
https://github.com/SonarSource/sonar-java
https://github.com/SonarSource/sonar-python

yachoor:
Ahh, nevermind I see property sonarQubeMinVersion now, so it should work fine on travis 8.9.2.46101, but didn't
I'll have to investigate why it failed

yachoor:
One test has failed on 9.1, but runs fine on 8.9 - checking for log message on empty cppcheck report.
Judging by the logs from previous builds it's supposed to be:
`WARN: The 'Cppcheck V2' report is empty, report='/home/travis/build/SonarOpenCommunity/sonar-cxx/integration-tests/testdata/cppcheck_project/empty.xml', cause='WstxEOFException: Unexpected EOF in prolog  at [row,col {unknown-source}]: [1,0]', skipping`
but was:
`WARN: The 'Cppcheck V2' report is invalid, report='/home/travis/build/SonarOpenCommunity/sonar-cxx/integration-tests/testdata/cppcheck_project/empty.xml', cause='XMLStreamException: ParseError at [row,col]:[1,1] Message: Premature end of file.', skipping`

It seems the plugin uses different XML parsers depending on SonarQube version

yachoor:
@guwirth I've fixed sonarQubeMinVersion to be actually used, but left it at 7.9 - the functions that were removed in 9.1 were deprecated since 5.5 so all the changes should be fine. Restored travis testing on 7.9, 8.9 and added 9.1 - seems to work fine on all of them

guwirth:
> checking for log message on empty cppcheck report.

Hi @yachoor, this check for 'empty XML files' was always a special case where we caught an `WstxEOFException `exception and were throwing an own `EmptyReportException` exception. All XML sensors has such code. 

> plugin uses different XML parsers

Don't think that the plugin is using another parser, seems to be more likely that a newer version of the library is throwing a different (new) exception.

Using a different XML library could also have an impact to the performance. That's a thing we have to investigate.

guwirth:
Hi @yachoor,

first of all: thanks a lot for your contribution! Have to think a bout what we are doing with this PR now?

First problem is a little bit that after the change of the business model of Travis we have only a limited number of credits to build the plugin. Building for 3 SQ versions in parallel is expensive.

Second question for me is: If we add 9.x support, should we remove the SQ 7.9 support? If answer is yes, this seems to be a major release then (3.x, or at least 2.1)?

I would also like to proceed with some beta testing. In the past some people were already asking for 9.x support. Maybe they can test the snapshot.

Regards,

yachoor:
> Don't think that the plugin is using another parser, seems to be more likely that a newer version of the library is throwing a different (new) exception.
> 
> Using a different XML library could also have an impact to the performance. That's a thing we have to investigate.

It was using default `XMLInputFactory`, which on SonarQube 9.1 was using implementation from JRE java.xml (`java.xml/com.sun.org.apache.xerces.internal.impl.XMLStreamReaderImpl`), throwing different exception earlier

I [fixed it](https://github.com/SonarOpenCommunity/sonar-cxx/pull/2274/commits/d3c5c6319c86e73fc13848712a77b8c1852553d3#diff-ad44cc5de730767c1d62b797708d9a649dd7e8f3ac72f774ee1435533481e98cL66) by using the expected `WstxInputFactory`

yachoor:
> Hi @yachoor,
> 
> first of all: thanks a lot for your contribution! Have to think about what we are doing with this PR now?
> 
> First problem is a little bit that after the change of the business model of Travis we have only a limited number of credits to build the plugin. Building for 3 SQ versions in parallel is expensive.
>
> Second question for me is: If we add 9.x support, should we remove the SQ 7.9 support? If answer is yes, this seems to be a major release then (3.x, or at least 2.1)?

Hi @guwirth

Well, you can still officially support only previous/current LTS - 7.9 + 8.9. Or drop support for 7.9 - are people still using it?
I needed to get it running on 9.1 because I'm working on project in kotlin + C++, so I wanted recent SonarQube with kotlin support. The changes were generally to get rid of code deprecated since 5.5 that was finally removed in 9.x and should be fine with all SonarQube versions. But that doesn't mean anything newer than current LTS has to be officially supported (although it's nice to be ahead and not wait until next LTS release to make all necessary changes at once to get it working)

I would go with official support for current LTS and travis testing on current LTS + most recent SonarQube that works (update when new one comes out to see if any work is needed; if it doesn't work, file an issue and go back to previous until fixed). Not sure about `sonarQubeMinVersion` - setting it to 8.9 will prevent people from using plugin with older versions, even though it would still work. But not setting it to a version that is actually tested does not seem good either.
 
> I would also like to proceed with some beta testing. In the past some people were already asking for 9.x support. Maybe they can test the snapshot.

Good idea, I'll use it for sure 👍

guwirth:
Hi @yachoor,

could you squash and rebase this PR please. I fixed some import issues in 2.0.6. That's makes it easier to review it.

Regards,

yachoor:
Hi @guwirth, I've squashed and rebased the PR.

guwirth:
Use this version for testing: https://github.com/SonarOpenCommunity/sonar-cxx/releases/tag/cxx-2.1.x-beta

yachoor:
Hi @guwirth, I'm a little confused about cxx-2.1.x-beta - should it contain changes from this PR (it doesn't)?

yachoor:
Ah, nvm was looking at commits and sources, binary is fine :)

guwirth:
@yachoor because it's not merges I can't create a tag, so automagically attached sources are wrong.

yachoor:
Works fine for me on 9.1, will test on next one as soon as it comes out (seems the only way to get updates to builtin plugins since 8.9 is to update whole sonarqube instance and 9.1 is one version of java plugin and four versions of kotlin plugin behind sonarcloud)

guwirth:
I created a branch `master-2.1.x` to attach the right source code for the release:
https://github.com/SonarOpenCommunity/sonar-cxx/releases/tag/cxx-2.1.x-beta1

DJViking:
Tried the latest 2.1-beta on SonarQube 9.2.4. Places the JARs in SonarQube extensions/plugin directory.
SonarQube will not start with this plugin.
```
2022.01.24 14:36:49 ERROR web[][o.s.s.p.PlatformImpl] Web server startup failed
java.lang.NullPointerException: Plugin key is missing from manifest
        at java.base/java.util.Objects.requireNonNull(Objects.java:246)
        at org.sonar.core.platform.PluginInfo.<init>(PluginInfo.java:104)
        at org.sonar.server.plugins.ServerPluginInfo.<init>(ServerPluginInfo.java:32)
        at org.sonar.server.plugins.ServerPluginInfo.create(ServerPluginInfo.java:38)
        at org.sonar.server.plugins.PluginJarLoader.lambda$getExternalPluginsMetadata$5(PluginJarLoader.java:205)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
        at java.base/java.util.LinkedList$LLSpliterator.forEachRemaining(LinkedList.java:1239)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
        at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
        at org.sonar.server.plugins.PluginJarLoader.loadPluginsFromDir(PluginJarLoader.java:235)

guwirth:
@yachoor maybe you have an idea?

That's the reason why we normally provide only plugins for LTS versions. The changes in non LTS versions from one to the next SQ versions are mostly too big.

PR #2278, opened 2021-11-12
Update targeted version to current LTS - 8.9.3.48735 and fix deprecat…

guwirth:
…ed code

* Add necessary dependencies - sonar-plugin-api-impl (test only), gson
* Replace now abstract org.sonar.api.issue.NoSonarFilter with org.sonar.api.batch.sensor.issue.internal.DefaultNoSonarFilter
* Remove deprecated (since 5.5) call to setDebtSubCharacteristic
* Mark unused annotation SqualeSubCharacteristic as Deprecated for removal
* Replace old deprecated functions with new ones
* Remove reference to ProfileDefinition, so the plugin can be built with sonar 9.1.0.47736 too
* Reduce minimal enforced artifact size (sonar-plugin-api dependency was reduced in size. Now the artifact size is around 7.7MB)
* Fix compatible SonarQube version declared by the plugin (sonarQubeMinVersion was not used)
* Use woodstox for XML parsing instead of some default provider (fixes different XML parsing on newer SonarQube versions)

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/2278)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ jdk:
 env:
   - SONARLABEL=sonarqube-7.9.6 SONARAPI=SqApi79
   - SONARLABEL=sonarqube-8.9.2.46101 SONARAPI=SqApi79
+  - SONARLABEL=sonarqube-9.1.0.47736 SONARAPI=SqApi79
 
 addons:
   # shorten the VM hostname with the new workaround
--- a/cxx-checks/pom.xml
+++ b/cxx-checks/pom.xml
@@ -42,6 +42,11 @@
       <artifactId>commons-lang</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.sonarsource.sonarqube</groupId>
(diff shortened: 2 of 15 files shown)


.travis.yml
```diff
@@ -8,6 +8,7 @@ jdk:
 env:
   - SONARLABEL=sonarqube-7.9.6 SONARAPI=SqApi79
   - SONARLABEL=sonarqube-8.9.2.46101 SONARAPI=SqApi79
+  - SONARLABEL=sonarqube-9.1.0.47736 SONARAPI=SqApi79
```
lg2de, line 11: Should be `SqApi89`, isn't it?

guwirth:
@lg2de current version is working with 7.9 to 9.x, so `SqApi79` is correct.
If we are moving to 2.x we remove the 7.9 support and will also update this.

PR #2291, opened 2022-01-05
fix parsing error with right angle bracket detection

### LINKED ISSUE #2286: Question: Reproduce parsing errors on local machine
I have a lot of parsing errors in my codebase.

https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Detect-and-fix-parsing-errors has some suggestions but I lack a short hello-world project for testing.
Is it possible to reproduce the parse errors on my local machine without a full-blown analysis?

guwirth:
- close #2286

```C++
void check() {
    a() < 0 || c >= 1;
    a() < 0 || c > 1;
    a() < 0 || c >> 1;

    if(a() < 0 || c >= 1) {}
    if(a() < 0 || c > 1) {}
    if(a() < 0 || c >> 1) {}

    Y<X<(6<1)>>;
    Y<X<(6>=1)>>;
    Y<X<(6<=1)>>;
    Y<X<(6>>1)>>;
    Y<X<(6<<1)>>;
    Y<X<(6<=>1)>>;
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/2291)

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/channels/RightAngleBracketsChannel.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/channels/RightAngleBracketsChannel.java
@@ -43,29 +43,45 @@ import org.sonar.sslr.channel.CodeReader;
  */
 public class RightAngleBracketsChannel extends Channel<Lexer> {
 
-  private int angleBracketLevel = 0;
-  private int parentheseLevel = 0;
+  private int angleBracketLevel = 0; // angle brackets <  >
+  private int parentheseLevel = 0;   // parentheses / round brackets ( )
 
   @Override
   public boolean consume(CodeReader code, Lexer output) {
     var ch = (char) code.peek();
     var consumed = false;
 
-    if (ch == '<') {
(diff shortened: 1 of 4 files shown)


amai2012:
@guwirth  What about this?
`void check() {
    a() < 0 || c >> 1;
}`

guwirth:
> What about this? void check() { a() < 0 || c >> 1; }
@amai2012 ...I also had this in mind, but more complicated :-). Need more timer to think about this, do you have a proposal?

amai2012:
> > What about this? void check() { a() < 0 || c >> 1; }
> 
> @amai2012 ...I also had this in mind, but more complicated :-). Need more time to think about this, do you have a proposal?

I didn't try, just got curious while looking at the commit...
In case it fails we could add another known issue to be addressed later?

guwirth:
@amai2012 please have a look to current solution, this should fix your issue. You can test with https://ci.appveyor.com/project/SonarOpenCommunity/sonar-cxx/builds/42156789/artifacts.

PR #2315, opened 2022-02-15
fix placement new issue

### LINKED ISSUE #2314: new with placement-params is not parsed
The new with pacement params ( 4th syntax  at https://en.cppreference.com/w/cpp/language/new ) is not recognized.
It is used to create an object on an already allocated memory position pointed by p
```::new(p)A(1,2,3);```

sslr reports an error: ```com.sonar.sslr.api.RecognitionException: Parse error at line 1 column 2:```
and sonarqube on similar code: ```C++ Parser can't read code. Declaration is skipped (last token=')', line=102, column=17).```

### PR DESCRIPTION
**Pull request by guwirth**
- close #2314

Sample:
```C++
::new(p)A(1,2,3)
```

---
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/2315)

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/parser/CxxGrammarImpl.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/parser/CxxGrammarImpl.java
@@ -766,16 +766,13 @@ public enum CxxGrammarImpl implements GrammarRuleKey {
           "gcnew" // C++/CLI
         ),
         b.firstOf(
-          b.sequence("(", typeId, ")"), // syntax sugar (todo)
-          b.sequence(
-            b.optional(newPlacement),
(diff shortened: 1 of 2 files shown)


slnj:
Hi Günter,
Its now accepting the placement-param, but it still requiren the returned value to be assigned.
It's accepting ```p = ::new(p)A(1,2,3)``` but not just ```::new(p)A(1,2,3)```
When the pointer is passed as placement-param,  it is returned unchanged so the returned value should be able to be discarded.
Regards

slnj:
My fault. I tried with sslr toolkit without putting it on a function and there it's only accept the assignment one, but putting it inside a function it is accepted.
It's running ok.

Regards

PR #2430, opened 2022-09-30
Only analyze files that are contained in the 'JSON Compilation Database'

guwirth:
In case of `sonar.cxx.jsonCompilationDatabase` is used, limit the analyzed files to the files contained in the 'JSON Compilation Database' file. The sensor is an intersection of the files configured via `sonar.projectBaseDir` and the files contained in the 'JSON Compilation Database' file.

In the past, all files configured via `sonar.projectBaseDir` were always analyzed and only the configuration (INCLUDES, MACROS) was read from the 'JSON Compilation Database'.

- filter `sonar.projectBaseDir` files result
- add a 'JSON Compilation Database' support integration test

**Hint:** Unfortunately, the sensor cannot control the indexing of the source files. Therefore, all files that belong to language CXX and are located within the directory `sonar.projectBaseDir` are loaded onto the server, but are no longer analyzed, which saves time. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/2430)

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/config/CxxSquidConfiguration.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/config/CxxSquidConfiguration.java
@@ -25,6 +25,7 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedList;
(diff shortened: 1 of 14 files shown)


guwirth:
@thyros and @Kinokin there is another improvement for the Json Compilation Database sensor. Maybe you can try it and give feedback. 

https://github.com/SonarOpenCommunity/sonar-cxx/suites/8897425714/artifacts/407082091

Kinokin:
Well, this may break our analysis. Just a few days ago we re-enabled sonar-cxx scanning after
* using a unity build to get clang-tidy from 2h down to 10 minutes analysis time
* using 'latest' plugin (2.1 (build 311))
which worked well for us.

Unity build looks like

unity_lib.cpp:
include "source_a.cpp"
include "source_b.cpp"
include "source_c.cpp"
include "source_d.cpp"
and compilation database only contains unity_lib.cpp

Filtering by compilation database will drop all source_*.cpp files now.
Can this behaviour be disabled?

guwirth:
Hi @Kinokin,

currently there is no option but we can add one.
```ini
# default False
sonar.cxx.jsonCompilationDatabase.analyzeOnlyContainedFiles=True/False
```
What do you think?

Regards

Kinokin:
Yes, this would be sufficient, thanks!

guwirth:
Hi @Kinokin you can try with latest snapshot, see #2449

PR #2449, opened 2022-11-17
sonar.cxx.jsonCompilationDatabase.analyzeOnlyContainedFiles=True/False

guwirth:
If `analyzeOnlyContainedFiles=True` is used, the analyzed files will be limited to the files contained in the 'JSON Compilation Database' file - the intersection of the files configured via `sonar.projectBaseDir` and the files contained in the 'JSON Compilation Database' file (default is False).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sonaropencommunity/sonar-cxx/2449)


Kinokin:
Thanks. However, it will take some time until I can update our server again to test.

PR #2606, opened 2024-01-08
New feature to report unknown rules / warnings

### LINKED ISSUE #2586: Rule template for GCC custom rules not working due to key validation: - sign not allowed
**Describe the bug**
When analyzing gcc compiler warnings I wanted to add a new one for -Wstringop-truncation

When trying to enter this key, I get a validation error, that key is only allowed to be upper and lower case letters, digits and "_" , so "-" is not allowed. Since this is used as id from the regex, I don't know how to use this key...

Also backup, manual edit of xml and restore does not work

**To Reproduce**
Try to create a new custom rule from gcc rule template 

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Linux]
 - SonarQube version: 9.9LTS
 - cxx plugin version: 2.11
 - sonar-scanner version: 5.0.1.3006

jenspopp:
This is a fix for https://github.com/SonarOpenCommunity/sonar-cxx/issues/2586

 I tested it with gcc, cppcheck and clangtidy.

Open points:

1. Default unknown rule is right now only for the tested frameworks in the xml definitions
2. We could use protobuf instead of json (but at least for my local use case it was so fast, that I didn't see a reason to invest time here)
3. In the long run we might need to replace all "-" with "_" in the rule ids.... (due to sonar community changes)
4. I'm not sure, I covered all possible exceptions (e.g. with login/authentication); but they should occur already before the cxx plugin
5. Right now only login with token is supported

close #2586

- - -
This change is [Reviewable](https://reviewable.io/reviews/SonarOpenCommunity/sonar-cxx/2606)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,9 @@ This plugin is free software; you can redistribute it and/or modify it under the
   * GNU extensions
   * `CUDA` extensions
 * Microsoft Windows and Linux for runtime environment
+* Handling of unknown / new errors and warnings (e.g. new compiler warnings)
+  * unknown warnings/errors (that are not defined in sonar) will now be mapped to the rule id="unknown"
+  * the rule id can be customized with the property sonar.cxx.unknown.rule.id and is valid for all sensors
 
 Sensors for **static and dynamic code analysis**:
 * **Cppcheck** warnings support (http://cppcheck.sourceforge.net/)
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/gcc/CxxCompilerGccSensor.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/compiler/gcc/CxxCompilerGccSensor.java
@@ -43,7 +43,7 @@ public class CxxCompilerGccSensor extends CxxCompilerSensor {
   public static final String DEFAULT_ID = "default";
   public static final String DEFAULT_REGEX_DEF
                                = "(?<file>[^:]*+):(?<line>\\d{1,5}):\\d{1,5}:\\x20warning:\\x20"
-                                   + "(?<message>.*?)(\\x20\\[(?<id>.*)\\])?\\s*$";
+                                   + "(?<message>.*?)(\\x20\\[(?<id>[^\\[]*)\\])?\\s*$";
 
   public static List<PropertyDefinition> properties() {
     var subcateg = "GCC";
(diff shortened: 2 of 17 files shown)


guwirth:
Hi @jenspopp,

thanks for providing this I will have a look to it.

Regards,

guwirth:
@jenspopp thx for fixing the tests

jenspopp:
> @jenspopp thx for fixing the tests

Hi, please note that some tests will fail with Java 17 due to some Exception changes in 17....

guwirth:
Risk too high to add this to V2.1 version,will be merged into V2.2 version.

guwirth:
@jenspopp I like to start merging this pull request ... Are you still using it? Any additional comments/experiences?

jenspopp:
Hi, we are using it for a few month now. Since it is only 2 projects, it is not to big of a user base, but it helped already to identify some potential issues. We use an opensuse leap with latest gcc compilers to detect potential risks early on, and that works.

Out current Sonar is LTS 9.9(.2)

guwirth:
@jenspopp For my understanding: Why did you decide to load the rules from the server? An alternative would have been to read the rules from the plugin, as the XML files are part of it.

jenspopp:
Sorry, been a while. But if I remember correctly, it was to allow the user to create a new rule (e.g. if a new compiler warning is introduced) in sonar web ui, without changing the code/xml files.

guwirth:
> Sorry, been a while. But if I remember correctly, it was to allow the user to create a new rule (e.g. if a new compiler warning is introduced) in sonar web ui, without changing the code/xml files. So that it is not marked as unknown and can be classified and searched...

Yes you are right that's possible with template rules.

guwirth:
@jenspopp What is the use case of changing the `ruleId` for unkown (with `sonar.cxx.unknown.rule.id`)? Additional settings must be tested, variants are to be tested, ...

jenspopp:
If I read my comments correctly, it was to be able to configure the rule id set to unknown rules. This can also be a fixed value in code... I think if the id was not found, there would be no error/warning in sonar otherwise...

  * unknown warnings/errors (that are not defined in sonar) will now be mapped to the rule id="unknown"
  * the rule id can be customized with the property sonar.cxx.unknown.rule.id and is valid for all sensors

guwirth:
Hi @jenspopp, struggling a little bit with the key for the unkown rule:
* keep key "unkown"
* use key "UnkownRule"
* ???

Also think severity should be INFO and type CODE_SMELL

 ```XML
<rule>
    <key>unknown</key>
    <name>Unknown compiler warnings</name>
    <description>
      Unknown compiler warnings.
    </description>
    <severity>INFO</severity>
    <type>CODE_SMELL</type>
  </rule>
```

What are you thinking?

Regards,

PR #2691, opened 2024-06-07
no files found from JSON Compilation Database on linux if project contains upper case file names

kragens:
convert files in JSON Compilation Database to lower case because unifyPath() converted file names to lower case

- - -
This change is [Reviewable](https://reviewable.io/reviews/SonarOpenCommunity/sonar-cxx/2691)

### DIFF
--- a/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxSquidSensor.java
+++ b/sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxSquidSensor.java
@@ -23,6 +23,7 @@ import com.sonar.cxx.sslr.api.Grammar;
 import java.io.File;
 import java.io.Serializable;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -400,7 +401,7 @@ public class CxxSquidSensor implements ProjectSensor {
       // then analyze only the files contained in the db.
       var inputFilesInConfig = squidConfig.getFiles();
       var result = StreamSupport.stream(inputFiles.spliterator(), false)
-        .filter(f -> inputFilesInConfig.contains(Path.of(f.uri())))
+        .filter(f -> inputFilesInConfig.contains(Paths.get((Path.of(f.uri()).toString()).toLowerCase(Locale.getDefault()))))
         .collect(Collectors.toList());
       inputFiles = result;
 


guwirth:
Hi @kragens,

thanks for providing this. This issues seems to be related to #2700?

Not sure if your fix works on all operating systems. Reading here https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Troubleshooting-Reports#file-path-issues there are also case-sensitive operating systems where making all lower-case is not working?

> Depending on the operating system and SonarQube Database settings paths can be case-sensitive (also on case-insensitive operating systems). Report sensors always try to restore the case-senitive path to avoid problems.

Have to check this.

Regards,

sonar-cxx-plugin/src/main/java/org/sonar/plugins/cxx/CxxSquidSensor.java
```diff
@@ -400,7 +401,7 @@ private Iterable<InputFile> getInputFiles(SensorContext context, CxxSquidConfigu
       // then analyze only the files contained in the db.
       var inputFilesInConfig = squidConfig.getFiles();
       var result = StreamSupport.stream(inputFiles.spliterator(), false)
-        .filter(f -> inputFilesInConfig.contains(Path.of(f.uri())))
+        .filter(f -> inputFilesInConfig.contains(Paths.get((Path.of(f.uri()).toString()).toLowerCase(Locale.getDefault()))))
```
guwirth, line 404: Depending on the operating system and SonarQube Database settings paths can be case-sensitive (also on case-insensitive operating systems). Think making it always lowercase does not work?

guwirth:
Below the solution from the report sensors:

https://github.com/SonarOpenCommunity/sonar-cxx/blob/398033454fca8749927a9f6bc14a22e77de4b828/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportSensor.java#L70

https://github.com/SonarOpenCommunity/sonar-cxx/blob/398033454fca8749927a9f6bc14a22e77de4b828/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportSensor.java#L115

kragens:
> thanks for providing this. This issues seems to be related to #2700?
I don't think so, because all files in compile_commands.json are lowercase in #2700

The problem here is that alle files in inputFilesInConfig were made lowercase by
https://github.com/SonarOpenCommunity/sonar-cxx/blob/398033454fca8749927a9f6bc14a22e77de4b828/cxx-squid/src/main/java/org/sonar/cxx/config/CxxSquidConfiguration.java#L566

guwirth:
Hello @kragens, thx for the hint will have a look to it and try to find out why all is converted to lowercase …

guwirth:
solved with #2725

guwirth:
Hi @kragens , you can test it with
https://github.com/SonarOpenCommunity/sonar-cxx/actions/runs/10574618430/artifacts/1858527022

kragens:
Tested successfully in release v2.2.0. thanks for fixing it @guwirth

PR #2747, opened 2024-09-10
Improve handling of relative file paths in report files

### LINKED ISSUE #2741: Relative paths with double dot at the beginning (..\) fail in reports
PathUtils::sanitize in sonar-plugin-api is described as "Normalize path and replace file separators by forward slash", but the only thing it does is call `return FilenameUtils.normalize(path, true);` (https://github.com/SonarSource/sonar-plugin-api/blob/master/plugin-api/src/main/java/org/sonar/api/utils/PathUtils.java#L42). https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/FilenameUtils.html#normalize(java.lang.String,boolean) explains that

> A double dot will cause that path segment and the one before to be removed. If the double dot has no parent path segment to work with, null is returned. 

So the CxxReportLocation [constructor](https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-squid/src/main/java/org/sonar/cxx/utils/CxxReportLocation.java#L39) may return something with a line and a column, but a "null" file, if the "file" it's given is something like "../dir/file.c".

In https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxIssuesReportSensor.java#L164 only the file is checked. So, probably I should not have those `../` paths in my build log to start with, but if somebody happens to have them he ends up with issues being listed in SonarQube with no location. Usually, when a file is not "known", the issue is simply ignored (https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportSensor.java#L84), so there is a strange discrepancy here.

The issue can be reproduced with sonar-cxx 2.1.2 making `-Dsonar.cxx.gcc.reportPaths=` point to a file with something like `../file.c:1:1: warning: message [-Wdiv-by-zero]` (supposing `-Wdiv-by-zero` is in the quality profile).

guwirth:
- Up to know we were stating in our [Wiki](https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Troubleshooting-Reports): Relative paths in report files are always relative to the project base directory. Start relative paths always with .\ on Windows or ./ on Linux.
- relative paths starting with '..' are resolved to null. "No file" results in adding an issue on project level (project issue). In such a case we replace null with the filename now. This leads at least to an error message and the issue is not added to the project because typically no such indexed file exists.
- The real problem is that tools generate reports with relative paths without defining which base directory is referred to. This problem remains and must be solved via the CI/CD and tool configuration.
- close #2741

- - -
This change is [Reviewable](https://reviewable.io/reviews/SonarOpenCommunity/sonar-cxx/2747)

### DIFF
--- a/cxx-squid/src/main/java/org/sonar/cxx/utils/CxxReportLocation.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/utils/CxxReportLocation.java
@@ -19,9 +19,9 @@
  */
 package org.sonar.cxx.utils;
 
-import java.nio.file.Paths;
 import java.util.Objects;
 import javax.annotation.Nullable;
+import org.apache.commons.io.FilenameUtils;
 import org.sonar.api.utils.PathUtils;
 
 /**
@@ -42,10 +42,13 @@ public class CxxReportLocation {
     // FilenameUtils.normalize internally, relative paths starting with a double dot will cause that path segment
     // and the one before to be removed. If the double dot has no parent path segment to work with, null is returned.
     // null would mean 'project issue' which is wrong in this context. To avoid this we extract the filename to
-    // generate at least a meningful error message.
+    // generate at least a meningful error message (#2747).
     var normalized = PathUtils.sanitize(file);
     if (normalized == null && (file != null && !file.isBlank())) {
-      normalized = Paths.get(file).getFileName().toString();
+
+      // use FilenameUtils.getName because this works on Windows and Linux also if
+      // report is generated on the one and consumed on the other
+      normalized = FilenameUtils.getName(file);
     }
 
     this.file = normalized;
--- a/cxx-squid/src/main/java/org/sonar/cxx/utils/CxxReportLocation.java
+++ b/cxx-squid/src/main/java/org/sonar/cxx/utils/CxxReportLocation.java
@@ -19,6 +19,7 @@
  */
 package org.sonar.cxx.utils;
 
+import java.nio.file.Paths;
 import java.util.Objects;
 import javax.annotation.Nullable;
 import org.sonar.api.utils.PathUtils;
@@ -35,8 +36,19 @@ public class CxxReportLocation {
 
   public CxxReportLocation(@Nullable String file, @Nullable String line, @Nullable String column, String info) {
     super();
-    // normalize file, removing double and single dot path steps => avoids duplicates
-    this.file = PathUtils.sanitize(file);
+
+    // Normalize file using separators in UNIX format, removing double and single dot path steps. This is to avoid
+    // duplicates in the issue containers because they are using the file as key. PathUtils.sanitize uses
+    // FilenameUtils.normalize internally, relative paths starting with a double dot will cause that path segment
+    // and the one before to be removed. If the double dot has no parent path segment to work with, null is returned.
+    // null would mean 'project issue' which is wrong in this context. To avoid this we extract the filename to
+    // generate at least a meningful error message.
+    var normalized = PathUtils.sanitize(file);
+    if (normalized == null && (file != null && !file.isBlank())) {
+      normalized = Paths.get(file).getFileName().toString();
+    }
+
+    this.file = normalized;
     this.line = line;
     this.column = column;
     this.info = info;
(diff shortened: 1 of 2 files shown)


guwirth:
Hello @cmorve-te,

thanks for your test. Can you provide a report file to reproduce the issue. With the unit tests it works? Like to understand what’s different.

Regards,

cmorve-te:
I guess anything starting with ../../ will fail. Which makes sense, it gets converted to ./../XXX, which is equivalent to ../XXX, which was the original problem.

```
--- a/cxx-squid/src/test/java/org/sonar/cxx/utils/CxxReportLocationTest.java
+++ b/cxx-squid/src/test/java/org/sonar/cxx/utils/CxxReportLocationTest.java
@@ -57,6 +57,8 @@ public class CxxReportLocationTest {
     assertThat(loc.getFile()).isEqualTo("dir/File");
     loc = new CxxReportLocation("../dir/File", null, null, "");
     assertThat(loc.getFile()).isEqualTo("dir/File");
+    loc = new CxxReportLocation("../../dir/File", null, null, "");^M
+    assertThat(loc.getFile()).isEqualTo("dir/File");^M
 
     loc = new CxxReportLocation("c:\\dir\\file.ext", null, null, "");
     assertThat(loc.getFile()).isEqualTo("c:/dir/file.ext");
```

fails with
```
org.opentest4j.AssertionFailedError: 

expected: "dir/File"
 but was: null
	at org.sonar.cxx.utils.CxxReportLocationTest.testPathSanitize(CxxReportLocationTest.java:61)
```

Honestly, not even sure what I would write in the assertion. `../../XXX` isEqualTo `XXX` start to be forced.
I guess the best solution would be making anything starting with .. relative to sonar.projectBaseDir. But I don't really mind, in my specific case the path is wrong anyway. The only important thing is that it doesn't end up being null/a "project" issue.

guwirth:
Hi @cmorve-te,

you are right, I was not testing with relative path with more than one level like `../../file`. This will fail again.

> I guess the best solution would be making anything starting with .. relative to sonar.projectBaseDir

This can also fail: In case `projectBaseDir` is `/home/sample` and path in report file `../../../file` it's the same again. 

In the meantime I tend to simply add the file name in case `PathUtils.sanitize(file)` returns `null`. This way at least no project issue is created and the error message remains helpful.

What do you think?

cmorve-te:
> This can also fail: In case projectBaseDir is /home/sample and path in report file ../../../file it's the same again.

Indeed.

> In the meantime I tend to simply add the file name in case PathUtils.sanitize(file) returns null. This way at least no project issue is created and the error message remains helpful.

That would work for me, but can you just add the filename?
My understanding was that you need an "InputFile" (https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/CxxReportSensor.java#L70), that must exist in the "context.fileSystem()". That just a string to a non-existing path would not be accepted by the Sonar API.

In general, that's something I would like from Sonar. Instead of hoping the warnings don't point to an (existing) file that's not in `sonar.sources`, it would be nice if any file a warning points to would be automatically added to the list of sources (but not for coverage purposes!) and I could see it from SonarQube. Even if it's a system header, I don't care too much about that code, but having it SonarQube could help understanding the warnings.

So your "simply add the file name" is probably the best option. If not possible, I guess I would just fallback to just drop any warnings that resolve to a null file in the same way they are dropped if the resolve to a valid file not in `sonar.sources`.

cmorve-te:
Oh, you meant only the file, without any path, and only inside the CxxReportLocation constructor!
Hopefully everybody has his sources under src, fine with me.

cmorve-te (approved):

guwirth:
@cmorve-te maybe you can give this a try: https://github.com/SonarOpenCommunity/sonar-cxx/actions/runs/10972510569/artifacts/1961586003

In case result is null try with filename only. This should avoid project issue generation and should give you a warning in the LOG file.

guwirth:
Hi @cmorve-te, like to release V2.1.3 and add this one. Any feedback?

cmorve-te:
Verified with my real output, it's all fine now.

Again, unless the file name happens to match the name of a real file in the root of the project, I guess. My project doesn't have source files in the root of the project at all.

guwirth:
@cmorve-te thanks for testing I know this is not a perfect solution but an improvement.
Solution in release available: https://github.com/SonarOpenCommunity/sonar-cxx/releases/tag/cxx-2.1.3

PR #3001, opened 2026-01-22
Add new env vars for integration test suite

Magyarimiki:
SonarQube recently prefers token based authentication instead of username and password.

This PR adds support for SONAR_TOKEN and SONAR_HOST_URL variables.
Also, the integration test readme is updated as the information for setting the credentials was completely missing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/SonarOpenCommunity/sonar-cxx/3001)

### DIFF
--- a/integration-tests/Readme.txt
+++ b/integration-tests/Readme.txt
@@ -16,6 +16,11 @@ Make sure the following preconditions are met, before running the test suite:
 
 Usage
 =====
+SonarQube credentials are read from environment variables:
+ - Set SONAR_TOKEN to use token based authentication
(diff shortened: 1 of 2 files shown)


guwirth (changes requested):
Think all locations must be adapted that this PR is working.

guwirth:
Hi @Magyarimiki,

for my understanding there should be default values for SONAR_URL and SONAR_TOKEN or the integration test should use predefined values stored in the environment variables. In case of using the predefined external values they should be used at all code locations and not only in some?

Regards,

PR #3044, opened 2026-04-14
bugfix - correctly resolve truststore path on filesystem

Hakky54:
Someone on stackoverflow mentioned that there is an issue with the sonar scanner and it crashes. See here for the question on stackoverflow: https://stackoverflow.com/questions/79923615/how-to-disable-ssl-in-sonar-cxx-plugin-when-using-an-http-sonarqube-server

Basically during startup it crashes because it has trouble loading the truststore. The stacktrace is:
```
09:00:49.892 ERROR Error during SonarScanner Engine execution
nl.altindag.ssl.exception.GenericKeyStoreException: java.lang.IllegalArgumentException: Failed to load the keystore from the classpath for the given path: [/usr/lib/jvm/java-21-amazon-corretto.x86_64/lib/security/cacerts]
    at nl.altindag.ssl.util.KeyStoreUtils.loadKeyStore(KeyStoreUtils.java:91)
    at nl.altindag.ssl.SSLFactory$Builder.withTrustMaterial(SSLFactory.java:308)
    at org.sonar.cxx.sensors.utils.SSLContextBuilder.createSSLContext(SSLContextBuilder.java:107)
    at org.sonar.cxx.sensors.utils.SonarServerWebApi.setServerConfig(SonarServerWebApi.java:174)
    at org.sonar.cxx.sensors.utils.CxxIssuesReportSensor.downloadRulesFromServer(CxxIssuesReportSensor.java:95)
    at org.sonar.cxx.sensors.utils.CxxIssuesReportSensor.executeImpl(CxxIssuesReportSensor.java:84)
    at org.sonar.cxx.sensors.utils.CxxReportSensor.execute(CxxReportSensor.java:101)
    ...
Caused by: java.lang.IllegalArgumentException: Failed to load the keystore from the classpath for the given path: [/usr/lib/jvm/java-21-amazon-corretto.x86_64/lib/security/cacerts]
    at nl.altindag.ssl.util.internal.ValidationUtils.requireNotNull(ValidationUtils.java:41)
    at nl.altindag.ssl.util.KeyStoreUtils.loadKeyStore(KeyStoreUtils.java:88)
    ... 26 common frames omitted
```

The user is asking for help to disable ssl, which is not available in sonar actually. however the root cause is clear. The ayza library is loading the truststore from the classpath as the path is defined as a String type. As it tries to find it on the classpath it fails as it does not exist there. By changing the type of the truststorepath from String to Path it will look into the filesystem and it will load correctly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/SonarOpenCommunity/sonar-cxx/3044)

### DIFF
--- a/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/SSLContextBuilder.java
+++ b/cxx-sensors/src/main/java/org/sonar/cxx/sensors/utils/SSLContextBuilder.java
@@ -101,12 +101,13 @@ public class SSLContextBuilder {
           password -> sslFactoryBuilder.withIdentityMaterial(keyStorePathString, password.toCharArray(), "PKCS12"),
           () -> sslFactoryBuilder.withIdentityMaterial(keyStorePathString, "changeit".toCharArray(), "PKCS12"));
     }
-    var trustStorePathString = config.get("sonar.scanner.truststorePath")
-        .orElse(findSonarHome(config).resolve("ssl/keystore.p12").toString());
-    if (trustStorePathString != null && Files.exists(Path.of(trustStorePathString))) {
+    var trustStorePath = config.get("sonar.scanner.truststorePath")
+        .map(Path::of)
+        .orElse(findSonarHome(config).resolve("ssl/keystore.p12"));
+    if (trustStorePath != null && Files.exists(trustStorePath)) {
       config.get("sonar.scanner.truststorePassword").ifPresentOrElse(
-          password -> sslFactoryBuilder.withTrustMaterial(trustStorePathString, password.toCharArray(), "PKCS12"),
-          () -> sslFactoryBuilder.withTrustMaterial(trustStorePathString, "changeit".toCharArray(), "PKCS12"));
+          password -> sslFactoryBuilder.withTrustMaterial(trustStorePath, password.toCharArray(), "PKCS12"),
+          () -> sslFactoryBuilder.withTrustMaterial(trustStorePath, "changeit".toCharArray(), "PKCS12"));
     }
     return sslFactoryBuilder.build().getSslContext();
   }


guwirth:
@jenspopp please review

jenspopp:
The change will not introduce any difference in the behavior in my opinion. The real issue is, that the getResourceAsStream throws a IOException and returns "null". The only change in the original issue will be, that the error will than have no validation but an IOException (from Files.newInputStream method) as cause of the error. 

For me the originating problem is, that opening /usr/lib/jvm/java-21-amazon-corretto.x86_64/lib/security/cacerts causes an IOException. We already did a Files.exists(Path.of(trustStorePathString)), so the file exists, but opening the Stream causes an IOException... So this will not change, if we just switch to Path.

The issue seems to be that sonar.scanner.truststorePath is set to /usr/lib/jvm/java-21-amazon-corretto.x86_64/lib/security/cacerts; but that file is not readable to the user/service running the thread. (it can be seen as exists, but not read) Please be aware that each run of scanner downloads a new java runtime and unzips it in the cache directory and runs the plugins in that jdk... and that might have different rights in a container.

Hakky54:
The `getResourceAsStream` tries to get the file from the classhpath actually. The file `/usr/lib/jvm/java-21-amazon-corretto.x86_64/lib/security/cacerts` is obviously not on the classpath but on the filesystem. The Ayaz library makes a decision to look on the classpath if the location is defined as a String type or on the filesystem when it is defined as a Path object.

So that is why I wrapped the file location in a Path object in my code changes.

In the original code there was this check:
```
if (trustStorePathString != null && Files.exists(Path.of(trustStorePathString))) {
```

Which is first wrapping the location in a Path object and then using Files utility class to check whether the file exists on on the filesystem. And it passes that check, the validation is saying that the file exists and after that SSLFactory tries to load it, however the variable passed to SSLFactory is of the type String so it tries to load it from the classpath and that is actually the flow and the reason of the error. So a simple change of wrapping the location in a Path objects ensures that SSLFactory will load it from the filesystem.

I will try to provide more context to justify the code changes for the pull request:
So currently the file is being loaded at this place: https://github.com/Hakky54/ayza/blob/master/ayza/src/main/java/nl/altindag/ssl/util/KeyStoreUtils.java#L86-L93

After the code changes of this PR, it should load the file from a different method. To be exact it should use this one: https://github.com/Hakky54/ayza/blob/master/ayza/src/main/java/nl/altindag/ssl/util/KeyStoreUtils.java#L111-L117

The input of those two methods differentiate on the type of `keystorePath` which you can see in the method parameter. I hope this makes sense and why this issue is occuring

jenspopp:
> The `getResourceAsStream` tries to get the file from the classhpath actually. The file `/usr/lib/jvm/java-21-amazon-corretto.x86_64/lib/security/cacerts` is obviously not on the classpath but on the filesystem. The Ayaz library makes a decision to look on the classpath if the location is defined as a String type or on the filesystem when it is defined as a Path object.
> 

Sorry, you are completely correct... shouldn't have done the review on the train ;-) Just checked the sonarqube code, they wrapped the config in a mapper, that does the conversion to a CertificateStore object and get the path from there. 

It would still be better to get the configured wsclient object directly from the sonar scanner, to have an identical configuration and save on overhead, but since it is not public.... 

So from my side the correction is valid, was my mistake using the string and not the path variant. Also checked other instances where I used @Hakky54's library and always used path... don't know why I didn't in this case... @Hakky54 - thanks for the catch!

Hakky54:
It was my pleasure 😊 have a nice day

guwirth:
@Hakky54 new version for download available https://github.com/SonarOpenCommunity/sonar-cxx/releases/tag/latest-snapshot

guwirth:
@Hakky54 and @jenspopp thx for your support!

PR #3102, opened 2026-08-03
Add symbol resolution and semantic API 

chmodshubham:
Currently, the AST visitors in sonar-cxx only see syntax - identifiers, tokens, node types - with no idea what an identifier actually refers to. This PR adds a semantic layer on top: a Symbol API (with TypeSymbol, VariableSymbol, FunctionSymbol specializations) plus a new CxxSymbolResolverVisitor that walks the AST during the normal scan and resolves declarations and usages into that model.

Signed-off-by: Shubham Kumar [chmodshubham@gmail.com](mailto:chmodshubham@gmail.com)

- - -
This change is [Reviewable](https://reviewable.io/reviews/SonarOpenCommunity/sonar-cxx/3102)

### DIFF
--- a/cxx-squid-bridge/src/main/java/org/sonar/cxx/squidbridge/api/AstNodeSymbolExtension.java
+++ b/cxx-squid-bridge/src/main/java/org/sonar/cxx/squidbridge/api/AstNodeSymbolExtension.java
@@ -20,6 +20,8 @@
 package org.sonar.cxx.squidbridge.api;
 
 import com.sonar.cxx.sslr.api.AstNode;
+import java.util.Collections;
+import java.util.Map;
 import java.util.WeakHashMap;
 import javax.annotation.CheckForNull;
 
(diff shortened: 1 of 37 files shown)


guwirth:
Hello @chmodshubham,

thanks for providing this.

I have some questions:

1. Could you describe the value from this PR from user perspective? What is the value added?
2. Did the added tests cover the new functionality? Coverage for New Code > 80%?
3. Did you do some performance tests? Is the scanner slower with this change?

Regards,

chmodshubham:
1. Right now, sonar-cxx can't properly resolve a variable used inside a function back to its declaration. It sees the two as unrelated tokens, but with this PR, sonar-cxx can resolve that kind of symbol usage. This is the same kind of symbol/semantic layer sonar-java already has ([Symbol.java](https://github.com/SonarSource/sonar-java/blob/master/java-frontend/src/main/java/org/sonar/plugins/java/api/semantic/Symbol.java)) — it's what lets Java rules know that a name used somewhere is the same as the thing declared elsewhere, instead of just matching text patterns. After this, we can write checks for unused variables/parameters in the source code.

Also, on a personal note: I'm using sonar-cxx in my own project and need symbol resolution there. Rather than build it inside my project, figured it's better long-term to contribute it upstream so everyone using sonar-cxx benefits.

2. Yes, it should be above 80%. I have missed some tests, so I added a few more.

3. I didnt find any tool to do the performance testing. Is there any? let me know.

guwirth:
Hello @chmodshubham,

the main purpose of the cxx plugin is to forward reports from 3rd party tools to SonarQube. This is the main difference to other plugins doing the static code analysis by their own (https://github.com/SonarOpenCommunity/sonar-cxx/wiki/Design-Guidelines).

What we don't want to do is:
* to develop another static code analyzer

In case your PR provides the basis to close #1401 we can think about adding it. But precondition is, that there is no (little) performance impact. Scanning a project should not be slower.

Regards,

chmodshubham:
Not another code analyzer, and added support for #1401 too.

Issue #3120, opened 2026-09-21
sonar-cxx default severity levels of PCLint-Plus rules


tnielens:
sonar-cxx defines severity levels of PCLint-Plus rules [here](https://github.com/SonarOpenCommunity/sonar-cxx/blob/master/cxx-sensors/src/main/resources/pclint.xml). These severity levels don't match the official warning levels from PCLint-Plus defined [here](https://help.vector.com/pclintplus/Messages.html).

For example sonar-cxx sets the severity to blocker and tags the issue with 'tool-error' to a series of rules in the 9xxx range although PCLint-Plus sets the warning level of 'note' for this rule range. A concrete example, rule 9233:
```xml
  <rule>
    <key>9233</key>
    <name>L9233: bitwise operator __operator__ may not be applied to operand with signed underlying type</name>
    <description>
      <![CDATA[An expression with a MISRA C signed underlying type was provided as an
operand to a bitwise operator.]]>
    </description>
    <tag>tool-error</tag>
    <severity>BLOCKER</severity>
    </rule>
  <rule>
    <key>9234</key>
    <name>L9234: shift amount exceeds size of operand's underlying type</name>
    <description>
      <![CDATA[An expression was shifted by a negative amount or an amount greater than
the bit width of the expression's MISRA C underlying type.]]>
    </description>
    <tag>tool-error</tag>
    <severity>BLOCKER</severity>
    </rule>
```
Is sonar-cxx deliberately remapping the pclint warning levels in pclint.xml?
If not, would it make sense to stick to the PCLint-Plus default mapping?

# Annex 
Official PCLint-Plus warning levels:
Range | Description | Warning Level
-- | -- | --
1-199 | C Syntax Errors | 1
200-299 | Internal Errors | 1
300-399 | Fatal Errors | 1
400-699 | C Warnings | 2
700-899 | C Informational | 3
900-999 | C Elective Notes | 4
1000-1199 | C++ Syntax Errors | 1
1200-1299 | Internal Errors | 1
1300-1399 | C++ Fatal Errors | 1
1400-1699 | C++ Warnings | 2
1700-1899 | C++ Informational | 3
1900-1999 | C++ Elective Notes | 4
2000-2199 | C Syntax Errors | 1
2200-2399 | Reserved
2400-2699 | C Warnings | 2
2700-2899 | C Informational | 3
2900-2999 | C Elective Notes | 4
3000-3199 | C++ Syntax Errors | 1
[...]

PR #3121, opened 2026-09-21
use pclint-plus warning levels for the corresponding sonar rules severity

tnielens:
This PR proposes to define the Sonar rules severity of PC-lint Plus lints according to the PC-lint Plus warning levels.

PC-lint Plus warning levels are defined here https://help.vector.com/pclintplus/Messages.html .

The proposed relation between the PC-lint Plus warning levels and the Sonar rules severity is this one:

| PC-lint Plus Warning Level | Sonar Severity |
| - | - |
| - | Blocker |
| Error | Critical |
| Warning | Major |
| Info | Minor |
| Note | Info | 

This PR sets the Sonar severity level to the Sonar rules according to the PC-lint Plus default warning levels.

Relates to issue #3120.

- - -
This change is [Reviewable](https://reviewable.io/reviews/SonarOpenCommunity/sonar-cxx/3121)


guwirth:
Hi @tnielens,

thanks for providing this PR.

Regards,

cxx-sensors/src/main/resources/pclint.xml
```diff
@@ -1732,49 +1733,55 @@ This occurs for a use of the GCC attribute mode() as in:</p>
...
-    <severity>BLOCKER</severity>
+    <severity>CRITICAL</severity>
     </rule>
   <rule>
     <key>401</key>
     <name>L0401: symbol 'Symbol' not previously declared static at Location</name>
     <description>
       <![CDATA[The indicated Symbol declared static was previously declared without the static storage class. This is technically a violation of the ANSI standard. Some compilers will accept this situation without complaint and regard the Symbol as static.]]>
     </description>
+    <severity>MAJOR</severity>
```
guwirth: > Possible values are INFO, MINOR, MAJOR (default), CRITICAL, BLOCKER.

To keep the plug-in small we avoid to set the default value in the xml file.

guwirth (changes requested):

cxx-sensors/src/main/resources/pclint.xml
tnielens: I committed the removal in 28c7e55c30a2e52.
Imo it's better to leave the MAJOR, because pclint should use MAJOR, and not the default severity set by sonar-cxx. If the default severity changes in the future, it will invalidate the pclint.xml rule mapping.

